Issue #14, opened 2024-08-15
AttributeError: 'MARBLComponent' object has no attribute 'time_step'


TomNicholas:
When running the notebook I got this error (I ran some cells out of order)

```python
marbl_component=cstar.MARBLComponent(
                    base_model=marbl_base_model,
                    additional_code=marbl_additional_code
)
marbl_component   
```
```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/formatters.py:711](http://localhost:8888/lab/tree/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/formatters.py#line=710), in PlainTextFormatter.__call__(self, obj)
    704 stream = StringIO()
    705 printer = pretty.RepresentationPrinter(stream, self.verbose,
    706     self.max_width, self.newline,
    707     max_seq_length=self.max_seq_length,
    708     singleton_pprinters=self.singleton_printers,
    709     type_pprinters=self.type_printers,
    710     deferred_pprinters=self.deferred_printers)
--> 711 printer.pretty(obj)
    712 printer.flush()
    713 return stream.getvalue()

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/lib/pretty.py:419](http://localhost:8888/lab/tree/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/lib/pretty.py#line=418), in RepresentationPrinter.pretty(self, obj)
    408                         return meth(obj, self, cycle)
    409                 if (
    410                     cls is not object
    411                     # check if cls defines __repr__
   (...)
    417                     and callable(_safe_getattr(cls, "__repr__", None))
    418                 ):
--> 419                     return _repr_pprint(obj, self, cycle)
    421     return _default_pprint(obj, self, cycle)
    422 finally:

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/lib/pretty.py:787](http://localhost:8888/lab/tree/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/lib/pretty.py#line=786), in _repr_pprint(obj, p, cycle)
    785 """A pprint that just redirects to the normal repr function."""
    786 # Find newlines and replace them with p.break_()
--> 787 output = repr(obj)
    788 lines = output.splitlines()
    789 with p.group():

[...]

Issue #12, opened 2024-08-14
ValueError: supplied checkout_target does not appear to be a valid reference for this repository


TomNicholas:
Running the first two cells of the example notebook gives

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 3
      1 #BaseModel object, ROMS:
----> 3 roms_base_model=cstar.ROMSBaseModel(
      4                            source_repo='https://github.com/dafyddstephenson/ucla-roms.git',
      5                            checkout_target='marbl_improvements_20240611')
      7 marbl_base_model=cstar.MARBLBaseModel(
      8                             source_repo='https://github.com/marbl-ecosys/MARBL.git',
      9                             checkout_target='marbl0.45.0')

File [~/Documents/Work/Code/C-Star/cstar_ocean/base_model.py:89](http://localhost:8888/lab/tree/~/Documents/Work/Code/C-Star/cstar_ocean/base_model.py#line=88), in BaseModel.__init__(self, source_repo, checkout_target)
     81 self.source_repo = (
     82     source_repo if source_repo is not None else self.default_source_repo
     83 )
     84 self.checkout_target = (
     85     checkout_target
     86     if checkout_target is not None
     87     else self.default_checkout_target
     88 )
---> 89 self.checkout_hash = _get_hash_from_checkout_target(
     90     self.source_repo, self.checkout_target
     91 )
     92 self.repo_basename = os.path.basename(self.source_repo).replace(".git", "")
     94 self.local_config_status = self.get_local_config_status()

File [~/Documents/Work/Code/C-Star/cstar_ocean/utils.py:119](http://localhost:8888/lab/tree/~/Documents/Work/Code/C-Star/cstar_ocean/utils.py#line=118), in _get_hash_from_checkout_target(repo_url, checkout_target)
    117         return checkout_target
    118     else:
--> 119         raise ValueError(
    120             "supplied checkout_target does not appear "
    121             + "to be a valid reference for this repository"
    122         )
    123 else:
    124     return ls_remote.split()[0]

ValueError: supplied checkout_target does not appear to be a valid reference for this repository
```

TomNicholas:
Is there supposed to be a branch called `marbl_improvements_20240611`? I don't see it here: https://github.com/dafyddstephenson/ucla-roms/branches

TomNicholas:
Changing `checkout_target` to `main` does work.

[...]

Issue #15, opened 2024-08-15
Discretization info as separate class


TomNicholas:
The discretization info is conceptually a separate set of choices from choosing a component or input datasets. It should be defined using a separate class (e.g. a small dataclass), then passed in only when necessary

dafyddstephenson:
I agree that it would make some sense to bundle these into a class, but disagree that it is a separate set of choices - it is a part of the make-up of a component. Changing the grid discretisation or time step would change the model output like changing the surface forcing dataset would. Different components of the same case can also have, e.g., different time steps. I can see e.g. a ROMSDiscretization object being an argument to ROMSComponent much like a ROMSBaseModel instance currently is.

In any case _most_ of the info contained in the discretisation should _not_ need to be specified by the user in the future. For example, the domain having 100 levels shouldn't need to be specified in the blueprint; that information either exists in the metadata of the grid file or in the roms-tools recipe, and C-Star should, long-term, be obtaining it from there, ideally before/without downloading anything or calling to roms-tools.

PR #16, opened 2024-08-15
Fix bug in Component.__str__() identified in #14; update repo urls an…

dafyddstephenson:
…d checkout targets in examples per #12

### DIFF
--- a/cstar_ocean/component.py
+++ b/cstar_ocean/component.py
@@ -101,7 +101,6 @@ class Component(ABC):
         base_str += "\n" + "-" * (len(name) + 7)
 
         # Attrs
-        base_str += f"\ntime_step: {self.time_step} seconds"
         base_str += "\nBuilt from: "
 
         NAC = 0 if self.additional_code is None else 1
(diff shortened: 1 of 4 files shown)


examples/cstar_blueprint_roms_marbl_example.yaml
```diff
@@ -11,8 +11,8 @@ components:
         source_repo: 'https://github.com/marbl-ecosys/MARBL.git'
         checkout_target: 'marbl0.45.0'
       additional_code:
-        source_repo: 'https://github.com/dafyddstephenson/roms_marbl_example.git'
-        checkout_target: 'cstar_latest'
+        source_repo: 'https://github.com/CWorthy-ocean/cstar_blueprint_roms_marbl_example'
+        checkout_target: '69ee87b554082c46e3f138a9b3faf17439966e90'
```
TomNicholas, line 15: Is this supposed to change? Shouldn't this still point to the latest version, we're just about to update the latest version?

cstar_ocean/component.py
```diff
@@ -118,30 +117,33 @@ def __str__(self):
...
+            disc_str += "\nnx:" + str(self.nx)
         if hasattr(self, "ny") and self.ny is not None:
-            base_str += "\nny:" + str(self.ny)
+            disc_str += "\nny:" + str(self.ny)
         if hasattr(self, "exe_path") and self.exe_path is not None:
-            base_str += "\n\nIs compiled: True"
-            base_str += "\n exe_path: " + self.exe_path
+            disc_str += "\n\nIs compiled: True"
+            disc_str += "\n exe_path: " + self.exe_path
+        if len(disc_str)>0:
```
TomNicholas, line 144: Regardless of how the inputs are constructed (https://github.com/CWorthy-ocean/C-Star/issues/15) this should be a separate class internally, so that this becomes something like

```python
return base_str + str(self.discretization)
```

If this works though we can deal with that in a separate PR.

examples/cstar_blueprint_roms_marbl_example.yaml
TomNicholas: If this is fine then happy to merge this, with the understanding that we come back to it to solve #15.

dafyddstephenson: While I was working on v0 I had a fluid checkout target (the branch `cstar_latest`) so I could stay on top of regular changes I was making to the blueprint repo and C-Star. I think that's no longer necessary and it's more in line with C-Star design philosophy to have a static checkout target that is known to work (currently the latest merge on [`main`](https://github.com/CWorthy-ocean/cstar_blueprint_roms_marbl_example/commit/69ee87b554082c46e3f138a9b3faf17439966e90)). I think changes to the blueprint repo will now be infrequent enough that we can verify and update the target accordingly.

TomNicholas: That makes sense - but then presumably this line should point to either "main" or a tag, rather than a specific commit?

If this is intentional though we can merge this fix now and talk about this in a another issue.

dafyddstephenson: There's not a tag yet as I still anticipate a "first final version" once we have the proper BGC datasets locked in. `main` is not fixed. The commit hash is a temporary swap-in for what would be a tag.

cstar_ocean/component.py
dafyddstephenson: Agree - the restructuring here isn't in response to #15 but just for consistency in the case where no discretization info is present.

Issue #9, opened 2024-07-24
Use pathlib


TomNicholas:
> There are a lot of places here that would be neater if we used [pathlib](https://docs.python.org/3/library/pathlib.html). That would also help us if we later want to generalize to using [cloudpathlib](https://cloudpathlib.drivendata.org/stable/).

_Originally posted by @TomNicholas in https://github.com/CWorthy-ocean/C-Star/pull/24#discussion_r1673064819_

PR #30, opened 2024-08-17
Support local paths for `InputDataset.source`

### LINKED ISSUE #4: support local paths to input datasets
Currently the `source` attribute on `Inputdataset` only supports URLs to remotely hosted files, while `AdditionalCode` is described by a `source_repo`. We should support `source` strings that are local paths.

### LINKED ISSUE #18: bash tree: command not found
When running the example notebook
```
%%bash
tree $roms_marbl_caseroot
```
```
bash: line 1: tree: command not found
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
Cell In[14], line 1
----> 1 get_ipython().run_cell_magic('bash', '', 'tree $roms_marbl_caseroot\n')

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/interactiveshell.py:2541](http://localhost:8888/lab/tree/examples/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/interactiveshell.py#line=2540), in InteractiveShell.run_cell_magic(self, magic_name, line, cell)
   2539 with self.builtin_trap:
   2540     args = (magic_arg_s, cell)
-> 2541     result = fn(*args, **kwargs)
   2543 # The code below prevents the output from being displayed
   2544 # when using magics with decorator @output_can_be_silenced
   2545 # when the last Python token in the expression is a ';'.
   2546 if getattr(fn, magic.MAGIC_OUTPUT_CAN_BE_SILENCED, False):

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/magics/script.py:155](http://localhost:8888/lab/tree/examples/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/magics/script.py#line=154), in ScriptMagics._make_script_magic.<locals>.named_script_magic(line, cell)
    153 else:
    154     line = script
--> 155 return self.shebang(line, cell)

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/magics/script.py:315](http://localhost:8888/lab/tree/examples/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/IPython/core/magics/script.py#line=314), in ScriptMagics.shebang(self, line, cell)
    310 if args.raise_error and p.returncode != 0:
    311     # If we get here and p.returncode is still None, we must have
    312     # killed it but not yet seen its return code. We don't wait for it,
    313     # in case it's stuck in uninterruptible sleep. -9 = SIGKILL
    314     rc = p.returncode or -9
--> 315     raise CalledProcessError(rc, cell)

CalledProcessError: Command 'b'tree $roms_marbl_caseroot\n'' returned non-zero exit status 127.
```
Instead of trying to use a bash command (that apparently isn't installed in all systems by default), can we use some python library that we can verify is installed instead?

### LINKED ISSUE #21: simplest example should run really fast
I'm running the roms_marble example on my mac and it's been running for a few minutes and still hasn't completed. I think the simplest example we have should run in literally seconds. It doesn't need to be scientifically interesting, it just needs to generate some data of the right dimensionality.

### LINKED ISSUE #22: ls command in example notebook assumes you are Daffydd!
This cell in the example notebook obviously doesn't work on my machine:

`ls /Users/dafyddstephenson/Code/my_c_star/cstar_ocean/roms_marbl_example_cstar_case/output/`

### LINKED ISSUE #23: Missing dependencies for analysis
When running the final cell of the roms_marbl example notebook:

```python
BGC_DS=xr.open_dataset(roms_marbl_case.caseroot+'/output/MARBL_bgc.20120103120000.nc')
```
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[29], line 2
      1 ######################################################
----> 2 BGC_DS=xr.open_dataset(roms_marbl_case.caseroot+'[/output/MARBL_bgc.20120103120000.nc](http://localhost:8888/output/MARBL_bgc.20120103120000.nc)')
      3 var='DOC'
      4 BGC_DA=BGC_DS[var].isel(s_rho=99)

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/xarray/backends/api.py:569](http://localhost:8888/lab/tree/examples/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/xarray/backends/api.py#line=568), in open_dataset(filename_or_obj, engine, chunks, cache, decode_cf, mask_and_scale, decode_times, decode_timedelta, use_cftime, concat_characters, decode_coords, drop_variables, inline_array, chunked_array_type, from_array_kwargs, backend_kwargs, **kwargs)
    566     kwargs.update(backend_kwargs)
    568 if engine is None:
--> 569     engine = plugins.guess_engine(filename_or_obj)
    571 if from_array_kwargs is None:
    572     from_array_kwargs = {}

File [~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/xarray/backends/plugins.py:197](http://localhost:8888/lab/tree/examples/~/miniconda3/envs/cstar_environment/lib/python3.12/site-packages/xarray/backends/plugins.py#line=196), in guess_engine(store_spec)
    189 else:
    190     error_msg = (
    191         "found the following matches with the input file in xarray's IO "
    192         f"backends: {compatible_engines}. But their dependencies may not be installed, see:\n"
    193         "https://docs.xarray.dev/en/stable/user-guide/io.html \n"
    194         "https://docs.xarray.dev/en/stable/getting-started-guide/installing.html"
    195     )
--> 197 raise ValueError(error_msg)

ValueError: found the following matches with the input file in xarray's IO backends: ['netcdf4', 'h5netcdf']. But their dependencies may not be installed, see:
https://docs.xarray.dev/en/stable/user-guide/io.html 
https://docs.xarray.dev/en/stable/getting-started-guide/installing.html
```

This is probably because it looks like netcdf4-python is not installed.

Generally we should probably just ship a full conda environment that includes all the necessary python dependencies.

dafyddstephenson:
closes #4 .
Summary of changes:

Core:
- add `_get_source_type` method in `utils` (returns 'url' or 'path')
- add call to `_get_source_type` in `InputDataset.__init__()` and set `local_path` and `exists_locally` attributes if `source` attribute is a local path
- add logic to `InputDataset.get()` which creates a local symbolic link to the dataset path if it is elsewhere on the system
- update `InputDataset.check_exists_locally()` to reflect above changes

CI:
- add a second section to `tests/test_roms_marbl_example.py`. The first section creates and runs the case `roms_marbl_remote_case` where all input datasets are URLs, creating a blueprint `test_blueprint.yaml` as before. The second section first moves all the fetched datasets to an unrelated directory and then modifies the `test_blueprint.yaml` file to replace URLs with the path to this dir, before creating and running another case `roms_marbl_local_case` from this modified yaml file.

Other/bugfixes:
- Caught several problems with `Case.persist()` and `Case.from_blueprint()` when modifying the test routine above:
    -  `valid_end_date` was overwriting `valid_start_date` entry in blueprint output
    - `start_date` and `end_date` associated with input datasets were not being written to blueprint output
    - `Case.check_is_setup()` was not correctly checking for the local presence of additional code
- Updated example notebook to:
    - check for presence of `tree` on system before calling it (closes #18)
    - Advise user to modify `Case.end_date` if they want quick results (closes #21)
    - remove references to paths specific to my laptop (closes #22)
- Update `ci/environment.yml` to include `netCDF4` and `xarray` (closes #23) and remove `nco` and `ncview` (issue not yet raised but was mentioned in [Tom's notes](https://docs.google.com/document/d/1FQSHn456TaXwvKKL_LIdAoGHcAcXrpii4CAGAVNTwbg/edit))


cstar_ocean/input_dataset.py
```diff
@@ -48,17 +49,23 @@ def __init__(
...
         self.base_model: BaseModel = base_model
+
         self.source: str = source
         self.file_hash: str = file_hash
+
         self.exists_locally: Optional[bool] = None
         self.local_path: Optional[str] = None
+        if _get_source_type(source) == "path":
+            self.exists_locally = True
+            self.local_path = source
```
TomNicholas, line 67: Again I wonder if this could all be abstracted behind a class. e.g.
```python
self.source = DataSource(source)
```
then later use `self.source.exists_locally` and `self.source.path`.

cstar_ocean/input_dataset.py
```diff
@@ -91,35 +98,59 @@ def __str__(self):
...
+    def get(self, local_dir: str):
         """
-        Fetch the file containing this input dataset and save it to `local_path` using Pooch.
+        Make the file containing this input dataset available in `local_dir/input_datasets`
 
-        This method updates the `local_path` attribute of the calling InputDataset object
+        If InputDataset.source is...
+           - ...a local path: create a symbolic link to the file in `local_dir/input_datasets`.
+           - ...a URL: fetch the file to `local_dir/input_datasets` using Pooch
+                       (updating the `local_path` attribute of the calling InputDataset)
```
TomNicholas, line 108: Yeah this is another argument for abstracting out the concept of a "source" into a standalone class / type. 

This is a similar idea to using `cloudpathlib`, but maybe we need to attach more info, in which case we would need our own `Source` class with a `.path` attribute.

cstar_ocean/input_dataset.py
```diff
@@ -91,35 +98,59 @@ def __str__(self):
...
-        to_fetch = pooch.create(
-            path=tgt_dir,
-            base_url=os.path.dirname(self.source),
-            registry={os.path.basename(self.source): self.file_hash},
-        )
-
-        to_fetch.fetch(os.path.basename(self.source), downloader=downloader)
-        self.exists_locally = True
-        self.local_path = tgt_dir + "/" + os.path.basename(self.source)
+        # If the file is somewhere else on the system, make a symbolic link where we want it
```
TomNicholas, line 120: This could all go behind a `Source` interface.

TomNicholas (commented):
Nice. I have one main comment about the idea of abstracting out the "source" of the data. It's related to https://github.com/CWorthy-ocean/C-Star/issues/9 and https://github.com/drivendataorg/cloudpathlib/issues/455.

cstar_ocean/input_dataset.py
dafyddstephenson: Yes, I began implementing this for this PR but it got hairier than expected. I think it requires a lot of design considerations and should be returned to once the core feature set is complete. `AdditionalCode` and `BaseModel` also have a `source_repo` attribute (that I _think_ is about to be renamed `source` to accommodate local files) that could be replaced with an instance of whatever this class will be. I think this could be part of the tidy-up that brings in `pathlib`. I'll raise an issue for now.

TomNicholas (approved):

PR #34, opened 2024-08-20
Add Apache 2.0 license

### LINKED ISSUE #33: License
We should probably license the C-Star repo as part of the alpha release.

TomNicholas:
- [x] Closes #33

@matt-long is this definitely the licence we want to use for C-Star?


NoraLoose:
Is this good to merge @matt-long?

NoraLoose:
In our most recent conversation with @aliciak-cworthy we talked about using an Apache-2.0 license. Should we move forward with that?

dafyddstephenson:
I think that'll be fine - are you ok to take that on?

aliciak-cworthy:
@NoraLoose @dafyddstephenson 
<Consider this your official cover to proceed with the Apache 2.0 license> :)

TomNicholas:
Great thanks @aliciak-cworthy ! I'll just change this PR now to add the Apache-2.0 license.

LICENSE
```diff
@@ -0,0 +1,191 @@
...
+
+To apply the Apache License to your work, attach the following boilerplate
+notice, with the fields enclosed by brackets "[]" replaced with your own
+identifying information. (Don't include the brackets!) The text should be
+enclosed in the appropriate comment syntax for the file format. We also
+recommend that a file or class name and description of purpose be included on
+the same "printed page" as the copyright notice for easier identification within
+third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
NoraLoose: This line should be changed to
```
Copyright 2024 [C-Worthy]
```

TomNicholas: Presumably the copyright holder is [C]Worthy the company (@aliciak-cworthy)?
```suggestion
   Copyright 2024 [C]Worthy LLC.
```

TomNicholas: good catch, see https://github.com/CWorthy-ocean/C-Star/pull/34/files#r1834715608

TomNicholas: Actually no wait, if you read the context this part of the license is a set of instructions for how to add some text to other code, it's not something that needs to be replaced in this file.

That would also explain why xarray's version of this license also doesn't replace this line.

TomNicholas: Actually nevermind, see https://github.com/CWorthy-ocean/C-Star/pull/34#discussion_r1834721429

TomNicholas: I think what we are supposed to do here is leave the `LICENSE.txt` file unchanged, but copy this section with your suggested alteration and put it in the README

https://github.com/pydata/xarray/blob/a78f45a9c33b9fc9c10a69453e11722eb972cb92/README.md?plain=1#L113

NoraLoose: I don't agree. To me, it reads as if you are supposed to change this line. I think xarray just missed this.

Earthmover did this for icechunk as well:
https://github.com/earth-mover/icechunk/blob/main/LICENSE#L189

NoraLoose: > To apply the Apache License to your work, attach the following boilerplate
> notice, with the fields enclosed by brackets "[]" replaced with your own
> identifying information. (Don't include the brackets!)

NoraLoose: I can see how you can interpret this both ways. 
1. To only replace the brackets when you add this line **somewhere else**, or
2. To replace it also in the license so that other devs can simply copy from the license.

dafyddstephenson: I don't think it could hurt to do both?

TomNicholas: > To me, it reads as if you are supposed to change this line.

To me this reads as "you should copy-paste this section, changing this line, and put it somewhere prominent elsewhere in your code (e.g. the readme)", not "change this line in-place here".

> Earthmover did this for icechunk as well:

That's weird - Icechunk doesn't "attach" the boilerplate notice anywhere. Somebody is doing it wrong 😆 

> I don't think it could hurt to do both?

I agree, especially as it seems that [Android itself](https://github.com/android/architecture-samples) just changes it in both places - I think we should follow that example.

dafyddstephenson: > Somebody is doing it wrong 😆

This sets the stage for a daring caper 🥷

TomNicholas: I've updated the PR to change it in both places

NoraLoose:
Looks good to me! Feel free to merge.

PR #38, opened 2024-08-20
Move pre-commit-config to top-level directory

TomNicholas:


TomNicholas:
I think this was done already in https://github.com/CWorthy-ocean/C-Star/commit/e081da1b7e3426aabcd229d059ea1652302d878d

PR #41, opened 2024-08-20
Separate ABCs

### LINKED ISSUE #36: Separate out ROMS/Marbl-specific classes
Right now we have the abstract base class `BaseModel` and the the concrete implementations `ROMSBaseModel` and `MarblBaseModel` living in the same `cstar_ocean.base_model.py`. Instead we should have the abstract classes live in one file, and the implementations live in a separate module(s). 

One reason to do this is making it easier to understand what would be needed later in order to generalize to supporting other components, e.g. Oceananigans.

Another reason to do this is to make it clearer what the imports and dependencies are for the ABCs vs the actual concrete implementations.

TomNicholas:
- [x] Closes #36

All the internal ABCs are now imported through `cstar_ocean.base` (which could be some other word), but the point is they are separate from the concrete implementations in `cstar_ocean.roms` and `cstar_ocean.marbl`.


cstar_ocean/__init__.py
```diff
@@ -1,41 +1,8 @@
...
-    "ModelGrid",
-    "TidalForcing",
-    "BoundaryForcing",
-    "SurfaceForcing",
-    "InitialConditions",
-    "Component",
-    "ROMSComponent",
-    "MARBLComponent",
     "Case",
 ]
```
TomNicholas, line 8: I removed most of these on the basis that 
- many users will only need to import `Case`, 
- all ROMS/Marbl-specific imports should be behind a dedicated corresponding namespace, 
- ABCs also won't need to be imported by users, only by developers of new components

cstar_ocean/case.py
```diff
@@ -304,9 +304,13 @@ def from_blueprint(
...
                 case "roms":
+                    from cstar_ocean.roms import ROMSBaseModel, ROMSComponent
+
                     ThisComponent = ROMSComponent
                     ThisBaseModel = ROMSBaseModel
                 case "marbl":
+                    from cstar_ocean.marbl import MARBLBaseModel, MARBLComponent
+
                     ThisComponent = MARBLComponent
                     ThisBaseModel = MARBLBaseModel
```
TomNicholas, line 317: Notice the `Case` now only imports roms- and marbl-specific classes when it needs them.

cstar_ocean/case.py
```diff
@@ -18,6 +16,9 @@
     SurfaceForcing,
 )
 
+if TYPE_CHECKING:
+    from cstar_ocean.base import Component, BaseModel
```
TomNicholas: There are a lot of imports that are actually only needed for type hints. Putting these behind the `if TYPE_CHECKING` statement avoids circular imports, and makes it clearer when you actually need to import the class constructor and when you are only dealing with pre-instantiated instance of that class.

TomNicholas: (... but if you do this you need to use `"<ClassName>"`, not `<ClassName>`, which I forgot)

TomNicholas:
The new directory structure looks like this

Screenshot 2024-08-20 at 6 47 58 PM

TomNicholas:
It's not obvious to me whether `base/` should include `additional_code.py` and `InputDataset`, but at least what I have done here groups everything that is not specific to roms or marbl into one separate directory.

EDIT: Okay my decision to put `InputDataset` in `base` is vindicated by https://github.com/CWorthy-ocean/C-Star/pull/43#issue-2476617135

cstar_ocean/case.py
TomNicholas: Also FYI it goes against python conventions to use `CamelCase` for variable names, even if those variables are pointing to types. So it should be e.g. `this_component_cls = ROMSComponent`. I only didn't change it in this PR to make it less confusing because I'm already changing a lot of names.

TomNicholas:
@dafyddstephenson there aren't too many lines changed in the example notebook

```bash
(cstar_environment) tom@Toms-MBP C-Star % git diff --stat main:examples/cstar_example_notebook.ipynb examples/cstar_example_notebook.ipynb
 examples/cstar_example_notebook.ipynb | 392 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------
 1 file changed, 241 insertions(+), 151 deletions(-)
```

So I think this should be okay to merge now?

Issue #32, opened 2024-08-19
`source` as instance of dedicated class


dafyddstephenson:
Currently `source` (or `source_repo` in some cases) is a `str` containing a URL (or [soon](https://github.com/CWorthy-ocean/C-Star/pull/30), a local path). Meanwhile every class with a `source`-like attribute has a `get()` method. As we support more flexibility (e.g. local files, roms-tools integration), `source` should be an instance of a dedicated class, rather than `str`. This class should have logic for handling remote vs. local, different file types, repos, etc., and could likely also encapsulate a single `get()` method rather than this being defined separately for multiple different classes.

PR #43, opened 2024-08-20
ROMS-specific `InputDataset` classes; introduce `DataSource` class

dafyddstephenson:
This PR mostly focuses on restructuring `InputDataset` subclasses. In particular:
- `InputDataset` is now an ABC 
- there is now a `ROMSInputDataset`ABC that subclasses `InputDataset`
- All other `InputDataset` subclasses have been renamed to reflect their ROMS association, (e.g. `ModelGrid` -> `ROMSModelGrid`) and now subclass `ROMSInputDataset`
- There is now a check in `Case.from_blueprint` whereby if an input dataset entry is not `None` and is not a `ROMSInputDataset`, a `NotImplementedError` is raised

Additionally, this PR introduces a `DataSource` class that made the above changes easier. 

`DataSource` objects are initialised from a `location` string (e.g. `https://github.com/CWorthy-ocean/input_datasets_roms_marbl_example/raw/main/roms_frc.201201.nc`) and maintain this under the `location` attribute while also having non-settable properties based on parsing this string:
- `location_type` (here "URL", can currently also be "path")
- `source_type` (here "netcdf", can currently also be "yaml" or "repository" though the latter is as-yet unused)
- `basename` (here "roms_frc.201201.nc") 

This addresses #32 but this issue should remain open as we should consider how to improve this class in future: verifying file types rather than just parsing their extensions from a path string; applying it to other classes like `AdditionalCode` and `BaseModel` that have `source_repo` attributes; etc.

edit, 2024-08-22: have updated `test_roms_marbl_example.py` to remove any leftover files from previous executions (addresses #46 but should remain open for now)


cstar_ocean/input_dataset.py
```diff
@@ -193,41 +273,61 @@ def check_exists_locally(self, local_dir: str) -> bool:
...
+        """
+    ROMS-specific implementation of `InputDataset` (doc below)
+
+    Extends `get()` method to generate dataset using roms-tools in the case that `source`
+    points to a yaml file.
+
+    Docstring for InputDataset:
+    ---------------------------
+    """
+    ) + (InputDataset.__doc__ or "")
```
TomNicholas: Are you trying to edit the docstring of the `ROMSInputDataset` class here? Because you can access `__doc__` directly, so I think you can do

```python
ROMS_DATASET_DOC = """..."""

class ROMSInputDataset(InputDataset, ABC):
    __doc__ = InputDataset.__doc__ + ROMS_DATASET_DOC
```

cstar_ocean/cstar_case.py
```diff
@@ -10,12 +10,13 @@
...
-    ModelGrid,
-    InitialConditions,
-    TidalForcing,
-    BoundaryForcing,
-    SurfaceForcing,
+    ROMSModelGrid,
+    ROMSInitialConditions,
+    ROMSTidalForcing,
+    ROMSBoundaryForcing,
+    ROMSSurfaceForcing,
```
TomNicholas: These don't need to imported until we know that ROMS is one of the components of the Case, see https://github.com/CWorthy-ocean/C-Star/pull/41#discussion_r1724033961

cstar_ocean/input_dataset.py
dafyddstephenson: Neat. I'm now remembering that I meant to ask you about how you'd go about this as PEP-257 isn't very clear. Almost all the information is contained in the `InputDataset` docstring _but_ the user will never actually be handling an `InputDataset` directly now it's an ABC, and we can imagine them basically only ever handling `ROMSInputDataset` types. It seems a bit daft to write a docstring that says "go look at this other docstring that you'd never otherwise encounter, this is basically the same except for the following differences". OTOH just dumping the contents of that parent docstring into this one (as I've done here) doesn't feel very clean either. Any ideas?

cstar_ocean/input_dataset.py
```diff
@@ -193,41 +273,61 @@ def check_exists_locally(self, local_dir: str) -> bool:
         return self.exists_locally
 
 
-class ModelGrid(InputDataset):
+class ROMSInputDataset(InputDataset, ABC):
```
TomNicholas: It's not entirely clear to what the point of differentiating between a `ROMSInputDataset` and an `InputDataset` is, as there is no common logic in `ROMSInputDataset`. Are you anticipating there being some logic that is relevant for all input datasets that are needed for ROMS, but not for input datasets needed for non-ROMS components?

dafyddstephenson: Yes I anticipate that in future this will be needed. Tools-Roms (the fortran ones) currently operate on any and all things covered by `ROMSInputDataset`, for example. I think if anything that `InputDataset` is the redundant class here

dafyddstephenson:
@TomNicholas  Merged #41 in but could you review all the imports etc. and check it's consistent with what you were setting out to do? I think the main issue here is that there is now an ABC in `roms` (`ROMSInputDataset`). We  could reasonably do away with this for now per your comment above and subclass e.g. `ROMSModelGrid(InputDataset)` but I'm quite convinced this will rear its head in future (e.g. methods that are necessary for all ROMS input files but make no sense outside the context of ROMS)

dafyddstephenson:
Have now also merged in #47  and updated `test_roms_marbl_example.py` in response to #46 (see comment there)

cstar/base/input_dataset.py
```diff
@@ -1,27 +1,103 @@
...
+        else:
+            raise ValueError(
+                f"{os.path.splitext(self.location)[-1]} is not a supported file type"
+            )
+
+    @property
+    def basename(self) -> str:
+        """Get the basename (typically a file name) from the location attribute"""
+        return os.path.basename(self.location)
+
```
TomNicholas, line 81: This suggestion isn't really important, but it demonstrates the difference in how `__repr__` is recommended to be used vs how `__str__` is recommended to be used. Ideally `call(repr(obj)) == obj`, whereas `__str__` can be more verbose.

```suggestion
def __repr__(self) -> str:
    return f"DataSource(location={self.location})"

```

tests/test_roms_marbl_example.py
```diff
@@ -37,9 +51,9 @@
     url_prefix="https://github.com/CWorthy-ocean/input_datasets_roms_marbl_example/raw/main/"
     if url_prefix in line:
         fileurl=line.split()[-1] # Just isolate URL from e.g. source: URL
-        filepath=f"{os.getcwd()}/local_input_files/{os.path.basename(fileurl)}"
-        if os.path.exists(filepath):
-            test_blueprint[i]=line.replace(fileurl,filepath)
+        filepath=Path(f"{Path.cwd()}/local_input_files/{Path(fileurl).name}")
```
TomNicholas, line 54: `pathlib.Path` understands the `/` operator, so you can literally just do this:
```suggestion
        filepath = Path.cwd() / "local_input_files" / Path(fileurl).name
```

tests/test_roms_marbl_example.py
```diff
@@ -1,9 +1,23 @@
-import os
+import cstar
 import shutil
-import subprocess
+from pathlib import Path
```
TomNicholas, line 3: This is the way

TomNicholas:
> I think the main issue here is that there is now an ABC in roms (ROMSInputDataset).

I think this is fine. There is a roms-specific ABC in `.roms`, that isn't going to be used anywhere else. That seems appropriate.

cstar_ocean/input_dataset.py
TomNicholas: I've seen this problem before but don't know of a neater solution.

> just dumping the contents of that parent docstring into this one (as I've done here) doesn't feel very clean either

This is probably fine imo.

TomNicholas (approved):

tests/test_roms_marbl_example.py
dafyddstephenson: just learned about this today, quite cool.

PR #35, opened 2024-08-20
Remove underscores from module name (i.e. `cstar_ocean` -> `cstar`)

TomNicholas:
- [x] Closes #7

PR #47, opened 2024-08-21
Remove underscores from module name (i.e. `cstar_ocean` -> `cstar`)

### LINKED ISSUE #7: Module naming conventions
> the `cstar_*` prefixes here are unnecessary, we should change the names of the modules to be short and simple, e.g. `cstar.base`, `cstar.inputs`.

_Originally posted by @TomNicholas in https://github.com/CWorthy-ocean/C-Star/pull/24#discussion_r1673048712_

TomNicholas:
Supercedes #35, by insteading building on top of #41. (So should not be merged until after #41 is merged)

- [x] Closes #7


TomNicholas:
Still need to update the `README.md`

TomNicholas:
This should be good now too.

I wasn't totally sure if we need to change anything else in `setup.py` but we can worry about that when we actually publish it to PyPI (#48)

dafyddstephenson:
Thanks @TomNicholas . I think we can keep underscores in module files on grounds of "improves readability" and would be consistent with `Case` attributes - so if we can move `basemodel.py` back to `base_model.py` (to reflect `Case.base_model`) this can go straight in by the looks.

EDIT: have merged this branch into #43 and applied the name change there instead. Merging.

PR #52, opened 2024-08-22
Better input dataset hash handling

### LINKED ISSUE #51: `file_hash` is required for `InputDataset` even when `source` is local
`InputDataset.file_hash` should be optional and default to `None`. `InputDataset.get()` should then throw an error if `file_hash` is not provided in the case that `InputDataset.source.location_type` is "url".

dafyddstephenson:
Closes #51 . Branched off #43 - do not merge until after merging #43.

This PR makes `InputDataset.file_hash` an optional attribute but throws an error in `InputDataset.__init__()` if `source.location_type` is "url" and `file_hash` is None. It also updates `Case.from_blueprint` to use `dict.get` rather than attempting to directly access potentially nonexistent keys for optional attributes (such as `file_hash`).

This behaviour improves consistency following #30  .

### DIFF
--- a/cstar/case.py
+++ b/cstar/case.py
@@ -381,8 +381,8 @@ class Case:
                     model_grid = [
                         ROMSModelGrid(
                             base_model=base_model,
-                            source=DataSource(f["source"]),
-                            file_hash=f["hash"],
+                            source=DataSource(f.get("source")),
+                            file_hash=f.get("hash", None),
                         )
                         for f in input_dataset_info["model_grid"]["files"]
                     ]
@@ -397,10 +397,10 @@ class Case:
                     initial_conditions = [
                         ROMSInitialConditions(
                             base_model=base_model,
-                            source=DataSource(f["source"]),
-                            file_hash=f["hash"],
-                            start_date=f["start_date"],
-                            end_date=f["end_date"],
+                            source=DataSource(f.get("source")),
+                            file_hash=f.get("hash", None),
+                            start_date=f.get("start_date", None),
+                            end_date=f.get("end_date", None),
                         )
                         for f in input_dataset_info["initial_conditions"]["files"]
                     ]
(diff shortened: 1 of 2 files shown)


dafyddstephenson:
TIL backticks in commit messages make words disappear...
"[make Optional with ValueError if None and is](https://github.com/CWorthy-ocean/C-Star/pull/52/commits/e8bdf4358eff4d94ab06eed6cef40cd32cd2180f)" = "Make `InputDataset.file_hash` Optional with ValueError if None and `InputDataset.source.location_type` is `'url'`"

dafyddstephenson:
@TomNicholas Hoping to put this in so I can update the changelog in #55 (branched off this one). Any comments?

cstar/case.py
```diff
@@ -381,8 +381,8 @@ def from_blueprint(
                     model_grid = [
                         ROMSModelGrid(
                             base_model=base_model,
-                            source=DataSource(f["source"]),
-                            file_hash=f["hash"],
+                            source=DataSource(f.get("source")),
+                            file_hash=f.get("hash", None),
```
TomNicholas, line 385: What is the type of `f` here? Couldn't we just change `f.get` to return `None` if it doesn't have a hash instead of adding this `, None`?

dafyddstephenson: This is a list comprehension that ruff made a bit screwy looking:
`for f in input_dataset_info["model_grid"]["files"]` is at the end. `f` is a dictionary in this case from parsing this section of the blueprint yaml file:
```
      input_datasets:
        model_grid:
          files:
            - location: 'https://github.com/CWorthy-ocean/input_datasets_roms_marbl_example/raw/main/roms_grd.nc'
              hash: 'fd537ef8159fabb18e38495ec8d44e2fa1b7fb615fcb1417dd4c0e1bb5f4e41d'

```
`.get` here is a python builtin for cleaner dictionary access, the `,None` part is doing exactly what you say: no error if the key isn't present, `get` just returns `None`.

TomNicholas: Okay, I wasn't sure if the type we were calling `.get` on was a `DataSource` or something. But if it's a `dict` this seems fine.

TomNicholas (approved):

Issue #46, opened 2024-08-21
Error running test


TomNicholas:
```
Traceback (most recent call last):
  File "/Users/tom/Documents/Work/Code/C-Star/tests/test_roms_marbl_example.py", line 56, in <module>
    roms_marbl_local_case.setup()
  File "/Users/tom/Documents/Work/Code/C-Star/cstar/case.py", line 711, in setup
    inp.get(self.caseroot)
  File "/Users/tom/Documents/Work/Code/C-Star/cstar/base/input_dataset.py", line 131, in get
    raise FileExistsError(
FileExistsError: A file by the name of roms_grd.ncalready exists at /Users/tom/Documents/Work/Code/C-Star/roms_marbl_local_case/input_datasets/ROMS/.
```

I ran the tests again locally and got an error. I think this is related to both using pytest https://github.com/CWorthy-ocean/C-Star/issues/42 (as we can use pytest `tmpdir` to clean up after our tests run), and also potentially https://github.com/CWorthy-ocean/C-Star/issues/17 (functions should perhaps be designed so that they don't care if they have already been run before).

dafyddstephenson:
Yes, related to pytest, not 17 though - this error is triggered by `test_roms_marbl_example.py` attempting to move files around rather than anything internal to C-Star. I could add something at the start of the test to just delete any files created by previous runs?

TomNicholas:
> not 17 though

Are you sure? It literally happens when we call `.setup()`.

> `test_roms_marbl_example.py` attempting to move files around rather than anything internal to C-Star.

Why does it move files around? Why does it do anything outside of C-Star?

> I could add something at the start of the test to just delete any files created by previous runs?

pytest's `tmdir` fixture can be used to automatically delete any files created during the running of that test, which is neater (cleaning up immediately is neater than cleaning up only on the next time the test runs).

TomNicholas:
In fact we may want to think more broadly about structuring the building/running of cstar in such a way that all files created are only created in new directories (which can be temporary directories, to make cleaning up easier).

dafyddstephenson:
> Are you sure? It literally happens when we call `.setup()`.

[...]

PR #54, opened 2024-08-26
Use pytest and temporary directories

TomNicholas:
WIP to use pytest for the tests.

EDIT: I'm trying to change our tests from a run-once python script to an actual extensible testing framework. In particular that means tests should fully clean up after themselves automatically, and example blueprint generation should be extensible and robust.


TomNicholas:
When I run `pytest` in the root I get this error:

```python
E               OSError: System environment variable 'MARBL_ROOT' points toa github repository whose remote: 
E                '' 
E               does not match that expected by C-Star: 
E               https://github.com/marbl-ecosys/MARBL.git.Your environment may be misconfigured.

cstar/base/basemodel.py:221: OSError
```

dafyddstephenson:
This is most likely because you deleted your `externals` and didn't delete `cstar_local_config.py` which tracked the (now non-existent) location of MARBL and ROMS on your system

TomNicholas:
That gets me further, thanks!

NoraLoose:
What's the status on this? Let me know if I can help @TomNicholas

cstar/tests/test_roms_marbl_example.py
```diff
@@ -0,0 +1,102 @@
...
+                 "test_blueprint.yaml",
+                 "modified_test_blueprint.yaml",
+                 "test_blueprint_local.yaml"]:
+    
+    oldpath=Path(oldfiles)
+    if oldpath.exists():
+        if oldpath.is_dir():
+            shutil.rmtree(oldpath)
+        else:
+            oldpath.unlink()
```
TomNicholas: We also want cstar to be robust to these files existing / not existing, which means we want to test with and without automatic deletion.

@dafyddstephenson are these directories all configurable at runtime in cstar? Because if so I could make a `tmpdir` for each of them and pass that in.

cstar/tests/test_roms_marbl_example.py
```diff
@@ -0,0 +1,102 @@
...
+    blueprint_filepath = tmp_path / 'blueprint.yaml'
+
+    with open(blueprint_filepath, 'w') as file:
+        yaml.dump(blueprint_dict, file)
+
+    return blueprint_filepath
+
+
+# TODO similar fixture but the blueprint file contains paths to local files instead
+# TODO should these local files be in a given temporary directory?
```
TomNicholas: I'm not entirely sure what the best way to handle blueprints is, but I think we should change the current approach of modifying a blueprint to use machine-specific local filepaths.

The point of this fixture was supposed to be so that we could start with one blueprint and modify it in-memory, then write the modified blueprint back out to a temporary directory, then pass the (temporary) path to that file back into the test.

However it would be better to just have a set of pre-defined blueprint files to read from. If paths in blueprints can be relative paths, then perhaps we can avoid having to make the blueprint specific to a machine?

cstar/tests/test_roms_marbl_example.py
dafyddstephenson: > @dafyddstephenson are these directories all configurable at runtime in cstar? 

Not sure I understand what you mean by this, let me know if this helps: The directory names are set by the calls to `Case.from_blueprint` later in the script using the `caseroot` argument. C-Star creates the directory if it doesn't already exist. You could create a temporary directory and use it as long as you supply it as a `Path` to `Case.from_blueprint()` using the `caseroot` argument.

cstar/tests/test_roms_marbl_example.py
dafyddstephenson: > However it would be better to just have a set of pre-defined blueprint files to read from.

I'm not strictly against having two separate blueprints (one local, one remote) if that's what you mean here. However, I deliberately avoided this as these two files would differ only by a single substring throughout. Any change to the remote blueprint thus needs to be also tracked in the local version (this could be a lot if we switch all the input datasets, for example) whereas currently we only need to track the two substrings (base URL, base filepath). 

> I'm not entirely sure what the best way to handle blueprints is, but I think we should change the current approach of modifying a blueprint to use machine-specific local filepaths.

This is definitely fine for testing, and any user is free to make blueprints with relative paths, but I would maintain the behaviour that, within C-Star, and in any blueprints it writes out, absolute paths are used (to maintain the most exact information available about the location of a file). C-Star-produced blueprints should not break when moved from one directory to another because of the use of relative paths. Thus if we want to use relative paths in a blueprint describing local files _produced by C-Star_, we should have to manually modify it.

dafyddstephenson: I just stumbled on this discussion #46 which is similar. I forgot another reason for recycling the same blueprint was to test C-Star's ability to create a case from a C-Star-created blueprint. This is something we should definitely test if we aim to prepare separate local/remote blueprints. It may also be worth planning around that this will soon double to local/remote x netcdf/yaml ?

cstar/tests/test_roms_marbl_example.py
TomNicholas: > You could create a temporary directory and use it as long as you supply it as a Path to Case.from_blueprint() using the caseroot argument.

Okay great, that's exactly what I wanted to know (and what I'm already doing in this PR!)

cstar/tests/test_roms_marbl_example.py
TomNicholas: > I'm not strictly against having two separate blueprints (one local, one remote) if that's what you mean here.

That was my second suggestion yes.

> However, I deliberately avoided this as these two files would differ only by a single substring throughout. Any change to the remote blueprint thus needs to be also tracked in the local version (this could be a lot if we switch all the input datasets, for example)

Yeah, that's why I made this yaml fixture that reads the blueprint into memory. We could alter the yaml structure at this point, then parametrize that and auto-generate the different blueprints we need as we need them. The result would be similar to what you're already doing but hopefully less error-prone because the base blueprint is unchanging and the per-test blueprints are created (and deleted) as needed.

> C-Star-produced blueprints should not break when moved from one directory to another because of the use of relative paths.

That's fair.

>  I forgot another reason for recycling the same blueprint was to test C-Star's ability to create a case from a C-Star-created blueprint.

We should test this, but in a dedicated test. I'm trying to think how to structure all this so that we don't just end up with a small number of mega-tests.

> It may also be worth planning around that this will soon double to local/remote x netcdf/yaml ?

This is why I want to be able to test one thing per test with the blueprints etc. auto-generated as needed. Then we can just add more `@pytest.mark.parametrize` decorators to tests and we will cover all the possibilities.

dafyddstephenson: Thanks for explaining, this makes a lot more sense to me now. Keeping a single yaml dict in memory then modifying it between tests is a very clean way to handle it, nice.

cstar/tests/test_roms_marbl_example.py
TomNicholas: I think this cleaning up works now 🍾

TomNicholas: To clarify, right now if I run pytest on a freshly-cloned cstar then the only things that are still around after the test are `cstar_local_config.py` and `cstar/externals`. I think the latter is unavoidable, but the former can be removed via refactoring hopefully (or at least leave only a machine-specific configuration `.yml` file).

The test passes, and passes even if I run it again with no other actions.

cstar/tests/test_roms_marbl_example.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _blueprint_as_path(name: str, make_local: bool = False) -> Path:
+        # TODO add options to edit in-memory blueprints here?
+
+        blueprint_dict = example_blueprint_as_dict(name)
+
+        # TODO define `make_local`function
+        if make_local:
+            # blueprint_dict = make_local(blueprint_dict)
+            raise NotImplementedError()
```
TomNicholas: @dafyddstephenson do you think this approach will work? Is the difference between your two tests representable like this? I want the `make_local` function to make the same changes that you do here:

https://github.com/CWorthy-ocean/C-Star/blob/c83b1a1ad8599df0ee2879afb7f4feac42c3fa21/tests/test_roms_marbl_example.py#L58

TomNicholas: Also how am I supposed to know what the value of these needs to be for any given test?

```python
id_url_prefix = "https://github.com/CWorthy-ocean/input_datasets_roms_marbl_example/raw/main/"
ac_url = "https://github.com/CWorthy-ocean/cstar_blueprint_roms_marbl_example.git"
```

dafyddstephenson: I'm not really sure what's going on here I don't really see why not represent that change as a function, it's just a string replacement (url -> path).

dafyddstephenson: they're in the blueprint, those are just the strings to replace. If you're only working with one core blueprint, then these won't change.

TomNicholas: > why not represent that change as a function

I was just wondering if that was sufficient.

> If you're only working with one core blueprint, then these won't change.

Okay, sounds like I can hard code these values for now and revisit this in a follow-up to generalize it.

TomNicholas: I've generalized these by putting them all in a top-level global `EXAMPLE_BLUEPRINTS` dict.

cstar/tests/test_roms_marbl_example.py
TomNicholas: This generation and modification of the yaml as a dict in-memory is now implemented.

TomNicholas:
The CI failure is https://github.com/CWorthy-ocean/C-Star/issues/66 (separate from this PR).

conftest.py
```diff
@@ -0,0 +1,151 @@
...
+    # TODO
+    # blueprint_with_local_additional_code = ...
+
+
+def process_nested_dict(
+    dictionary: dict,
+    condition: Callable[[str, str], bool],
+    update: Callable[[str, str], str],
+    parent_key: str = "",
+):
```
TomNicholas: This is meant to be a general-purpose function for altering in-memory blueprint yamls.

conftest.py
```diff
@@ -0,0 +1,151 @@
...
+
+# Prevent errors caused by pytest trying to collect tests from externals 
+# (e.g. from errors whilst importing python code in the UCLA-roms repo, see https://github.com/CWorthy-ocean/C-Star/issues/53)
+collect_ignore_glob = [
+    "cstar/cstar_local_config.py",
+    "cstar/externals/**"
+]
+
+
+EXAMPLE_BLUEPRINTS = {
```
TomNicholas: I feel like this list of example blueprints should live somewhere other than in `conftest.py`.

conftest.py
```diff
@@ -0,0 +1,151 @@
...
+    "ROMS_MARBL": {
+        # TODO this assumes you are running pytest from the root directory of the cloned repo, which is fragile
+        "base": "./examples/cstar_blueprint_roms_marbl_example.yaml",
+        "input_datasets_url_prefix": "https://github.com/CWorthy-ocean/input_datasets_roms_marbl_example/raw/main/",
+        "additional_code_url": "https://github.com/CWorthy-ocean/cstar_blueprint_roms_marbl_example.git",
```
TomNicholas: Do we actually have to ship these additional urls along with the blueprint? Or can they always be determined from the blueprint?

conftest.py
```diff
@@ -0,0 +1,151 @@
...
+    blueprint_with_local_input_datasets = process_nested_dict(
+        dictionary=blueprint,
+        condition=is_a_location_containing_input_datasets_url,
+        update=update_path_to_input_datasets,
+    )
+
+    return blueprint_with_local_input_datasets
+
+    # TODO
+    # blueprint_with_local_additional_code = ...
```
TomNicholas: My main question with this PR is what is going on with local additional code? 

I expected to have to implement this function here in order for the test to pass, but it already passes without me having implemented this function. Does that mean this test doesn't actually test the use of local additional code at all?

conftest.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@pytest.fixture
+def example_blueprint_as_path(
+    example_blueprint_as_dict, tmp_path
+) -> Callable[[str], Path]:
+    """Given the name of a pre-defined blueprint, returns it as a (temporary) path to an on-disk file."""
+
+    def _blueprint_as_path(name: str, make_local: bool = False) -> Path:
+        # TODO add options to edit in-memory blueprints here?
```
TomNicholas: `make_local` is an example of one such option.

cstar/tests/test_roms_marbl_example.py
```diff
@@ -0,0 +1,36 @@
...
+            # TODO do we actually need user input for all these steps?
+            roms_marbl_case.setup()
+
+            # TODO why are we persisting this blueprint file then not using it again in the test?
+            roms_marbl_case.persist(tmpdir / "test_blueprint.yaml")
+
+            roms_marbl_case.build()
+            roms_marbl_case.pre_run()
+            roms_marbl_case.run()
+            roms_marbl_case.post_run()
```
TomNicholas: As an aside, I would like us to have tests for each of these steps individually.

conftest.py
dafyddstephenson: they can be determined from the blueprint

TomNicholas: Okay. But how? I mean I see that that `input_datasets_url_prefix` exists as a substring of the `location` fields of each of the files under `input_datasets`. But is that actually a parameter that I must derive and then pass? Or can I just say "any input dataset file needs to have its location changed to be local", in which case I don't actually need to know what the url prefix is in order to modify that paths.

conftest.py
TomNicholas: Turns out the local test was passing but for the wrong reasons. The dict was not being edited correctly, and the pre-existing remote paths were being kept around, so the second test ended up basically being equivalent to the first test.

Obviously our tests should not pass if they behave unexpectedly! A better solution (as suggested by @dafyddstephenson ) might be to start with an invalid blueprint that contains placeholders of somekind, and fill it in with valid paths for both tests. Rather than starting from a valid input and changing it to another valid input. That way if the replacement step fails the input will be invalid, and the test fail.

dafyddstephenson: Thanks for noting this down - it may be worth adding assert statements for things like `{AdditionalCode,InputDataset}.source.location_type` to verify the modification has gone ahead also

TomNicholas:
We decided that we really need a way to generate local test data on demand in order to perform tests with local data. Now that roms-tools is integrated with C-Star, we can do that using roms-tools. This could take the form of a monster fixture that creates input data ready for C-Star to use.

Given the size of that effort, it should happen in a different PR to this one. But we also don't want to leave this PR dangling. But if we merge it in we don't want to delete existing tests, as currently this PR has smaller coverage.

So @dafyddstephenson and I decided to merge basically what is here but without altering the existing tests. There will be an awkward crossover period with two independent sets of tests existing in `main` at once, but that seems like the least worst way to do it.

.github/workflows/tests.yaml
```diff
@@ -45,8 +45,9 @@ jobs:
         shell: micromamba-shell {0}        
         run: |
            python -V
+           coverage run --rcfile=coverage.toml -m pytest --verbose cstar/tests/*
            python -u -m coverage run --rcfile=coverage.toml tests/test_roms_marbl_example.py
```
TomNicholas: This PR now runs both the old and the new tests.

TomNicholas:
@dafyddstephenson I think this might be ready to merge now? It runs both sets of tests in the CI successfully, and you can run either one independently locally. #99 can be addressed in a follow-up PR.

.github/workflows/tests.yaml
dafyddstephenson: Any chance we could separate these into two separate steps in the workflow? Hard to tell where one ends and the next begins

TomNicholas: Fair point. I actually should probably just make them separate CI jobs.

dafyddstephenson:
Sorry I should have taken a better look before your last commit - I think having `tests` and `cstar/tests` is a little confusing, perhaps renaming `tests` to `tests_old` in line with the CI job names or something. Other than that looks good to go.

TomNicholas:
Moved to `tests_old`. The additional changes are just from `pre-commit` / to shut mypy up.

dafyddstephenson (approved):

PR #55, opened 2024-08-29
Local additional code

### LINKED ISSUE #31: Support local paths to additional code
This one is a bit trickier than #4 :
The current implementation of the AdditionalCode class assumes that all code associated with a blueprint is coming from a single repository. Take a look at the relevant section in [`cstar_blueprint_roms_marbl_example`](https://github.com/CWorthy-ocean/C-Star/blob/b332842a727ee0843373e80093dd83014fa0bcdf/examples/cstar_blueprint_roms_marbl_example.yaml#L33C1-L48C46):

```
      additional_code:
        source_repo: 'https://github.com/CWorthy-ocean/cstar_blueprint_roms_marbl_example'
        checkout_target: '69ee87b554082c46e3f138a9b3faf17439966e90'
        source_mods:
          - "source_mods/ROMS/bgc.opt"
          - "source_mods/ROMS/bulk_frc.opt"
          - "source_mods/ROMS/cppdefs.opt"
          - "source_mods/ROMS/diagnostics.opt"
          - "source_mods/ROMS/ocean_vars.opt"
          - "source_mods/ROMS/param.opt"
          - "source_mods/ROMS/tracers.opt"
          - "source_mods/ROMS/Makefile"
          - "source_mods/ROMS/get_makefile"
          - "source_mods/ROMS/Make.depend"
        namelists:
          - "namelists/ROMS/roms.in_TEMPLATE"
```

It seems weird to require that if the user wants to use local files that these would have to be git controlled, so the class will have to be restructured a bit.

The main issue here is that a single `AdditionalCode` object currently represents a list of files, whereas, e.g. `InputDataset` is one per file. I think there should be some distinction in how we handle these two classes, as it makes a lot more sense for a bunch of, e.g. source code modifications to co-exist in a single place under one `source` header than it does for a bunch of disparate forcing datasets.

Any ideas? @TomNicholas @matt-long @NoraLoose

dafyddstephenson:
-> This PR branches from #52 , do not merge until merging #52 <-
---
Closes #31.
## Summary
This PR introduces support for local additional code, as well as improving the handling of model namelists that are to be modified by C-Star (the latter changes being necessary to test the former). Local `AdditionalCode` instances should be created using a path to a directory, instead of a remote repository, but the structure of this directory should be the same:
```                                                                                                                                                                                                                                                                                                   
    <additional_code_dir>                                                                                                                                                                                                                                                                          
       ├── namelists                                                                                                                                                                                                                                                                               
       |      └ <base_model_name>                                                                                                                                                                                                                                                                  
       |              ├ <namelist_file_1>                                                                                                                                                                                                                                                          
       |              |       ...                                                                                                                                                                                                                                                                  
       |              └ <namelist_file_N>                                                                                                                                                                                                                                                          
       └── source_mods                                                                                                                                                                                                                                                                             
              └ <base_model_name>                                                                                                                                                                                                                                                                  
                      ├ <source_code_file_1>                                                                                                                                                                                                                                                       
                      |       ...                                                                                                                                                                                                                                                                  
                      └ <source_code_file_N>      
```
Meanwhile, calls to `AdditionalCode.get()` now check whether namelist filenames have `_TEMPLATE` as a suffix. If so, a new copy without this suffix is created alongside the template file. The template is tracked with `AdditionalCode.namelists` as before, the copy is tracked using a new attribute `AdditionalCode.modified_namelists`. This is an improvement on previous handling but not a viable long term solution. I have raised Issue #56  in response.

## List of changes
### Changes directly required for local additional code support:
-  Move `DataSource` out of `base/input_dataset.py` and into dedicated module `base/datasource.py`
- Expand `DataSource` to check if location is a directory or local repository
- Rename `AdditionalCode.source_repo`:
   - `source_repo` attribute is renamed to `AdditionalCode.source`, type changed from `str` -> `DataSource`
   - `source_repo` argument to `AdditionalCode.__init__()` is renamed to `location` (`str`). 
   This location is used by `__init__` to create a `DataSource` instance and set the `AdditionalCode.source` attribute.
- Similarly rename `source` and `source_repo` to `location` in:
   -  `Case.from_blueprint()`
   - `Case.persist()`
   - `cstar_blueprint_roms_marbl_example.yaml` 
- Rename "local_path" to "local_dir" in AdditionalCode.get to avoid confusion with AdditionalCode.local_path
- Make `AdditionalCode.checkout_target` optional (in case source is not a repository)
- Update `AdditionalCode.get()`. This previously cloned a repository to a temporary location and then moved the files needed by the `AdditionalCode` instance. It now does this only if `AdditionalCode.source.location_type` is `url` and `AdditionalCode.source.source_type` is `repository`. If `location_type` is `path` then the files are simply copied from the provided path.
- The `test_roms_marbl_example.py` script now also uses local additional code as well as local input datasets in its second (local) stage 

### Related changes associated with namelist handling:
These changes were necessary to implement the test described above, as previously the template namelist file was edited in place and so couldn't be re-used by a second case.
- Add an attribute `AdditionalCode.modified_namelists` to keep track of edited namelists based on templates
- Add logic in `AdditionalCode.get()` that makes copies of any namelist files with the suffix `_TEMPLATE` to a file without this suffix and adds it to `AdditionalCode.modified_namelists`, e.g.:
`AdditionalCode.namelists = ["roms.in_TEMPLATE",]` -> `AdditionalCode.modified_namelists = ["roms.in",]`
- Update `ROMSComponent.pre_run()` and `ROMSComponent.run()` to reflect this namelist handling

### Other/bugfixes:
- Update InputDataset.local_path to point to the symlink created by InputDataset.get() after call (if dataset is local, addressing an inconsistency in #52 )
- update pre-commit config file to rename cstar_ocean -> cstar
- swap out os for pathlib in additional_code.py and datasource.py 
- minor changes to case.py with additional type checking
- Update docstrings, `__str__`s and example notebook to reflect above changes


TomNicholas (approved):
I'm looking through this and so far it looks good. We should probably chat about #56 at some point though.

cstar/base/additional_code.py
```diff
@@ -73,81 +88,139 @@ def __init__(
...
+                    # Special case for template namelists:
+                    if (
+                        file_type == "namelists"
+                        and str(src_file_path)[-9:] == "_TEMPLATE"
+                    ):
+                        print(
+                            f"copying {tgt_file_path} to editable namelist {str(tgt_file_path)[:-9]}"
+                        )
+                        shutil.copy(tgt_file_path, str(tgt_file_path)[:-9])
+                        if hasattr(self, "modified_namelists"):
```
dafyddstephenson, line 214: I think this needs an else to create the attribute if it's missing

cstar/base/datasource.py
```diff
@@ -0,0 +1,82 @@
...
+    def location_type(self) -> str:
+        """Get the location type (e.g. "path" or "url") from the "location" attribute"""
+        urlparsed_location = urlparse(self.location)
+        if all([urlparsed_location.scheme, urlparsed_location.netloc]):
+            return "url"
+        elif Path(self.location).exists():
+            return "path"
+        else:
+            raise ValueError(
+                f"{self.location} is not a recognised URL or local path pointing to an existing file"
```
dafyddstephenson: *or directory

cstar/base/additional_code.py
dafyddstephenson: done in [a1a5bb6](https://github.com/CWorthy-ocean/C-Star/pull/55/commits/a1a5bb6bfef5c90986f060b629eb7faf0f6b8834)

cstar/base/datasource.py
dafyddstephenson: done in [a1a5bb6](https://github.com/CWorthy-ocean/C-Star/pull/55/commits/a1a5bb6bfef5c90986f060b629eb7faf0f6b8834)

PR #60, opened 2024-09-04
Setting up the documentation

NoraLoose:
This PR sets up a sphinx-based documentation that supports both markdown / [MyST](https://myst-parser.readthedocs.io/en/latest/) and ReStructuredText (RST).

### Notes
1. I ~~moved~~ copied the notebook `cstar_example_notebook.ipynb` from the `examples` directory to the `docs` directory so that we can include that notebook in the documentation. (It can't sit outside the `docs` directory.) I removed the uppermost markdown cell with the caption "Installation" from the notebook because 
  * we have a separate installation section in the docs
  * both "Installation" and "Introduction to C-Star “v0” with the roms_marbl_example Case" are top-level headings ("#"), but sphinx wants there only to be one top-level heading, otherwise the formatting is off
2. I added the sphinx-based dependencies to `ci/environment.yml` so that we can test the documentation build as part of the CI
3. I added a test that builds the documentation to the GitHub workflows

### Still to do in future PRs

- [x]  Hook the documentation up with ReadTheDocs (RTD)
- [ ]  Reformat some of the docstrings (there are lots of warnings raised when building the API reference). This should be handled by `pre-commit`, so I suggest adding `pre-commit` to the CI and `cstar_env`. (I'm happy to do this in a follow-up PR).
- [ ] Refactor tests to use `pytest` because that's what the Contributor Guide (`docs/contributing.md`) promises.


NoraLoose:
@dafyddstephenson does either 1. or 2. in my notes above raise any red flags for you?

NoraLoose:
> @dafyddstephenson does either 1. or 2. in my notes above raise any red flags for you?

If not, I suggest to merge so that I can test whether RTD can successfully deploy the documentation.

docs/index.rst
```diff
@@ -0,0 +1,30 @@
+Welcome to the C-Star Documentation!
+====================================
+
+**C-Star** is a python package that orchestrates the execution of a complete start-to-finish run of ROMS-MARBL.
+
```
dafyddstephenson: I think I'd suggest something like

```
**C-Star** is a python package for setting up and running ocean model simulations, with a particular focus on marine carbon dioxide removal (mCDR) applications.
```

... we could add that its current capability is limited to ROMS-MARBL?

NoraLoose: Sounds good. Changed in `3cce721`.

NoraLoose:
There is still some work to do for future PRs (I will document these in issues), but this is good to merge for now.

PR #63, opened 2024-09-04
Change python version for RTD docs build

NoraLoose:
RTD fails to build the documentation with this error:

Screenshot 2024-09-04 at 12 24 19 PM

I haven't seen the weird `virtualenv` error when building the `roms-tools` docs, so I'm switching to the same python version that we are using in the ROMS-Tools `.readthedocs.yaml`

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,18 +9,14 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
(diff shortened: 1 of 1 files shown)


NoraLoose:
Let's hope this fixes the problem and makes the docs badge light up green! 🍏 No way to know before I merge...

PR #71, opened 2024-09-05
Discretization class

### LINKED ISSUE #15: Discretization info as separate class

dafyddstephenson:
Closes #15 

This PR creates a `Discretization` class in `base.component` and a `ROMSDiscretization` subclass in `roms.component`. In the resulting clean-up, the following additional changes were made:

- Massive simplification of the `Case.from_blueprint` method:
   - by ensuring consistency between blueprint entries and constructor arguments we can pass dictionaries read from the blueprint directly to constructor functions as kwargs
   - instantiating different `InputDataset` subclasses in a loop over yaml dictionary keys rather than checking for them all individually removes a large amount of redundant/repetitive code
- Made the `Component` constructor args explicit (formerly `def __init__(self, **kwargs: Any)`)


cstar/base/component.py
```diff
@@ -37,7 +40,13 @@ class Component(ABC):
         Execute any post-processing actions associated with this component
     """
 
-    def __init__(self, **kwargs: Any):
+    def __init__(
+        self,
+        base_model: BaseModel,
+        additional_code: Optional["AdditionalCode"],
+        input_datasets: List[InputDataset] = [],
```
TomNicholas: Technically mutable default arguments should be [avoided](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil).

Instead this should be

```python
def __init__(
    ...
    input_datasets: List[InputDataset] | None = None,
    ...
):
    if input_datasets is None:
        input_datasets = []
```

I'm actually surprised the linter didn't point this out.

cstar/base/component.py
```diff
@@ -158,3 +138,66 @@ def post_run(self) -> None:
...
+    """
+    Holds discretization information about a Component.
+
+    Attributes:
+    -----------
+
+    time_step: int
+        The time step with which to run ROMS in this configuration
+    nx,ny,n_levels: int
+        The number of x and y points and vertical levels in the domain associated with this object
```
TomNicholas: It might break docs formatting to not list these properly

```suggestion
    nx: int
        The number of x points in the domain
    ny: int
        The number of y points in the domain
    n_levels: int
        The number of vertical levels in the domain
```

cstar/base/component.py
```diff
@@ -158,3 +138,66 @@ def post_run(self) -> None:
...
+        --------
+        Discretization:
+            An initialized Discretization object
+
+        """
+
+        self.time_step: int = time_step
+        self.nx: Optional[int] = nx
+        self.ny: Optional[int] = ny
+        self.n_levels: Optional[int] = n_levels
```
TomNicholas: Surely this information is redundant with the `Grid` object?

cstar/roms/component.py
```diff
@@ -70,17 +61,14 @@ class ROMSComponent(Component):
...
 
+    discretization: "ROMSDiscretization"  # mypy misses type hint in constructor and assumes Discretization|None
+
     def __init__(
         self,
         base_model: "ROMSBaseModel",
-        time_step: int,
         additional_code: AdditionalCode,
+        discretization: "ROMSDiscretization",
         input_datasets: Optional[List["InputDataset"]] = [],
```
TomNicholas, line 71: This also has a mutable default.

cstar/roms/component.py
```diff
@@ -471,3 +442,80 @@ def post_run(self) -> None:
...
+
+class ROMSDiscretization(Discretization):
+    """
+    An implementation of the Discretization class for ROMS.
+
+
+    Additional attributes:
+    ----------------------
+    n_procs_x,n_procs_y: int
+        The number of processes following the x and y directions, for running in parallel
```
TomNicholas: Formatting and clarity nitpick

```suggestion
    n_procs_x,
        The number of parallel processors over which to subdivide the x axis of the domain.
    n_procs_y: int
        The number of parallel processors over which to subdivide the y axis of the domain.
```

cstar/roms/component.py
```diff
@@ -471,3 +442,80 @@ def post_run(self) -> None:
...
+        self.n_procs_x = n_procs_x
+        self.n_procs_y = n_procs_y
+
+    @property
+    def n_procs_tot(self) -> Optional[int]:
+        """Total number of processors required by this ROMS configuration"""
+        if (self.n_procs_x is None) or (self.n_procs_y is None):
+            return None
+        else:
+            return self.n_procs_x * self.n_procs_y
```
TomNicholas: How can this be `None`? Surely if you're not running in parallel the default for `n_procs_x`, `n_procs_y`, and `n_procs_tot` would be `1`, not `None`?

cstar/base/component.py
```diff
@@ -46,32 +55,26 @@ def __init__(self, **kwargs: Any):
...
             An intialized Component object
         """
-
-        # TODO: do Type checking here
-        if "base_model" not in kwargs or not isinstance(
-            kwargs["base_model"], BaseModel
-        ):
-            raise ValueError(
-                "base_model must be provided and must be an instance of BaseModel"
-            )
```
TomNicholas: You seem to have lost these instance checks. We could put them in property setters, but only if properties are supposed to be settable. We could autogenerate the `isinstance` checks by using a `dataclass`, or we could just write them out here.

TomNicholas (commented):
I like the abstraction, and it clearly simplifies some things, but I have some questions about the data model for the new class.

cstar/base/component.py
dafyddstephenson: This was actually to silence the linter - it wouldn't stop complain that None is not iterable, even when it's explicitly set to a list immediately. I think the above suggestion is implicit optional which will also wind the linter up.
This should work:
```
def __init__(
    ...
    input_datasets: Optional[List[InputDataset]]
    ...
):
   if input_datasets is None:
      input_datasets = []
```
... but if I remember correctly whenever mypy saw Optional it required an explicit check for `None` before doing anything listy with `input_datasets`, even though it's never `None`. The issue with the mutable default is dealt with a few lines later:

`self.input_datasets: List[InputDataset] = list(input_datasets)`

I can't remember the specifics but this was a nightmare and this outcome took a good while to land on as a solution. Would you be up for having a crack at it?

TomNicholas: > This was actually to silence the linter - it wouldn't stop complain that None is not iterable, even when it's explicitly set to a list immediately.

That's annoying, I don't know why mypy doesn't just narrow down what the type must be.

I think when I've had this problem before the simplest solution was just to do

```python
def __init__(
    ...
    input_datasets: Optional[List[InputDataset]]
    ...
):
    if input_datasets is None:
        _input_datasets = []
    else:
        _input_datasets = input_datasets
```
and then use `_input_datasets` only for the rest of the method.

TomNicholas: or on one line:
```python
_input_datasets = [] if input_datasets is None else input_datasets
```

TomNicholas: Reading into this, I'm pretty sure my suggestion is the least worst way. Unfortunately I think this is one of those ugly mypy things that is a consequence of static typing being added to python post-hoc, not when the language was invented.

dafyddstephenson: I'll give it a go, but I'm not sure this will address the underlying issue: IIRC the rest of the method was the only place where mypy _was_ happy with the typing - it was everywhere else that `Component.input_datasets` got used throughout the codebase... I'd have to revisit this again to be sure though.

dafyddstephenson: ... I think it was what I've committed here vs. `assert isinstance(Component.input_datasets, list)` in every other method that used it...

TomNicholas: `Component.input_datasets` is a different variable to the `input_datasets` being used inside this init method. That sounds like a different issue. You can specify what the type of `Component.input_datasets` should be like this:

```python
class Component:
    input_datasets: list[InputDataset]

    def __init__(self, ...):
        ...
```

> ... I think it was what I've committed here vs. `assert isinstance(Component.input_datasets, list)` in every other method that used it...

Though I'm not totally sure I understand what you're saying here.

dafyddstephenson: Right, I might even be confusing memories of mypy's various whinings here, so I think I'm too out of the loop to say anything about this with confidence without revisiting it for now.

TomNicholas: Your line below

```python
self.input_datasets: List[InputDataset] = list(input_datasets)
```
should achieve the same thing though... But if you type hint it in the class definition you don't need to type hint it during assignment.

dafyddstephenson: I actually meant to ask you to keep a general eye out for this and was going to raise an issue to broadly revisit and tidy up type hints - I think I've over-hinted throughout and used `Optional` in a way that probably isn't needed but makes mypy much stricter...

TomNicholas: Let me have a go at cleaning this up and pushing to this branch now

TomNicholas: Should be fixed in https://github.com/CWorthy-ocean/C-Star/pull/71/commits/c2c311cf751d2a3a9f2019a5c5fd1444e9823856

cstar/base/component.py
dafyddstephenson: Addressed in  b69800f70ca3a583ffdde906d663d847df97404b

cstar/roms/component.py
dafyddstephenson: Addressed in b69800f70ca3a583ffdde906d663d847df97404b

cstar/base/component.py
dafyddstephenson: reinstated this in b69800f70ca3a583ffdde906d663d847df97404b

cstar/roms/component.py
dafyddstephenson: Good point! Addressed in b69800f70ca3a583ffdde906d663d847df97404b

cstar/base/component.py
dafyddstephenson: It actually wasn't being used anyway so I've just taken it out in b69800f70ca3a583ffdde906d663d847df97404b

cstar/base/component.py
```diff
@@ -158,3 +141,49 @@ def post_run(self) -> None:
...
+class Discretization:
+    """
+    Holds discretization information about a Component.
+
+    Attributes:
+    -----------
+
+    time_step: int
+        The time step with which to run ROMS in this configuration
+    """
```
TomNicholas:
```suggestion
    time_step: int
        The time step with which to run the Component
    """
```

cstar/base/component.py
```diff
@@ -158,3 +141,49 @@ def post_run(self) -> None:
         This abstract method will be implemented differently by different Component types.
         """
         pass
+
+
+class Discretization:
```
TomNicholas: Shouldn't this be an ABC too?
```suggestion
class Discretization(ABC):
```

cstar/roms/component.py
```diff
@@ -114,23 +97,10 @@ def __init__(
...
         self.exe_path: Optional[Path] = None
 
-    @property
-    def n_procs_tot(self) -> Optional[int]:
-        """Total number of processors required by this ROMS configuration"""
-        if (self.n_procs_x is None) or (self.n_procs_y is None):
-            return None
-        else:
-            return self.n_procs_x * self.n_procs_y
+        assert isinstance(self.discretization, ROMSDiscretization)
```
TomNicholas: Technically you shouldn't use `assert` outside of tests because supposedly some python compilers compile them away as an optimization. Never actually had this happen to me though.

```suggestion
        if not isinstance(self.discretization, ROMSDiscretization):
            raise TypeError()
```

cstar/roms/component.py
```diff
@@ -471,3 +442,76 @@ def post_run(self) -> None:
...
+    n_procs_y: int
+        The number of parallel processors over which to subdivide the y axis of the domain.
+
+
+    Properties:
+    -----------
+    n_procs_tot: int
+        The value of n_procs_x * n_procs_y
+
+    """
```
TomNicholas, line 463: I expect the docs formatters to complain about these extra blank lines
```suggestion
    """
    An implementation of the Discretization class for ROMS.

    Additional attributes:
    ----------------------
    n_procs_x: int
        The number of parallel processors over which to subdivide the x axis of the domain.
    n_procs_y: int
        The number of parallel processors over which to subdivide the y axis of the domain.

    Properties:
    -----------
    n_procs_tot: int
        The value of n_procs_x * n_procs_y
    """
```

TomNicholas (approved):
Nice! Just some minor suggestions, then good to go.

cstar/roms/component.py
dafyddstephenson: I know, this is really nasty, was just to avoid the linter losing the thread. I don't see why your implementation wouldn't achieve the same thing though, I'll try it and see what pre-commit does

dafyddstephenson: I just deleted it, seems fine now whatever the issue was before

cstar/roms/component.py
dafyddstephenson: I'll issue a separate PR for docstring cleanup seeing as ruff has found hundreds of errors now I've added a check

PR #78, opened 2024-09-13
Refactor of `exists_locally` (and consequently a lot of other stuff...)

### LINKED ISSUE #25: Redundancy of public `exists_locally` vs `check_exists_locally`
I feel like we don't need two public methods that do basically the same thing. Instead we might just have one read-only property `exists_locally` which internally calls a private method `_check_exists_locally`.

dafyddstephenson:
Branched from #74 , do not merge until after merging #74 

The initial purpose of this PR was a short refactor of the `exists_locally` attribute to allow the case when a yaml file describing an input dataset existed locally, but the associated netCDF file did not. It kept unspooling from there with no clear finish line. 

Ultimately, this is a lot of spring cleaning that is going to be necessary for the roms-tools integration. Changes to the user experience are fairly minimal (though have been updated in the example notebook), but what's happening behind the scenes is a lot cleaner and more intuitive, with hopefully all bodges and hacks removed.

closes #25 

## Summary:
- rename `local_path` attribute on `InputDataset` and `AdditionalCode` to `working_path` to avoid suggesting that the file(s) described by the path exist.
- Change `InputDataset.get(local_dir)` and `AdditionalCode.get(local_dir)` to actually fetch files to `local_dir`, not a convoluted subdirectory structure of it. `Case.setup()` now calls these methods with the required structure as an argument.
- Replace `check_exists_locally` on `InputDataset` and `AdditionalCode` with a property `exists_locally` which was not previously possible (and do the same for `check_is_setup` on `Case`, `check_local_config_status` on `BaseModel`)
- To make this possible, restructure of `AdditionalCode` in terms of expected directory/repo layout.
- ... Added an optional `subdir` attr/init argument to `AdditionalCode` in case the additional code lives in the subdirectory of a remote repository.
- Have changed the main example blueprint to remove MARBL additional code. Confusingly, as ROMS handles all input and output on MARBL's behalf, including reading things like namelist files, these are in fact _ROMS_ additional code. There were lots of hacky things in the code to get around this simple fact just because it didn't fit in very nicely with the C-Star design concept. They are now gone.
- The main example blueprint now points to [my fork](https://github.com/dafyddstephenson/roms_marbl_example/tree/main) of the example repo, which has been restructured, but I will update the C-Worthy version and C-Star blueprint once this PR is accepted. This was to avoid breaking tests for other people working with that repo.
- Provide explicit `output_dir` argument to `Component.run()` so it doesn't have to try and figure it out from the additional code directory structure anymore, which was ugly and confusing.


cstar/base/additional_code.py
```diff
@@ -86,39 +94,31 @@ def __init__(
...
+        base_str += "-" * (len(base_str) - 1)
         base_str += f"\nBase model: {self.base_model.name}"
         base_str += f"\nLocation: {self.source.location}"
-        if self.exists_locally is not None:
-            base_str += f"\n Exists locally: {self.exists_locally}"
-        if self.local_path is not None:
-            base_str += f"\n Local path: {self.local_path}"
+        base_str += f"\nsubdirectory: {self.subdir}"
+        base_str += f"\nWorking path: {self.working_path}"
+        base_str += f"\nExists locally: {self.exists_locally}"
```
TomNicholas: You could use a multiline string instead of `+=` to add each line.

(`textwrap.dedent` will be useful)

cstar/base/additional_code.py
```diff
@@ -86,39 +94,31 @@ def __init__(
...
+        if not self.exists_locally:
+            base_str += " (get with AdditionalCode.get())"
```
TomNicholas, line 119: This might be a little more readable
```suggestion
            base_str.append(" (get with AdditionalCode.get())")
```

cstar/base/component.py
```diff
@@ -46,39 +55,33 @@ def __init__(self, **kwargs: Any):
...
+        self.discretization: Optional[Discretization] = discretization or None
 
     def __str__(self) -> str:
         # Header
         name = self.__class__.__name__
-        base_str = f"{name} object "
-        base_str = "-" * (len(name) + 7) + "\n" + base_str
-        base_str += "\n" + "-" * (len(name) + 7)
+        base_str = f"{name}"
+        # base_str = "-" * len(name) + "\n" + base_str
```
TomNicholas:
```suggestion
```

cstar/base/component.py
```diff
@@ -87,43 +90,34 @@ def __str__(self) -> str:
...
         return base_str
 
     def __repr__(self) -> str:
-        return self.__str__()
+        repr_str = f"{self.__class__.__name__}("
+        repr_str += f"\nbase_model = <{self.base_model.__class__.__name__} instance>, "
+        if self.additional_code is not None:
+            repr_str += f"\nadditional_code = <{self.additional_code.__class__.__name__} instance>, "
+        else:
+            repr_str += "\n additional_code = None"
```
TomNicholas: Isn't there some redundant here with the other classes? i.e. you could be calling `repr(additional_code)` here.

TomNicholas (commented):
I think you've done a great job of making complex objects more understandable here :clap:

cstar/base/additional_code.py
dafyddstephenson: I tried to keep each line modular to shift things around or add more things easily - though thinking about it a multi-line wouldn't really change that. Can do if it's more readable that way!

cstar/base/component.py
dafyddstephenson: As the class structure is so hierarchical I tried to avoid a cascading printout of hundreds of lines by just calling reprs within reprs all the way down. This is a change from #74 however, would be good to discuss strs/reprs over there as that PR in particular was supposed to be a first stab at something that could become a discussion (the PR post itself also contains examples of all the strs/reprs so they're easier to read than just the code!)

dafyddstephenson:
ok some stuff got chewed up in the merges but I think we're good to go now, have updated the diffs here

cstar/base/base_model.py
```diff
@@ -294,3 +290,10 @@ def handle_config_status(self) -> None:
     @abstractmethod
     def get(self, target: str | Path) -> None:
```
TomNicholas, line 291: This is not an urgent nor a particularly helpful comment at this time, but I think a less overloaded term for this method could be `fetch`, because `get` already has nother widely-used meaning in python.

cstar/base/base_model.py
```diff
@@ -294,3 +290,10 @@ def handle_config_status(self) -> None:
...
         """clone the basemodel code to your local machine"""
+
+    @abstractmethod
+    def _base_model_adjustments(self) -> None:
+        """
+        Perform any C-Star specific adjustments to the base model that would
+        be needed after a clean checkout.
+        """
```
TomNicholas, line 299: for example?

cstar/base/input_dataset.py
```diff
@@ -88,6 +75,14 @@ def __init__(
         assert self.start_date is None or isinstance(self.start_date, dt.datetime)
         assert self.end_date is None or isinstance(self.end_date, dt.datetime)
 
+    @property
+    def exists_locally(self) -> bool:
+        return (
+            True
+            if ((self.working_path is not None) and (self.working_path.exists()))
```
TomNicholas, line 82: Pretty sure most, if not all of these brackets are redundant
```suggestion
            if (self.working_path is not None) and self.working_path.exists()
```

TomNicholas (approved):
Just some very minor comments.

cstar/base/base_model.py
dafyddstephenson: Agree, I have been pondering a better alternative for a minute - even ignoring the conflict with python's use of get, ours is used in contexts going far beyond fetching, like `BaseModel.get()` which would be much better described as `BaseModel.setup()`. Fetch is fine, but in certain situations `get()` also creates symlinks, which I wouldn't really describe as "getting" or "fetching".

cstar/base/input_dataset.py
dafyddstephenson: They are, it's personal preference as most editors will highlight the matching bracket when you hover on one - makes it easier to see the nested structure / exactly what each condition is in chains of `and`s and `or`s

cstar/base/base_model.py
TomNicholas: hmm, okay, let's punt on this. I'll raise your comment as an issue.

PR #81, opened 2024-09-14
Input datasets from roms tools

### LINKED ISSUE #3: Integrate ROMS tools
For the `InputDataset` class, we should support the ability for the `source` attribute to point not only to a pre-created netcdf file containing the dataset, but also a yaml file containing instructions for _creating_ the dataset with [ROMS-tools](https://github.com/CWorthy-ocean/roms-tools). The `file_hash` attribute could still be used to verify that the created file is identical to the one that would be sourced remotely.

Related issue: [roms-tools # 63](https://github.com/CWorthy-ocean/roms-tools/issues/63)

dafyddstephenson:
Branched from #78 , do not merge until merging #78. Closes #3 .

As this PR chain is growing, it may be worth merging each and refreshing the next until we get to this one, so GitHub isolates the changes specific to this one (commits and files changed currently shows all previous PR changes too).

### This PR:
- Adds a `get_from_yaml()` method to `ROMSInputDataset`, which is called from `Case.setup()` in the situation that a dataset being fetched is a `ROMSInputDataset` with `source.source_type` equal to `'yaml'`. 
   - The method first calls `InputDataset.get()` on the yaml file itself, as usual, making a local copy.
   - It then modifies any keys related to start and end times in line with a user-supplied start and end time (typically `Case.{start,end}_date`). In the [ROMS-MARBL example](https://github.com/dafyddstephenson/roms_marbl_example/tree/main/roms_tools_yaml_files) (and I would suggest future blueprints made for distribution) these keys are given placeholder entries in the roms-tools yaml files.
   - It then calls `<RomsToolsObject>.save()` and updates `ROMSInputDataset.working_path` to point to the local netCDF.
   - If the user supplies `np_xi` and `np_eta` arguments, it partitions directly on save and updates the `ROMSInputDataset.partitioned_files` attribute (`List[Path]`)
- Replaces the use of `Tools-Roms/partition.F` in `Component.pre_run()` with `roms_tools.utils.partition_netcdf`, and sets the `ROMSInputDataset.partitioned_files` attribute
- Adds a new blueprint `tests/cstar_blueprint_roms_tools_example.yaml` and test `tests/test_roms_tools_integration.py` (basically the same as `test_roms_marbl_example.py` but beginning from the new blueprint)
- Fixes a couple of outstanding bugs from #78 

### NOTES:
- It is not possible to run `test_roms_tools_integration.py` on GitHub as it requires access to datasets needed to generate `roms-tools` data. The roms-tools yaml files in the github repo linked above have `path` entries that are local paths _on my laptop_. Once I move to getting everything ready on perlmutter next week I'll change them to that, which is slightly better as we can all use it, but still not good enough for CI. Fixing this should be very high priority after the alpha madness is over. @NoraLoose , perhaps we can emulate what you're doing with roms-tools testing (phony data?) and make an entire ROMS Case from scratch that way.


NoraLoose:
In `ROMS-Tools` I'm fetching the raw input datasets via `pooch` from https://github.com/CWorthy-ocean/roms-tools-test-data, see for example

https://github.com/CWorthy-ocean/roms-tools/blob/8f0faac33dc62e72d37cd9e9bc88a8ea4d951004/roms_tools/tests/test_setup/conftest.py#L59

and 

https://github.com/CWorthy-ocean/roms-tools/blob/8f0faac33dc62e72d37cd9e9bc88a8ea4d951004/roms_tools/setup/download.py

I think we should do the same to test the C-Star / ROMS-Tools integration. I'm happy to write these tests next week if we have `pytest` working.

TomNicholas:
> The method first calls `InputDataset.get()` on the yaml file itself, as usual, making a local copy.

Is this a good pattern to enshrine? The whole point of separating the concept of a "case" from a "blueprint" was that the blueprint is the on-disk serialisation format for the case, but if we copy and edit files on disk we are arguably breaking that abstraction.

Might it be better to instead read the remote yaml file into an in-memory dict, edit it in-memory, then only write it out if `.persist()` is called?

TomNicholas:
Also please can we make multiple smaller PRs? If I understand correctly this one does 4 seemingly unrelated things, that could have been 4 smaller PRs. It makes it harder to review / respond to reviews.

I can see the branches building on one another is getting annoying - one way to deal with that would simply be to have key feature branches live in this upstream repo instead of in your fork, 

This is especially important now because I recently changed the repository default to squash commits rather than merge commits into main (to avoid having hundreds of trivial commits in main, and to ensure that on each commit to main the tests are not broken). But if we combine many features into one PR it makes it hard to track down commits that introduce bugs, hard to revert just one of those buggy commits, and harder to follow conversations that refer to individual commits.

dafyddstephenson:
> > The method first calls `InputDataset.get()` on the yaml file itself, as usual, making a local copy.
> 
> Is this a good pattern to enshrine? The whole point of separating the concept of a "case" from a "blueprint" was that the blueprint is the on-disk serialisation format for the case, but if we copy and edit files on disk we are arguably breaking that abstraction.

Yes, this is desired behaviour. All the objects in C-Star have a "source" copy (described by the blueprint) and a "working" copy (handled by the Case). Both are tracked by C-Star. In this instance, the yaml file describing how datasets are constructed in `roms-tools` necessarily contains things that aren't strictly bound to a blueprint - for instance, we support `Case` instances which start a simulation from any time within a valid date range specified by a blueprint, but `roms-tools` creates an `InitialConditions` object from a fixed `ini_time` parameter. We need to be able to change this parameter in the working copy of the yaml file before we give it to `roms-tools` in order to have any flexibility in running the simulation. This is much the same as how namelist files are handled: the original source material is tracked under the `source` attribute, but `get()` obtains a _working_ copy in the `caseroot` folder which C-Star can then modify _within the bounds of the blueprint_. It is up to us how much C-Star is allowed to modify the working copy, and what checks we put in to measure divergence from the source copy, but they _do_ need to be separate entities. 

> Also please can we make multiple smaller PRs? If I understand correctly this one does 4 seemingly unrelated things, that could have been 4 smaller PRs. It makes it harder to review / respond to reviews.

Sorry this is a headache. I have stuck to one PR per issue since we last visited this problem (this one closes #3 for example), but sometimes (as in the case of #78, which closes #25 ), a single issue can spiral into a large refactor. A single small commit that appears to address an issue breaks the code or introduces an inconsistency somewhere else, and I'd rather not issue a PR with broken code or inconsistencies. To address this particular comment, the 4 things are not unrelated (two of them are related to issue 3, and two of them are needed to test the other two -- again, I'd rather not issue a PR without testing it first). I think this PR will appear a lot more sensible once the others are in and the diff has been updated. 

More generally, I suspect this will no longer be a problem now we are effectively at alpha and certain principles of how C-Star works are no longer fluid but properly established. However, I'm happy to break things down further if it makes things easier - I can set a line change limit on a branch before issuing a PR from it, or issue a PR immediately from a branch and leave it as a draft while working on it, or (not my favourite option) go back and post-hoc split a branch up/rebase the code into a more granular set of smaller PRs. Open to other suggestions also, let me know what makes things easiest (while hopefully balancing the overhead to some extent, hence me not favouring the last option...)

TomNicholas:
> they do need to be separate entities.

I agree that the original blueprint and the modified version need to be separate, but I'm questioning the idea that they both need to live on disk.

> In this instance, the yaml file describing how datasets are constructed in roms-tools necessarily contains things that aren't strictly bound to a blueprint [...] We need to be able to change this parameter in the working copy of the yaml file before we give it to roms-tools in order to have any flexibility in running the simulation.

Please correct me if I've misunderstood, but to me this simply sounds like we are communicating options from `C-Star` (in python memory) to `roms-tools` (in python memory). I don't understand why communicating this requires writing anything to disk.

> This is much the same as how namelist files are handled

namelist files are different because there we are trying to communicate options from python memory (C-Star) to a fortran program (`roms` itself). Whilst it's possible to bind fortran from python, it's a pain, and writing the namelist information to disk so that `roms` can read it is hopefully the least painful way to communicate with `roms`. As `roms` does not make direct use of our blueprint yaml, we shouldn't need to write out a modified version of the entire blueprint in order to communicate with `roms`.

> what checks we put in to measure divergence from the source copy

I want to entirely avoid us having to keep track of two similar but different yaml files whilst C-Star is running! Can we not instead do something more like this?

```python
# read configuration into memory
case = Case.from_blueprint('verified_example_blueprint.yaml')

# manually alter some aspect of the case's configuration, but keeping it all in memory
altered_case = case.change_grid(...)

# running only saves the bare minimum to disk that roms needs to know about (e.g. namelists)
case.run()

# okay that ran successfully, let's now save all those options to disk for future use / provenance
case.persist_to_blueprint('altered_blueprint.yaml')
```

---

> A single small commit that appears to address an issue breaks the code or introduces an inconsistency somewhere else

IMO this should not happen, and is a sign both that we need more clear separation of responsibilities in the code (which my suggestion about in-memory options above is motivated by) and that we need more targeted tests (i.e. unit tests, not just integration tests).

> go back and post-hoc split a branch up/rebase the code into a more granular set of smaller PRs

That would be a pain! I'm not suggesting you do that for this PR.

> a single issue can spiral into a large refactor.
> 
> Open to other suggestions also, let me know what makes things easiest

In my experience what works well is as soon as you realise that a refactor is required (which happens a lot, I definitely get that), you create a new branch from `main` that only implements that refactor. If the code is well-covered by unit tests, and the tests only test user-level API, then often this refactor can be done without actually changing the tests at all (if it changed the tests it would no longer technically be merely a refactor, as instead it would imply a change in observed behaviour for the user).

Then once the refactor branch is passing again, you can go back to the original feature branch and merge in the refactor branch, and keep working on the feature. Each new feature branch _should_ change the tests (usually by adding new unit tests to test the new functionality).

> More generally, I suspect this will no longer be a problem now we are effectively at alpha and certain principles of how C-Star works are no longer fluid but properly established.

I agree. But I am worried that we are establishing more and more "principles" without codifying them explicitly as expected behaviour in the form of targeted tests.

TomNicholas:
> if we have pytest working.

I think the pytest PR (https://github.com/CWorthy-ocean/C-Star/pull/54) is working now, which means the new test added in this PR should be rewritten to use the pytest setup. (That should be quite simple.) I just have some questions over on #54.

TomNicholas:
I would really like to see what the diff for this PR looks like after the previous PR and the pytest PR are merged in - I think that will be very illuminating.

dafyddstephenson:
ok, now passing

cstar/base/input_dataset.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         self.base_model: "BaseModel" = base_model
         self.source: DataSource = DataSource(location)
         self.file_hash: Optional[str] = file_hash
-        self.working_path: Optional[Path] = None
+        self.working_path: Optional[Path | List[Path]] = None
```
TomNicholas, line 61: Since python 3.9 you have been able to use the built-in collections for typing, so you can use `list` instead of `typing.List`
```suggestion
        self.working_path: Optional[Path | list[Path]] = None
```

cstar/marbl/component.py
```diff
@@ -5,13 +5,13 @@ class MARBLComponent(Component):
...
     def build(self) -> None:
-        print("source code modifications to MARBL are not yet supported")
+        pass
 
     def pre_run(self) -> None:
-        print("no pre-run actions involving MARBL are currently supported")
+        pass
 
     def run(self) -> None:
         print("MARBL must be run in the context of a parent model")
```
TomNicholas, line 14: Surely if this method is not intended to be called then?
```suggestion
        raise AttributeError("MARBL must be run in the context of a parent model")
```

cstar/roms/component.py
```diff
@@ -172,37 +174,73 @@ def pre_run(self) -> None:
 
             namelist_forcing_str = ""
             # Partition input datasets and add paths to namelist
+            if len(datasets_to_partition) > 0:
+                from roms_tools.utils import partition_netcdf
             for f in datasets_to_partition:
-                dspath = f.working_path
-                fname = f.source.basename
+                # fname = f.source.basename
```
TomNicholas, line 180: Doesn't look like this is used now
```suggestion
```

cstar/roms/component.py
```diff
@@ -172,37 +174,73 @@ def pre_run(self) -> None:
...
-                )
+                parted_files = []
+                for idfile in id_files_to_partition:
+                    parted_files += partition_netcdf(
+                        idfile,
+                        np_xi=self.discretization.n_procs_x,
+                        np_eta=self.discretization.n_procs_y,
+                    )
+
+                    # [p.rename(partdir / p.name) for p in parted_files[-1]]
```
TomNicholas, line 219:
```suggestion
```

cstar/roms/component.py
```diff
@@ -172,37 +174,73 @@ def pre_run(self) -> None:
...
+
                 # Namelist modification step
                 if isinstance(f, ROMSModelGrid):
-                    namelist_grid_str = f"     {partdir / dspath.name} \n"
+                    if f.working_path is None or isinstance(f.working_path, list):
+                        raise ValueError(
+                            f"ROMS only accepts a single grid file, found list {f.working_path}"
+                        )
+
+                    assert isinstance(f.working_path, Path), "silence, linter"
```
TomNicholas, line 231: What is this about 😆

cstar/roms/input_dataset.py
```diff
@@ -16,18 +22,146 @@ class ROMSInputDataset(InputDataset, ABC):
...
+                yaml_dict[roms_tools_class_name][key] = value
+
+        with open(yaml_file, "w") as F:
+            F.write(f"---{header}---\n" + yaml.dump(yaml_dict))
+
+        # Finally, make a roms-tools object from the modified yaml
+        import roms_tools
+
+        roms_tools_class = getattr(roms_tools, roms_tools_class_name)
+        roms_tools_class_instance = roms_tools_class.from_yaml(yaml_file)
```
TomNicholas, line 142: This is what I want to see removed to solve #84. (But it doesn't need to be done in this PR)

PR #85, opened 2024-09-16
EXERCISE C notebook

NoraLoose:
In this exercise we make a blueprint from scratch, which consists of four steps:

- [x]  Preparing the input data for a ROMS simulation (using `ROMS-Tools`)
- [x] Building a C-Star `Case`, which includes gathering
    * ROMS source model code (`cstar.base.BaseModel` and `cstar.base.AdditionalCode`)
    * ROMS input data from step 1 (`cstar.base.InputDataset`)
- [x]  Compiling and running the ROMS simulation (using `C-Star`)
- [x]  Making a `C-Star` blueprint so that steps 1, 2, and 3 are reproducible

The regridding step in `ROMS-Tools` is currently very inefficient - in particular, when being used with `use_dask=False` as required on perlmutter. Even though this is a small test case, it will take the user ~40 minutes to just create the input data. That is too slow. I may have to pick up https://github.com/CWorthy-ocean/roms-tools/issues/43 again, and refactor the regridding so that we re-use the same weights, even though this is technically a milestone for the beta release.


NoraLoose:
There is now a complete notebook draft for EXERCISE C. (I'm still waiting to hear back about the `account_key` we should use on perlmutter.)

PR #97, opened 2024-09-19
remove base_model attribute from AdditionalCode and InputDataset classes

### LINKED ISSUE #96: `AdditionalCode.base_model` and `InputDataset.base_model` attrs unused
`InputDataset`, `AdditionalCode`, and `BaseModel` are examples of lowest level C-Star classes, and it would be nice for them to not have to be built out of each other.

It is clear from the subclass of `InputDataset` which model a given dataset belongs to, we shouldn't need a specific attribute.

The existence of the `.base_model` attribute also prevents the ability of the objects in question to be initialised directly from a subset of the blueprint yaml dictionary (as it is redundant to add a `base_model` entry in the blueprint under `additional_code` and `input_dataset`, and would be impossible to make that entry point to a `BaseModel` instance anyway).

dafyddstephenson:
Closes #96

### DIFF
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -3,13 +3,12 @@ import tempfile
 from typing import Optional, List
 from pathlib import Path
 from cstar.base.datasource import DataSource
-from cstar.base.base_model import BaseModel
 from cstar.base.utils import _clone_and_checkout, _list_to_concise_str
 
 
 class AdditionalCode:
     """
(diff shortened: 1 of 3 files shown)


dafyddstephenson:
@TomNicholas this doesn't break anything blueprint related yet (that'll be the next PR), mentioned above, so if nothing alarming jumps out about this I think we can throw it straight in.

TomNicholas (approved):

Issue #99, opened 2024-09-19
Create end-to-end minimal example for testing


dafyddstephenson:
@TomNicholas and I discussed creating a new C-Star case from scratch using roms-tools as the foundation for all future tests with pytest.

The use of roms-tools means we could:
- Test remote and local datasets by uploading pre-computed datasets somewhere and reconstructing those same datasets from scratch on a user's machine
- Test roms-tools integration by also creating roms-tools yaml files of the datasets and providing these to C-Star
- Potentially parameterise things like grid size and location in future

Steps:
- [x] Create a set of input datasets in the same way that roms-tools currently does as part of its test routine, using a [minimal set of datasets hosted remotely](https://github.com/CWorthy-ocean/roms-tools-test-data)
- [ ] Create roms `.opt` and namelist files corresponding to these datasets
- [ ] Successfully run roms with these datasets
- [ ] Create a C-Star blueprint that combines all of this 
- [ ] Replace the coverage of the existing tests using pytest in conjunction with this setup

@NoraLoose and I discussed step 1, and she will be taking that on - thanks! I'll work on getting it integrated with ROMS and packaging it up as a blueprint. We'll then hand off to @TomNicholas to work it into his pytest scripts.

Related: #54

PR #101, opened 2024-09-19
Notebook that makes ROMS-Tools test data

NoraLoose:
Here is a notebook that performs step 1 of #99: it makes ROMS-Tools test data (based on raw input datasets in https://github.com/CWorthy-ocean/roms-tools-test-data), in preparation for an end-to-end testing framework.

This PR is **not meant to be merged**, I use it purely as a way to share code that later on should go into a pytest framework. We can close this PR either immediately or when further steps of #99 are completed.


NoraLoose:
I tried to make a minimal example (that still makes sense), after discussing the nature of the example with @dafyddstephenson. Despite minimality, making all datasets will take 40 seconds to 1 minute - not optimal, but the bottleneck is https://github.com/CWorthy-ocean/roms-tools/issues/134. I am working to solve the bottleneck on the ROMS-Tools side, so in a few days the data creation should hopefully be much faster.

dafyddstephenson:
Thanks Nora!

NoraLoose:
I'm closing this because this is part of #108 now.

PR #108, opened 2024-09-23
Update test case

### LINKED ISSUE #99: Create end-to-end minimal example for testing

### LINKED ISSUE #104: Testing CI not running roms-tools integration tests
Presumably this line means the tests CI job is not currently running the file `test_roms_tools_integration.py` introduced in https://github.com/CWorthy-ocean/C-Star/pull/81?

https://github.com/CWorthy-ocean/C-Star/blob/bfac303efa5ff76f98f76c61f42245b6d4e00004/.github/workflows/tests.yaml#L48

### LINKED ISSUE #106: pytest stubbornly collecting from `cstar/externals`
After C-Star has run `case.setup()` it has downloaded ucla-roms and more into the `cstar/externals` directory. Running `pytest` in the top-level directory of the repo will attempt to collect the contents of `cstar/externals`, which errors because it tries to import some kind of bad python code defined within the uclas-roms repo.

The solution to this should be to just tell pytest to ignore that directory during test collection, using e.g. `collect_ignore_glob` in `conftest.py` from this [pytest documentation](https://docs.pytest.org/en/stable/example/pythoncollection.html), i.e.
```python
# in conftest.py
collect_ignore_glob = [
    "cstar/cstar_local_config.py",
    "cstar/externals/**",
]
```

But this doesn't seem to work, `cstar/externals` is still collected, and so you get an error:

```python
ERROR cstar/externals/ucla-roms/Examples/code_check/test_roms.py - IndexError: list index out of range
```

I am completely stumped as to why `collect_ignore_glob` doesn't ignore the glob. I swear it used to work for me locally, but now it doesn't (and I have no idea what's changed). I even tried making a new package with just a `conftest.py` and a malformed python file and it worked as expected, so I am unable to reproduce the issue outside of `cstar`.

For now I think we have no choice but to just say that to run the tests you need to explicitly state the correct directory to collect the real tests from, i.e. `pytest cstar/tests/**`

dafyddstephenson:
This PR aims to update the current pytest framework with [a new test case](https://github.com/CWorthy-ocean/cstar_blueprint_test_case) that is defined on a smaller grid and integrates with `roms-tools` with minimal requirements for external data.

It was initially opened as a draft, the progression can be followed below.

Closes #106 
Closes #104 
Closes #99


cstar/tests/blueprints/fixtures.py
```diff
@@ -140,11 +140,11 @@ def modify_yaml(
 
 
 @pytest.fixture
-def test_blueprint_as_path(test_blueprint_as_dict, tmp_path) -> Callable[[str], Path]:
+def blueprint_to_path(blueprint_to_dict, tmp_path) -> Callable[[str], Path]:
     """Given the name of a pre-defined blueprint, returns it as a (temporary) path to an on-disk file."""
 
     def _blueprint_as_path(name: str, local: bool = False) -> Path:
-        blueprint_dict = test_blueprint_as_dict(name, local=local)
+        blueprint_dict = blueprint_to_dict(name, local=local)
```
TomNicholas: Renaming these to remove the `test_*` prefix is reasonable, but `blueprint_to_path` as opposed to `blueprint_as_path` I feel is a little misleading. When called in user (testing) code, this function doesn't accept a blueprint, it's used as an alternative way to way to fetch one, but `blueprint_to` to me implies that it should be passed a blueprint object to change.

dafyddstephenson: changed in [2a6ee87](https://github.com/CWorthy-ocean/C-Star/pull/108/commits/2a6ee87c427752851a19e99d94dc9551316cd34d)

dafyddstephenson:
@TomNicholas I had a crack at adding local sources back in today (as they can now be made available cheaply) but the two different test blueprints have different locations for `input_datasets`, so hardcoding the location as is currently implemented isn't ideal. 

In general though I think it'd be good to refactor this to make it a bit less complex, as right now the call sequence out of the primary test is `blueprint_as_path -> blueprint_as_dict -> set_locations -> modify_yaml -> condition_func -> modify_func` (I think) where `condition_func` and `modify_func` are set dynamically and can take multiple forms. This is difficult to follow and makes tracking changes through this sequence very complicated.

dafyddstephenson:
OK, I think this is ready to go. Sorry to basically start over @TomNicholas (what was there helped a lot with wrapping my head around what to do), I think modifying a temporary copy of the yaml before making a `Case` from it reduces the complexity a lot, so `blueprints/fixtures.py` is now down to a single short function. There are two more fixtures in `roms/fixtures.py` to fetch external data and a `tests/config.py` for defining globals, including the dict `TEST_CONFIG` over whose keys the primary test is parameterised.

I managed to sort out the pytest ignore issue, so it should work fine from anywhere now. Happy to set up a call to walk through if anything needs explaining. Looking forward to iterating and getting this in :)

NoraLoose:
As `ROMS-Tools` develops the NetCDF data (maybe even the NetCDF file names) produced by YAML files like https://github.com/CWorthy-ocean/cstar_blueprint_test_case/blob/main/roms_tools_yaml_files/roms_bry_bgc.yaml 
will change as well.

Should we read the header
https://github.com/CWorthy-ocean/cstar_blueprint_test_case/blob/d7996aea8d4fd4ca4148b34d2d898f019c90a8ff/roms_tools_yaml_files/roms_bry_bgc.yaml#L1-L3
and then check out the appropriate version of ROMS-Tools to get the expected behavior?

dafyddstephenson:
As in, make the roms-tools version associated with the test case the version that we fix as the dependency?

NoraLoose:
> As in, make the roms-tools version associated with the test case the version that we fix as the dependency?

This could be an option. Or we just fix the roms-tools version when running the tests? I'm adding new features to ROMS-Tools continuously, and wouldn't want the C-Star user to miss out on them. But on the other hand, we probably want to test the ROMS-Tools integration with the same version that the user will be using. Not sure...

dafyddstephenson:
hadn't occurred to me that we could have a separate version for the tests, but yes, that makes a lot more sense.

NoraLoose:
Maybe we could have a GitHub workflow that updates the data in https://github.com/CWorthy-ocean/cstar_blueprint_test_case? Too complicated?

dafyddstephenson:
I think it makes more sense to keep it (relatively) static given that it's supposedly another "blueprint". We can issue updates but probably not have it continuously change

NoraLoose:
Or should it be a feature of C-Star (**beta version**) to switch to the ROMS-Tools version given in the header? To ensure reproducibility of the ROMS input data for a given blueprint with ROMS-Tools YAML files?

dafyddstephenson:
Is there a way to hot-switch the version of a dependency without reinstalling/rebuilding your environment...?

NoraLoose:
I think we need to overwrite via
```
pip install roms-tools==x.x.x
```

dafyddstephenson:
Not sure I understand how that would be implemented, perhaps we can discuss on a dedicated issue?

TomNicholas:
> Or should it be a feature of C-Star (beta version) to switch to the ROMS-Tools version given in the header?

I'm not sure I fully follow this conversation, but one library changing the version of another library sounds not good. Shouldn't we just be aiming to get to a point where changes in roms-tools API don't affect the reproducibility of the result?

pyproject.toml
```diff
@@ -58,4 +58,17 @@ include = ["cstar"]
...
+norecursedirs = [".tox", ".git", "__pycache__", ".#*", "*~"]
+
+# Use addopts to ignore specific files and directories
+addopts = [
+    "--ignore-glob=cstar/cstar_local_config.py",
+    "--ignore-glob=cstar/externals/**",
+    "--ignore-glob=examples/**.py",
+    "--ignore-glob=*~",
+    "--ignore-glob=.#*"
+]
```
TomNicholas, line 81: So these get used every time automatically?

cstar/tests/blueprints/fixtures.py
```diff
@@ -1,156 +1,80 @@
...
+            modified_template_content = modified_template_content.replace(
+                oldstr, newstr
+            )
+        temp_path = tmpdir.join(template_blueprint_path.name)
+        with open(temp_path, "w") as temp_file:
+            temp_file.write(modified_template_content)
+
+        return Case.from_blueprint(temp_path, **kwargs)
+
+    return _template_blueprint_to_case
```
TomNicholas: The thing about this fixture is you are mixing some business logic (`Case.__init__`) in with testing setup logic. If instead this fixture returned a `tmp_path` to the modified blueprint file, and your tests took that as an input, then your unit tests would explicitly include all required public imports and business logic.

cstar/tests/blueprints/fixtures.py
```diff
@@ -1,156 +1,80 @@
...
+            template_content = template_file.read()
+
+        modified_template_content = template_content
+        for oldstr, newstr in strs_to_replace.items():
+            modified_template_content = modified_template_content.replace(
+                oldstr, newstr
+            )
+        temp_path = tmpdir.join(template_blueprint_path.name)
+        with open(temp_path, "w") as temp_file:
+            temp_file.write(modified_template_content)
```
TomNicholas, line 62: Happy with this rewrite to modify the file on disk if you find that easier (the call chain is definitely simpler), the important thing is that the blueprint used is a temporary object in a temporary directory that gets cleaned up, so that each test has no side-effects on any global state. 👍

cstar/tests/config.py
```diff
@@ -0,0 +1,70 @@
+"""
+Configuration module for testing C-Star with pytest.
+
+The first section sets or gets any globally accessible paths (e.g. that of the test directory itself, TEST_DIRECTORY).
```
TomNicholas, line 4: The test directory just means the directory from which you call `pytest`?

NoraLoose:
> Shouldn't we just be aiming to get to a point where changes in roms-tools API don't affect the reproducibility of the result?

ROMS-Tools is not at that point yet. For example, we just exchanged the filling method (https://github.com/CWorthy-ocean/roms-tools/pull/140), which altered the results. Next on the list are:
* https://github.com/CWorthy-ocean/roms-tools/issues/133
* https://github.com/CWorthy-ocean/roms-tools/issues/153
* https://github.com/CWorthy-ocean/roms-tools/issues/154
* https://github.com/CWorthy-ocean/roms-tools/issues/157

all of which will again alter the results.

Btw, the continuous alteration of results also causes a problem within ROMS-Tools itself: https://github.com/CWorthy-ocean/roms-tools/issues/152

cstar/tests/roms/test_cstar_test_blueprints.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TestCStar:
+    @pytest.mark.parametrize(
+        "test_config_key",
+        [
+            "test_case_remote_with_netcdf_datasets",
+            "test_case_remote_with_yaml_datasets",
+            "test_case_local_with_netcdf_datasets",
+            "test_case_local_with_yaml_datasets",
+        ],
```
TomNicholas, line 14: This seems relatively neat.

cstar/tests/config.py
```diff
@@ -0,0 +1,70 @@
...
+    spec = find_spec("cstar")
+    if spec is not None:
+        return Path(spec.origin).parent / "tests"
+    else:
+        raise RuntimeError("Cannot determine the package location of cstar")
+
+
+TEST_DIRECTORY = _get_test_directory()
+ROMS_TOOLS_DATA_DIRECTORY = Path(os_cache("roms_tools_datasets_for_cstar_test_case"))
+CSTAR_TEST_DATA_DIRECTORY = Path(os_cache("cstar_test_case_data"))
```
TomNicholas, line 32: Why can't these also be temporary directories, provided via fixtures to the other fixtures (e.g. `fetch_roms_tools_source_data`)?

cstar/tests/roms/fixtures.py
```diff
@@ -0,0 +1,155 @@
...
+            )
+
+        symlink_path.symlink_to(ROMS_TOOLS_DATA_DIRECTORY)
+
+        def cleanup():
+            if symlink_path.is_symlink():
+                print(f"removing {symlink_path}")
+                symlink_path.unlink()
+
+        request.addfinalizer(cleanup)
```
TomNicholas, line 93: This seems like a complicated way to just end up replicating what `pytest`'s built-in `tmpdir` fixture does?

cstar/tests/roms/fixtures.py
```diff
@@ -0,0 +1,155 @@
...
+            "https://github.com/CWorthy-ocean/cstar_blueprint_test_case/"
+        )
+        checkout_target = "d7996aea8d4fd4ca4148b34d2d898f019c90a8ff"
+
+        # Construct the URL of this commit as a zip archive:
+        archive_url = f"{test_case_repo_url.rstrip('/')}/archive/{checkout_target}.zip"
+
+        # Create a cache dir with pooch
+        # cache_dir = Path(pooch.os_cache("cstar_test_case_data"))
+
```
TomNicholas: delete

cstar/tests/roms/fixtures.py
```diff
@@ -0,0 +1,155 @@
...
+        # Unzip the files into a subdirectory `extract_dir` of `cache_dir`
+        with zipfile.ZipFile(zip_path, "r") as zip_ref:
+            extract_list = zip_ref.namelist()
+            extract_dir = CSTAR_TEST_DATA_DIRECTORY / extract_list[0]
+            zip_ref.extractall(CSTAR_TEST_DATA_DIRECTORY)
+
+        # Copy the contents of the subdirectory up one and remove it and the zip file
+        shutil.copytree(extract_dir, CSTAR_TEST_DATA_DIRECTORY, dirs_exist_ok=True)
+        shutil.rmtree(extract_dir)
+        Path(zip_path).unlink()
```
TomNicholas, line 150: Why is this moving around of files on-disk necessary?

cstar/tests/test_fixtures.py
```diff
@@ -7,38 +18,139 @@ def test_mock_input_fixture(mock_user_input):
...
 
 
-class TestGetBluePrints:
-    def test_get_blueprint(self, test_blueprint_as_dict):
-        blueprint_dict = test_blueprint_as_dict("ROMS_MARBL", local=False)
+def test_template_blueprint_to_case(template_blueprint_to_case, tmpdir):
+    """
+    This test verifies that the template_blueprint_to_case fixture correctly reads a
+    specified blueprint, performs string replacements, and returns a Case instance with
+    the correct parameters corresponding to the string replacements.
```
TomNicholas: See this test is testing your private test fixture, but also the public `Case.__init__` method. Ideally each of these would be tested independently.

TomNicholas:
Great progress @dafyddstephenson ! I think the fixture structure could be re-jigged a little still.

pyproject.toml
dafyddstephenson: yes

cstar/tests/blueprints/fixtures.py
dafyddstephenson: sounds good, should be easy to do

cstar/tests/config.py
dafyddstephenson: It refers to `<top_level>/cstar/tests` , so perhaps should be "tests" directory. I am trying to use absolute paths throughout rather than defining things relative to the directory in which pytest is run.

cstar/tests/config.py
dafyddstephenson: I've borrowed the roms-tools implementation where external data that are _needed by but not created by_ the test suite is fetched to a cache dir - a little softer than a temporary directory but allows re-use locally between tests without affecting the test outcomes themselves.

dafyddstephenson: to the second part, having a global name in the config module allows us to define where they go in a central location that is used consistently elsewhere. The directories aren't actually created here (I don't think), the path is just defined. They're then created in fixtures which use these vars.

cstar/tests/roms/fixtures.py
dafyddstephenson: I'm not sure if there's a way to create a temporary symlink with tmpdir but I'll happily switch to that if so. The idea here is that the directory isn't temporary (it's cached) but the local manifestation of it (the symbolic link to that cache dir seen by C-Star) is.

cstar/tests/roms/fixtures.py
dafyddstephenson: All the test data is in a git repo. This routine downloads the repo as a zip archive, rather than cloning it with all its history and baggage. Unfortunately it can only be unzipped to a subdirectory of where we want it to go, so this moves the contents of the subdirectory up one level.

cstar/tests/config.py
TomNicholas: > a cache dir - a little softer than a temporary directory but allows re-use locally between tests without affecting the test outcomes themselves.

But isn't it possible to achieve the same thing using a pytest `tmpdir` fixture with `scope='session'` or similar?

cstar/tests/roms/fixtures.py
TomNicholas: That's certainly a clever idea, but maybe if you did what I'm suggesting in https://github.com/CWorthy-ocean/C-Star/pull/108#discussion_r1779093213 (and don't modify what's in that `tmpdir`) you won't need this symlink any more?

cstar/tests/roms/fixtures.py
TomNicholas: Huh. You can't just unzip into the `tmpdir`?

cstar/tests/roms/fixtures.py
dafyddstephenson: I believe however we handle the fetching of roms-tools data, a symlink will be needed here. The roms-tools yaml files contain a "path" entry that here is just a relative path to `roms_tools_datasets/FILE.nc` in each case. So either we have to do (more) messy yaml editing with marked template files and placeholders, or we can just make sure something is at the location expected by roms-tools (as is done with this symlink implementation). I think in this case the former is avoidable and (personal opinion) the latter is neater.

cstar/tests/roms/fixtures.py
dafyddstephenson: not directly, it would go into a named subdirectory of wherever you ask it to be unzipped. The name isn't specifiable either, it's written into the zip file metadata, and is somewhat unpleasant, containing the full commit hash. Kind of annoying but livable imho.

cstar/tests/roms/fixtures.py
TomNicholas: Yeah editing the yamls further is messy...

> a relative path to roms_tools_datasets/FILE.nc in each case

This is relative to the `CASEROOT`? (which is a `tmpdir`?) 

We should really write out a description of what the eventual decision is here as part of some developer docs because this is confusing 😅

cstar/tests/roms/fixtures.py
TomNicholas: That is a bit annoying, but yeah not the end of the world.

cstar/tests/roms/fixtures.py
dafyddstephenson: I could just beef up the docstring for this fixture a bit if it helps?

dafyddstephenson:
@NoraLoose Have added a `tests` optional dependency list to the pyproject.toml and fixed `roms_tools` at 1.4.0 there.
@TomNicholas added a docstring to the main test and restricted the responsibility of the fixture to just modifying the blueprint template (the case is now created within the test itself)

cstar/tests/roms/fixtures.py
TomNicholas: Still needs to be deleted!

dafyddstephenson: done in 7ed3a60c751d0c36614db2ff2658918f2d682cb0

TomNicholas (approved):

PR #119, opened 2024-10-01
Refactor namelist modification

### LINKED ISSUE #118: Namelists are only modifiable once
From #114 :

- To do a restart, we need to update the namelist with the new IC file and number of timesteps
- However, `AdditionalCode.get()` clones the template namelist `roms.in_TEMPLATE` to a local working copy once and for all - all future modifications happen to the working copy.
- If we want to change the IC file entry and then change it again, we have lost the placeholder string to replace in the modified version. 
- A “simple” solution would be to have an `update_namelist` method that always goes back to the template version and replaces its placeholder strings with new values
- However, we would need to track all the other changes that have been happening too, not just the value that’s being replaced in the moment, or we would have an “updated” namelist that had N-1 placeholders and one correct value.
- In order to track namelist modifications we need a `namelist_modifications` property, that exists parallel to `namelists` (a list with one entry per namelist file). Each entry is a dictionary of values to replace (initially `{}` ).

Related: #113  , #114

dafyddstephenson:
Closes #118 . Branched from #117 , merge #117 and update diffs before review (should be +209, -123)

This PR:
- adds a `ROMSComponent._namelist_modifications` property (`list[dict]`) "parallel" to `ROMSComponent.namelists` that tracks needed modifications to the template file based on the state of `ROMSComponent`
- adds an `update_namelists` method to `ROMSComponent` that makes a new set of modified namelist files based on the current values in the `_namelist_modifications` property.
- Moves all namelist modification to immediately before runtime, in `ROMSComponent.run()`


dafyddstephenson:
Tests are failing as `pytest` code has diverged. Will be corrected after merging #108 .

PR #150, opened 2024-10-24
Restructured tests

dafyddstephenson:
@TomNicholas can we prioritise this PR, which simply reshuffles the test directory and makes separate workflows for unit and integration tests? All the others to come will branch from it so it'd be good to get this in first. 

It adds an empty unit test module as a placeholder, but this will be removed when we start adding others.


TomNicholas:
> separate workflows for unit and integration tests?

I agree that splitting up the code for unit tests and integration tests is a good idea, but I'm not sure why they need to be run in separate GH workflows? Especially as they seem to use the same `environment.yml`. Do they have different dependencies?

cstar/tests/unit_tests/base/test_nothing.py
```diff
@@ -0,0 +1,2 @@
+def test_nothing():
+    pass
```
TomNicholas, line 2: What's the purpose of this?

dafyddstephenson: > It adds an empty unit test module as a placeholder, but this will be removed when we start adding others.

Github workflow fails on test collection without something to collect

TomNicholas: Riight sorry.

dafyddstephenson:
> I agree that splitting up the code for unit tests and integration tests is a good idea, but I'm not sure why they need to be run in separate GH workflows? Especially as they seem to use the same environment.yml. Do they have different dependencies?

No it's the same environment which is a little annoying as I spent quite a bit of time trying to cache and share it with no success. Any ideas? I tried adding an independent workflow that set up and cached the env then calling it from both test sequences.

I think it's just nicer to have "unit tests failed"/"integration tests passed" in the CI summary without having to dig. They also run simultaneously, and unit tests also run _much_ faster (fractions of a second once the env is set up), so it'll fail earlier if something is fundamentally wrong

TomNicholas:
> No it's the same environment which is a little annoying as I spent quite a bit of time trying to cache and share it with no success. Any ideas? I tried adding an independent workflow that set up and cached the env then calling it from both test sequences.

This is frustrating but doesn't seem quite like what I was asking about - I just meant if you're running different sets of tests do those tests need all of the same depdendencies? Maybe they do.

> I think it's just nicer to have "unit tests failed"/"integration tests passed" in the CI summary without having to dig. They also run simultaneously, and unit tests also run much faster (fractions of a second once the env is set up), so it'll fail earlier if something is fundamentally wrong

These are good reasons, you've changed my mind

TomNicholas (approved):

dafyddstephenson:
One thing I have considered is that perhaps we should be calling our "integration tests" end to end or system tests as we don't have anything more granular than a complete workflow sequence in the "integration tests" currently and almost certainly want to change that, probably without mixing up testing minimally mocked internal sequences with full user-like tests like we currently have.

TomNicholas:
> end to end or system tests

That is already what I understand the term "integration tests" to mean. I would say if it tests more multiple methods / features at once it's arguably an integration test as opposed to a unit test.

dafyddstephenson:
Yep, I'm more referring to the difference between integration testing and system testing than between integration testing and unit testing, but we can revisit this after.

PR #163, opened 2024-11-05
Basic jira integration

dafyddstephenson:
This PR makes a first attempt at a Jira-Github integration.

If the user adds the `jira-story` label to a new issue, a GitHub actions workflow is triggered that:
- creates a story on Jira whose summary is the issue title and whose description is the issue body text
- Assigns the person assigned to the issue in Jira (currently only either Nora, Tom, or I)
- Creates subtasks of that story for every task marked in the issue and assigns the same person as before
- Adds a link to the Github issue in the Jira story description
- Comments on the original issue using a bot that links to the Jira story

If the user then closes the issue, a second workflow is triggered to mark the issue as "Done" in Jira

Still to address:
- Editing the issue does not update the issue in Jira
- There is no mechanism for creating epics or attaching the story to an epic [should be done manually in Jira for now]
- Everything is currently done under my email and API token - I suggest we create a general Jira account for the org or use Joe's or something to avoid confusion


dafyddstephenson:
@TomNicholas can we prioritise merging this one in particular as my main being diverged (necessary for this to work) is leading to some complications?

TomNicholas:
I don't see the `jira-story` label in the list of labels. Was it deliberately removed?

PR #167, opened 2024-11-09
Internal testing documentation updates

dafyddstephenson:
second round of doc updates following internal testing


NoraLoose:
Looks good!

Do you know why the figure in notebook 3 changed?

dafyddstephenson:
I didn't figure it out - any small change in the input files could have triggered it though. Will revert to 4 boundaries before merge, watch this space

NoraLoose:
> I didn't figure it out - any small change in the input files could have triggered it though.

Thinking about it, this is most likely triggered by a change in `roms-tools`: I have changed the extent of the raw before applying the fill method between versions, see https://github.com/CWorthy-ocean/roms-tools/issues/153. This might have mattered a lot for the coarse CESM BGC data. The new behaviour is more accurate, though.

dafyddstephenson:
Right, as we're in a fluid state of development and it's a test domain I'm not losing sleep unless the changes are drastic (the broad structures are largely the same but with _slightly_ different magnitudes, if you compare) - the figures show surface fields which are basically a fingerprint of the surface forcing anyway, so any change there makes sense.

PR #168, opened 2024-11-14
Refactor environment handling

dafyddstephenson:
This PR:
-  moves away from passing around ever-increasing numbers of non-public globals such as `_CSTAR_ROOT` to manage the environment, and instead introduces a `CStarEnvironment` ABC, with subclasses for each supported system. The class has only properties, most of which are determined dynamically (e.g. `environment.uses_lmod` ). A call to a new factory method, `set_environment()`, is made when `cstar.base.environment` is imported, returning the correct subclass instance.
- Ends the use of a dynamically created `cstar_local_config.py` file and handles environment variables using the [`dotenv` package](https://github.com/theskumar/python-dotenv), which reads them from a shell-like `.env` file, supporting variable expansion and appending. The user's environment is saved to `~/.cstar.env`
- Adds full unit test coverage of the `base/environment.py` module
- Stops using `env_modules_python.py`, the python module present on most HPC systems that use Linux Environment Modules. Importing it was inflexible and complicated, and the module is effectively a very simple wrapper for `subprocess`.

It will be followed in the next PR by a similar refactor for handling jobs on HPC schedulers, which will migrate many of the scheduler-related properties out of the `CStarEnvironment` class into a dedicated `Scheduler` class.

UPDATE (2024-11-22):
Have refactored this again - it still defines a `CStarEnvironment` class but it is not an ABC and there is no subclassing for different systems.

Instead, a `CStarSystem` class is defined in the module `system.py`, with a property `environment` that returns the correct `CStarEnvironment` instance based on your system configuration. 

`CStarSystem` will similarly have a `scheduler` property in the next PR. This is to avoid having two separate modules, `scheduler` and `environment` with a circular dependency, while maintaining a separation of concerns between scheduler management and environment management.


dafyddstephenson:
moving to WIP as this no longer seems to be working on Perlmutter after the downtime

dafyddstephenson:
> moving to WIP as this no longer seems to be working on Perlmutter after the downtime

fixed

dafyddstephenson:
Going to put back as WIP for some minor restructuring in order to accommodate future scheduler changes

dafyddstephenson:
Have refactored again (see "UPDATE (2024-11-22)" in the original PR decription) and fully test-driven. Ready to go.

PR #171, opened 2024-11-18
[WIP] Case unit tests

dafyddstephenson:
This draft PR picks up @NoraLoose 's earlier work on unit tests for `Case`. This branch also previously held unit tests for `BaseModel`, but these commits have been separately incorporated into PR #155 , so have now been removed here.


dafyddstephenson:
This will be addressed by #229

Issue #88, opened 2024-09-17
Use python's logging module


dafyddstephenson:
This issue is to track the implementation of python's logging. I was hoping to do this with PR #89  but the more I read about logging's capabilities, the more I'd rather implement it carefully.

Key things we'd like to achieve: 
- write subprocess output to log files and then hint to the user where they can find these files
- Convert current print statements to logs at the `info` level
- add debug information throughout the code using logging's `debug` level

Things to consider:
- Logging has a global configuration that should be set once at the entry point (the best candidate currently would be `environment.py`)
- "Loggers" can then be set up and re-used to track different things (we could imagine for example a `roms_compile_logger`
- "Handlers" decide what to do with the logs (to file, to screen, etc.)

So where we set the global configuration, how many loggers we have and what their responsibilities are, and how to handle their output (new file each time with a date string in the filename is what I imagined) is to be decided before implementation.

PR #175, opened 2024-11-30
Scheduler class

dafyddstephenson:
## Summary
This PR:
- expands CStarSystem to add a `scheduler` property that determines the correct scheduler for the system and returns an appropriate instance of the correct `Scheduler` subclass (e.g. `SlurmScheduler`, `PBSScheduler`)
- Adds a `Queue` ABC (related to `Scheduler`)
- Adds a `scheduler_job` module with a `SchedulerJob` ABC and `SlurmJob` and ` PBSJob` subclasses. 

## More on `Scheduler`:
The `Scheduler` instance is instantiated with, e.g.:
```
SlurmScheduler(
                    queues=[SlurmQueue(name="regular", query_name="regular_1"),
                            SlurmQueue(name="shared"),
                            SlurmQueue(name="debug")],
                    primary_queue_name="regular",
                    queue_flag="qos",
                    other_scheduler_directives={"-C": "cpu"},
                )

```
Here:
- `queue_flag` refers to the name used to choose the queue in scheduler submissions, typically `partition` or `qos`
- `other_scheduler_directives` refers to system-specific choices that are always made by C-Star.

Properties and methods of note:
- `global_max_cpus_per_node`
- `global_max_mem_per_node_gb` 
both query the system to determine the absolute limits on cores per node and memory per node.

## More on `Queue`:
- `name` in `__init__` refers to the name used to submit a job to the queue, while `query_name` refers to the name used by the system when querying with the `sacctmgr` command. These default to being the same, but sometimes differ, as in the above example.

Properties of note (obtained by querying using `sacctmgr show qos` on slurm):
- `max_walltime`
- `max_nodes`
- `priority`

## More on `SchedulerJob`
A SchedulerJob instance is instantiated with, e.g.:
```
sj = SlurmJob(scheduler = cs.scheduler,
              commands = "echo 'hello world'",
              cpus = 1,
              account_key = os.environ.get("ACCOUNT_KEY"),
              script_path = Path.cwd()/"my_script.sh",
              run_path = Path.cwd(),
              job_name = "test_job",
              output_file = "test_file.out",
              queue_name = "shared",
              send_email = True,
              walltime = "00:10:00")
```
Properties and methods of note:
- `script` : dynamically creates a submission script from these parameters
- `submit` : submits the job to the scheduler and captures/returns the job id
- `id` : the job id mentioned above
- `status` : submitted, cancelled, running, complete
- `cancel` cancels the job after submitting to the scheduler


cstar/base/scheduler.py
```diff
@@ -0,0 +1,103 @@
+from abc import ABC
+from typing import List, Dict, Optional, TYPE_CHECKING
+
+import subprocess
+
+if TYPE_CHECKING:
+    pass
```
TomNicholas:
```suggestion
```

cstar/base/scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+        queues: List["Queue"],
+        primary_queue_name: str,
+        other_scheduler_directives: Optional[Dict[str, str]],
+    ):
+        self.queue_flag = queue_flag
+        self.queues = queues
+        self.queue_names = [q.name for q in queues]
+        self.primary_queue_name = primary_queue_name
+        self.other_scheduler_directives = (
+            other_scheduler_directives if other_scheduler_directives is not None else {}
```
TomNicholas: A docstring containing a complete example command would make the role of each of these arguments a lot clearer

cstar/base/scheduler_job.py
```diff
@@ -0,0 +1,287 @@
+## STIL TODO:
```
TomNicholas: The prefix `scheduler_` in `scheduler_job.py` seems unnecessary? Or if this will only get used in conjunction with `scheduler.py`, perhaps they should both go in their own dedicated directory/module?

cstar/base/scheduler_job.py
```diff
@@ -0,0 +1,287 @@
...
+        queue_name=queue_name,
+        send_email=send_email,
+        walltime=walltime,
+    )
+
+
+class SchedulerJob(ABC):
+    @abstractmethod
+    def submit(self):
+        pass
```
TomNicholas: Would an enumeration of the possible statuses of a job be useful here?

```suggestion
class JobStatus(Enum):
    PENDING = auto()
    RUNNING = auto()
    COMPLETED = auto()
    FAILED = auto()


class SchedulerJob(ABC):
    @abstractmethod
    def submit(self):
        pass
    
    @property
    def status(self) -> JobStatus:
        return self._status
```

cstar/base/scheduler_job.py
```diff
@@ -0,0 +1,287 @@
...
+    @property
+    def id(self):
+        if self._id is None:
+            print("No Job ID found. Submit this job with SlurmJob.submit()")
+        return self._id
+
+    @property
+    def status(self):
+        if self.id is None:
+            return "unsubmitted"
```
TomNicholas: This really should be replaced with an `JobStatus` Enum to avoid "magic strings" (i.e. the string `"unsubmitted"` has a lot more meaning than the literal characters contained in it).

cstar/base/system.py
```diff
@@ -2,11 +2,36 @@
...
+    _scheduler: Optional[Scheduler] = None
+
+    @property
+    def scheduler(self) -> Scheduler:
+        """todo."""
+        if self._scheduler is not None:
+            return self._scheduler
+
+        match self.name:
+            case "perlmutter":
```
TomNicholas: Yeah we could be matching enums here.

cstar/base/system.py
```diff
@@ -2,11 +2,36 @@
...
 
 class CStarSystem:
     _environment: Optional[CStarEnvironment] = None
+    _scheduler: Optional[Scheduler] = None
+
+    @property
+    def scheduler(self) -> Scheduler:
+        """todo."""
+        if self._scheduler is not None:
+            return self._scheduler
```
TomNicholas: I think this is a bit of an anti-pattern - shouldn't we always have a `Scheduler`, so the type of `_scheduler` shouldn't be `Optional`?

Could you instead have the private attribute always be set when the class is instantiated?

```python
class CStarSystem:
    _scheduler: Scheduler

    def __init__(self) -> None:
        self._scheduler = determine_scheduler_from_env()
```

cstar/case.py
```diff
@@ -535,8 +535,8 @@ def run(
         account_key=None,
         walltime=cstar_system.environment.max_walltime,
         queue=cstar_system.environment.primary_queue,
-        job_name="my_case_run",
-    ) -> None:
+        job_name=None,
+    ) -> Optional["SchedulerJob"]:
```
TomNicholas, line 538: This type is confusing to me - does it return `None` if the job failed to submit? Or if it did submit but on a system where the job runs in the same process? This is kind of like returning a `dask.Delayed` object...

TomNicholas (commented):
This is cool. 

We should be trying really hard to keep the job scheduling and submission separate from the CStar-specific logic, because we maybe should be using one of the existing frameworks for job submission, e.g. https://parsl.readthedocs.io/en/stable/stubs/parsl.providers.SlurmProvider.html

---

I mentioned Enums a few times - I think now would be a good time to start using them. The `CStarEnvironment` class added in a previous PR could also potentially be an Enum.

dafyddstephenson:
Hi Tom, thanks for the feedback - will take a look tomorrow. Just a heads up that this is still WIP and there is significant work still to do - docstrings and such will come later (though I can provide ones for the things that are locked in place now if it will help read). Cheers!

cstar/case.py
dafyddstephenson: re. this and the above, comment, we do not always have a scheduler. Our integration tests on Github, for example, do not use a scheduler.

cstar/base/system.py
dafyddstephenson: see below

cstar/base/scheduler_job.py
dafyddstephenson: Yes, much better idea. This felt a bit wonky and I couldn't put my finger on why

cstar/case.py
TomNicholas: Okay. Just checking - does this mean that the syntax for running a case becomes different on different systems?

cstar/base/scheduler_job.py
dafyddstephenson: I have now restructured things so there's a dedicated `system` directory.
However `scheduler.py` and `environment.py` are sort of special cases in that they _define_ universal aspects of the system that should be available everywhere else in the package (via `CStarSystemManager`). As such, these are the only two modules that do not import from anywhere else in C-Star (to avoid circular importing). 

While seemingly closely related to `scheduler.py`, `scheduler_job.py` defines HPC job management-related methods, and should be able to import from elsewhere in C-Star (e.g. it requires access to system-level environment information, rather than providing this information). While on the surface `scheduler.py` and `scheduler_job.py` may seem related, this distinction means they should be kept separate.

The prefix is necessary to avoid ambiguity, as C-Star can run jobs without a scheduler.

cstar/base/scheduler_job.py
dafyddstephenson: Done in 907b767c24e687d7bf26448d6a668ffd674be5d4 and 23ab4d192f6299904fc71865b909f3311429ddd6

cstar/base/scheduler_job.py
dafyddstephenson: Done in https://github.com/CWorthy-ocean/C-Star/commit/907b767c24e687d7bf26448d6a668ffd674be5d4 and https://github.com/CWorthy-ocean/C-Star/commit/23ab4d192f6299904fc71865b909f3311429ddd6

cstar/base/system.py
dafyddstephenson: Done in https://github.com/CWorthy-ocean/C-Star/commit/907b767c24e687d7bf26448d6a668ffd674be5d4 and https://github.com/CWorthy-ocean/C-Star/commit/23ab4d192f6299904fc71865b909f3311429ddd6

cstar/case.py
dafyddstephenson: No, if `scheduler` is `None` then `Case.run()` will return `None` (and the execution will happen in the user's current python session). If there's a scheduler, it will return an appropriate `SchedulerJob` instance and the execution will happen somewhere else on the system.

TomNicholas: Okay that's great, nice.

cstar/base/scheduler_job.py
TomNicholas: That all makes sense, but then wouldn't a clearer name for `scheduler_job.py` be `scheduler/hpc.py`, and your standalone scheduler code that needs to be importable from anywhere can go in `scheduler/scheduler.py`?

> as C-Star can run jobs without a scheduler

Is it really a "job" if it's not using a scheduler? It's not a job in the normal sense of "HPC job submission queue". I would say C-Star can run a case without a scheduler, but not run the case as a job without a scheduler.

I mean this is not particularly important ofc - I don't want to waste your time.

dafyddstephenson: I could do `cstar/system/scheduler.py` and `cstar/scheduler/job.py`? I think that would perhaps even make more sense.  May just seem a bit more fragmented, e.g. `job.py` would be the only module in `scheduler/`.

TomNicholas: Happy with that also. We can also easily change this later as long as it's not too tangled.

dafyddstephenson: I can do it now, no sweat

dafyddstephenson:
> maybe should be using one of the existing frameworks for job submission, e.g. https://parsl.readthedocs.io/en/stable/stubs/parsl.providers.SlurmProvider.html

I cannot tell you how hard I looked for a python package to take care of all this nonsense, how did you find this?! 
Right now I'm going to stick to what I've written here (it mostly just significantly improves what was already in), but we should definitely be looking into moving to Parsl later if it seems like it can do what we're looking for.

dafyddstephenson:
Took some extra time to sharpen this up more now it wasn't deadline-constrained, but got delayed as always by having to update the notebooks on Perlmutter. 

Ready to go now.

cstar/case.py
```diff
@@ -8,14 +8,14 @@
 
 from cstar.base.component import Component
 
-# from cstar.base.environment import environment
-from cstar.base.system import cstar_system
+# from cstar.system.environment import environment
```
TomNicholas:
```suggestion
```

cstar/case.py
```diff
@@ -533,16 +533,22 @@ def pre_run(self) -> None:
     def run(
         self,
         account_key=None,
-        walltime=cstar_system.environment.max_walltime,
-        queue=cstar_system.environment.primary_queue,
-        job_name="my_case_run",
-    ) -> None:
+        walltime=None,
+        queue=None,
+        job_name=None,
```
TomNicholas: Type hints here would be nice - especially now as we are using some enums, it would be good to distinguish between types that are arbitrary strings and types that are one of a specific set of enum values.

cstar/case.py
```diff
@@ -533,16 +533,22 @@ def pre_run(self) -> None:
...
+    ) -> Optional["SchedulerJob"]:
         """Run the case by calling `component.run(caseroot)` on the primary component
         (to which others are coupled)."""
 
         # Assuming for now that ROMS presence implies it is the master program
         # TODO add more advanced logic for this
         # 20240807 - TN - set first component as main?
+
+        if (queue is None) and (cstar_sysmgr.scheduler is not None):
```
TomNicholas: These brackets are redundant in python. I'm suprised the pre-commit linter hasn't removed them automatically. Another example of https://github.com/CWorthy-ocean/C-Star/issues/183

```suggestion
        if queue is None and cstar_sysmgr.scheduler is not None:
```

cstar/case.py
```diff
@@ -533,16 +533,22 @@ def pre_run(self) -> None:
...
+            queue = cstar_sysmgr.scheduler.primary_queue_name
+        if (walltime is None) and (cstar_sysmgr.scheduler is not None):
+            walltime = cstar_sysmgr.scheduler.get_queue(queue).max_walltime
```
TomNicholas: So if `walltime` is `None` but scheduler is `None`, does `walltime` just stay as `None`? Is that okay?

This would be a case where having a public `walltime` variable and a private `_walltime` variable would allow mypy to catch possibly unwanted types sneaking through.

cstar/case.py
```diff
@@ -557,15 +563,17 @@ def run(
...
+                job_instance = component.run(
                     output_dir=self.caseroot / "output",
                     n_time_steps=ntimesteps,
                     account_key=account_key,
                     walltime=walltime,
                     queue=queue,
                     job_name=job_name,
                 )
+                return job_instance
+        return None
```
TomNicholas: So if none of the components are ROMS this just returns `None`? Is that ever correct? Perhaps it should just error if it doesn't return before it gets to this line?

cstar/roms/component.py
```diff
@@ -880,6 +881,11 @@ def run(
                 + "\n If you have already run Component.build(), either run it again or "
                 + " add the executable path manually using Component.exe_path='YOUR/PATH'."
             )
+        if (queue is None) and (cstar_sysmgr.scheduler is not None):
+            queue = cstar_sysmgr.scheduler.primary_queue_name
+        if (walltime is None) and (cstar_sysmgr.scheduler is not None):
+            walltime = cstar_sysmgr.scheduler.get_queue(queue).max_walltime
+
```
TomNicholas: Is this duplicated logic from the parent class?

cstar/roms/component.py
```diff
@@ -920,202 +926,111 @@ def run(
...
+            romsprocess = subprocess.Popen(
+                roms_exec_cmd,
+                shell=True,
+                cwd=run_path,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+            )
+
+            # Process stdout line-by-line
```
TomNicholas, line 969: All of this stdout processing should be abstracted into a function

cstar/roms/component.py
```diff
@@ -920,202 +926,111 @@ def run(
...
+            tstep0 = 0
+            roms_init_string = ""
+            if romsprocess.stdout is None:
+                raise RuntimeError("ROMS is not producing stdout")
+
+            # 2024-09-21 : the following is included as in some instances ROMS
+            # will exit with code 0 even if a fatal error occurs, see:
+            # https://github.com/CESR-lab/ucla-roms/issues/42
+
+            debugging = False  # Print raw output if true
```
TomNicholas, line 979: This should either be removed or made into proper logging, used only when a global logging flag is set to debug level (xref https://github.com/CWorthy-ocean/C-Star/issues/88)

cstar/roms/component.py
```diff
@@ -920,202 +926,111 @@ def run(
...
+                        break
+                    if output:
+                        print(output.strip())
+            else:
+                for line in romsprocess.stdout:
+                    # Split the line by whitespace
+                    parts = line.split()
+
+                    # Check if there are exactly 9 parts and the first one is an integer
+                    if len(parts) == 9:
```
TomNicholas, line 993: This is completely opaque without further comments giving context

TomNicholas: It's also possible that whatever is being done here could be replaced with a regex match

cstar/scheduler/job.py
```diff
@@ -0,0 +1,1001 @@
...
+
+        Returns
+        -------
+        id: int or None
+            The unique job ID assigned by the scheduler, or `None` if the job has not
+            been submitted.
+        """
+
+        if self._id is None:
+            print("No Job ID found. Submit this job with SchedulerJob.submit()")
```
TomNicholas, line 458: This property is nice, but this should probably raise (maybe a `RuntimeError`), because then the return type can always be `int`.

cstar/scheduler/job.py
```diff
@@ -0,0 +1,1001 @@
...
+            If no walltime is provided and the queue's maximum walltime is unavailable, or
+            if the provided walltime exceeds the queue's maximum allowed walltime.
+        EnvironmentError
+            If neither `nodes` nor `cpus_per_node` are provided and the scheduler cannot
+            determine the system's CPUs per node automatically.
+        """
+
+        self.scheduler = scheduler
+        self.commands = commands
+        self.cpus = cpus
```
TomNicholas: Currently the user could change these values by assigning to them after `SchedulerJob` is instantiated. Making them properties that returned private versions (e.g. `._scheduler`) would guard against this.

TomNicholas (commented):
Splitting this review into parts as this PR is huge! I think the overall design seems good so the comments are mostly small.

dafyddstephenson:
> Splitting this review into parts as this PR is huge! I think the overall design seems good so the comments are mostly small.

Indeed - it is mostly docstrings, notebooks, and unit tests adding to the line count, most of the big-ticket changes are in 
`cstar/scheduler/job.py` and `cstar/system/scheduler.py` and should be quite digestible.

Generally speaking the thing that should be reviewed (as you've already begun previously) is the structure: if the structure checks out then the rest is largely just a reshuffling prioritising the separation of C-Star-related code from scheduler-related code. 

It will also be probably a lot easier to start by seeing how it's implemented in the doc pages (there is now a dedicated one related to this PR but I realise I didn't actually add it yet, notebook 5)

cstar/case.py
dafyddstephenson: I think we've touched on this before - personal preference, if it's not a problem. Separating logical expressions into little packages like this makes it much more readable to me, especially as they get complex.

cstar/case.py
dafyddstephenson: if `Scheduler` is `None` then `walltime` is never needed or used, I'd suggest `None` is the ideal way to signal this.

cstar/roms/component.py
dafyddstephenson: Agree, this would be better as an issue though as this specific block only occurs in the absence of a scheduler. I was thinking of adding a parameter to the `SchedulerJob.updates` function that performs this same parsing rather than streaming raw output. Would be better in a separate PR though.

cstar/roms/component.py
dafyddstephenson: This is another example of cleaning that is needed (and is covered by 88) but is outside the scope of this PR (as `cstar_sysmgr.scheduler` is `None` for this block to be reached). There will be quite a few of these to get around to, this PR being the first and biggest example.

cstar/roms/component.py
dafyddstephenson: again this relates to parsing the stdout of ROMS which is again a clean-up job independent of this PR.

cstar/scheduler/job.py
dafyddstephenson: Could also set it as `0` or something as a placeholder if using a fixed type is particularly useful. I think a raise is a bit aggressive here, it's valid to query the id of a job that doesn't have one yet and I'd expect a gentle reminder rather than an error.

dafyddstephenson:
[Notebook 5](https://c-star--175.org.readthedocs.build/en/175/5_handling_jobs_on_hpc_systems.html) now added.

I see a few comments related to [the parsing of the ROMS output ](https://github.com/CWorthy-ocean/C-Star/blob/cf35a94ba9d0d058765024e054e25eba695e67cb/cstar/roms/component.py#L979-L1031) so I'm not sure if this is showing up as new code in the diff - just as a heads up, this has been in for a good while, and isn't related to the scheduling refactor. #203 should cover this.

cstar/base/scheduler_job.py
dafyddstephenson: > I can do it now, no sweat

done in 5aef23d08166d293be4b81f23819dd11bd02755d and c05de1f3598a7a558e5d557c9fe69ddd8e6a1d72

cstar/roms/component.py
dafyddstephenson: Revisiting this code this isn't actually covered by issue 88: the "debugging" flag here is used to prevent C-Star parsing ROMS stdout in case something has gone wrong with ROMS.
This can be handled more gracefully, though - #203 should cover it.

cstar/roms/component.py
dafyddstephenson: Have opened #203 which should address this.

cstar/case.py
dafyddstephenson: Fixed in [5565978](https://github.com/CWorthy-ocean/C-Star/pull/175/commits/5565978be084fa5a05ab14c6b3c4ca362c0fe729)

cstar/case.py
dafyddstephenson: done in fda70d91b2a12f8a42560b67a0d648542825d8ff

cstar/scheduler/job.py
dafyddstephenson: Done in 139092a86d34e8a4492693b88e9fe2c6dc298e18

Issue #147, opened 2024-10-22
Static installation of `C-Star` fails


NoraLoose:
Our internal tester @ubbu36 has found the first bug!

Installing C-Star statically via
```
pip install .
```
seems to lead to an incomplete installation of C-Star. Typing
```
import cstar
```
leads to the follwing error
```
ModuleNotFoundError: No module named 'cstar.base'
```

Installing C-Star dynamically (which is what the developers have done) via
```
pip install -e .
```
works fine.

TomNicholas:
I can't seem to reproduce this locally, but if anyone can, some possible culprits:

- [ ] The `pyproject.toml` is slightly different to others I've looked at, ours includes the lines
    
    ```
    # Automatically find the packages
    where = ["."]
    include = ["cstar"]
    ```
    
    which we might want to try removing

- [ ] There needs to be a `__init__.py` file in every directory that contains any python code that could be imported, but there appear to be some missing from the test file directories (`unit_tests`)

PR #176, opened 2024-12-02
Build distribution and upload to PyPI

NoraLoose:
This PR sets up a github workflow that builds C-Star locally, which creates distribution archives under the `dist/` directory. These are the files that will be uploaded to PyPI when publishing a new release.


dafyddstephenson:
Thanks Nora! Are we able to test this before merging? I think #147 might be potentially problematic here

NoraLoose:
@dafyddstephenson definitely! The "Build distribution / test-built-dist" test that fails in the CI is that exact test. I think it fails for the same reason as #147.

NoraLoose:
Obviously we won't merge before this issue is resolved. :)

dafyddstephenson:
ah sorry, I'm debilitated by jetlag right now! I saw the workflow is only triggered on push to main and thought we'd have to merge to check it - didn't notice it was also on PR. I was hoping #168 might fix the static install issue as there's no longer a dynamically generated python file in the configuration, but no dice...

NoraLoose:
@TomNicholas once #147 is fixed, all tests in this PR should pass. So we could try to fix the issue as part of this PR?

TomNicholas (approved):

TomNicholas:
#179, #180, and #181 together seemed to fix this

PR #181, opened 2024-12-05
Fix __version__ attribute

TomNicholas:
Try following the same choices as virtualizarr does in it's [`pyproject.toml`](https://github.com/zarr-developers/VirtualiZarr/blob/main/pyproject.toml)


TomNicholas:
That works in the sense that `cstar.__version__` returns '0.0.0' for me locally. I'm a bit confused why it's `0.0.0` and not the fallback I set of "9999", but this is really irrelevant because we shouldn't expect this to be the correct number until we actually have at least one version tag defined in the repo.

PR #190, opened 2024-12-05
Add table of which machines C-Star has been tested on

### LINKED ISSUE #186: have a table of which machines we have tried running C-Star on

TomNicholas:
- [x] Closes #186

### DIFF
--- a/docs/machines.md
+++ b/docs/machines.md
@@ -1,4 +1,4 @@
-# Tested Machines
+# Machines
 
 C-Star aims to be deployable on a wide range of HPC systems, but so far it has been tested on a limited number.
 
--- a/docs/machines.md
+++ b/docs/machines.md
@@ -1,0 +1,10 @@
+# Tested Machines
+
+C-Star aims to be deployable on a wide range of HPC systems, but so far it has been tested on a limited number.
+
+| Name of system | Testing Status | Notes |
+| -------------- | -------------- | ----- |
+| Local Mac      | Tested         |       |
+| Perlmutter     | Tested         |       |
+| Expanse        | Tested         |       |
+| Derecho        | Tested         |       |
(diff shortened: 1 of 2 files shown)


docs/machines.md
```diff
@@ -0,0 +1,10 @@
+# Tested Machines
+
+C-Star aims to be deployable on a wide range of HPC systems, but so far it has been tested on a limited number.
+
+| Name of system | Testing Status | Notes |
+| -------------- | -------------- | ----- |
+| Local Mac      | Tested         |       |
```
dafyddstephenson: Would call this MacOS ARM64 (Apple Silicon)

docs/machines.md
```diff
@@ -0,0 +1,10 @@
...
+| Perlmutter     | Tested         |       |
```
dafyddstephenson: Would go for NERSC Perlmutter, NCAR Derecho, SDSC Expanse. Would also suggest hyperlinking the names to the machines' homepages:

https://www.sdsc.edu/support/user_guides/expanse.html
https://docs.nersc.gov/systems/perlmutter/architecture/
https://ncar-hpc-docs.readthedocs.io/en/latest/compute-systems/derecho/

dafyddstephenson (changes requested):
Nice one, can't actually see it in situ as it looks like the RTD build is hanging, but these would be my suggestions assuming all looks good in the rendered version

docs/machines.md
TomNicholas: added in https://github.com/CWorthy-ocean/C-Star/pull/190/commits/5cd28f026ce7251a22550369418ca543ea8cb77a

docs/machines.md
TomNicholas: added in https://github.com/CWorthy-ocean/C-Star/pull/190/commits/5cd28f026ce7251a22550369418ca543ea8cb77a

PR #192, opened 2024-12-06
Release notes for v0.0.1-alpha

TomNicholas:


docs/releases.rst
```diff
@@ -1,12 +1,19 @@
...
 
-v0.0.1 (unreleased)
--------------------
+v0.0.1-alpha (6th Dec 2024)
+---------------------------
 
-First release of C-Star!
+The first release of C-Star!
+
+This release provides basic functionality, including the ability to create, import, and export "blueprints" for reproducible ocean model simulations (C-Star Cases) using supported ocean models (ROMS, optionally with MARBL biogeochemistry) and run those Cases locally or on supported HPC systems (via Slurm and PBS).
```
NoraLoose: We could link "blueprints", "Cases" etc to the glossary?

NoraLoose: Similar to how we did here:
https://github.com/CWorthy-ocean/C-Star/blob/aeee76cd93aaa83b0195940719f0754cb52bc9a2/docs/index.rst?plain=1#L21

NoraLoose: We should probably use local paths rather than linking to the external RTD URL.

TomNicholas: We could, I only didn't include hyperlinks because I wasn't sure if there was some place they might render weirdly (e.g. PyPI).

NoraLoose: PyPI doesn't show the documentation, it only links to the repo and the repo README (which in turn links to the RTD).

TomNicholas: Added local links in https://github.com/CWorthy-ocean/C-Star/pull/192/commits/857af7a9c2a7a5aea8b4adbbd6dd1ef57dc5df73, which seem to work in the RTD build.

PR #194, opened 2024-12-07
Use PyPI token in publishing workflow

NoraLoose:


TomNicholas:
Seems good to me, but the CI that would test this is skipped right? Feel free to merge

NoraLoose:
Yep, it's one of these things... we won't know until we merge.

PR #212, opened 2025-01-03
Persist Case instances between sessions

### LINKED ISSUE #158: Persist local state between sessions (make separate `Case.persist()` and `Case.to_blueprint()` methods)
## Issue:
Currently, there is no support for persisting C-Star objects in their current state between python sessions.

## Example
If we run
```
Case.from_blueprint(blueprint="my_blueprint.yaml", caseroot=Path("my/dir"))
Case.setup()
```
Then `Case.is_setup` returns `True` as expected.
However, if you exit this session and run
```
Case.from_blueprint(blueprint="my_blueprint.yaml", caseroot=Path("my/dir"))
Case.is_setup
```
we now get `False`, even though nothing has changed on the filesystem. This is because the `setup` method also calls methods like `get` on `InputDataset` and `AdditionalCode` objects that set attributes like `InputDataset.working_path` that are related to the _local state_ of the `Case`, but aren't relevant information to include in a blueprint file. These attributes are thus lost when exiting the session, unless the user runs `Case.setup()` again. The local working file associated with an `InputDataset` may be different to the source version (for instance, local netCDF generated from remote yaml), so the distinction between the source material (described in the blueprint) and the local state is nontrivial. 

## Proposed solution:

- [x] Rename the current `Case.persist()` method to `Case.to_blueprint()`. This is consistent with the existing `Case.from_blueprint()` complement method.
- [x] Add some sort of `local_file_hash` attribute that is set when `get()` is called on objects like `InputDatasets` and `AdditionalCode`. This becomes part of the state to be reloaded and can verify that files on disk have not changed between sessions
~~- [ ] Add an `include_state=True/False` parameter on the `to_dict` methods of fundamental classes. `to_dict` is used heavily by `Case.persist()` but currently just includes information for recreating the object from scratch (as is required for blueprint export), not restoring it to its current state.~~
- [ ] Define a new `Case.persist()` that exports the entire local state to a ~~`{caseroot}/{case_name}_state.yaml`~~ `{caseroot/{case_name}_state.pkl` file. This could be called automatically at the successful end of certain methods that update the state, like `Case.setup()`
- [ ] Define a `Case.restore(caseroot="wherever")` method that looks in the caseroot for the state file and re-establishes the exact state that was persisted.  

UPDATE 31/12/24:
For some reason or another I decided a while back not to use `pickle` to handle this issue (`Case` objects may simply have not been pickle-able, but as of now they are). Pickling is undoubtedly the most sensible way forward, so the (now struck-through) task above describing adding `include_state` to `to_dict` is irrelevant. The other tasks should still be completed as described.

dafyddstephenson:
- [x] Closes #158 
- [x] Full type hint coverage

- Adds `Case.persist()` method, called automatically in methods that change the instance's state. The method saves the `Case` instance to `Case.caseroot/case_state.pkl`
- Adds `Case.restore(caseroot = "/my/caseroot")` class method, that loads the `case_state.pkl` file from the `caseroot` directory
- Modifies `Case.__init__` to check whether the chosen `caseroot` is an empty directory. If not, the user is prompted to use `Case.restore` to restore a previously created `Case` instance


dafyddstephenson:
This is now failing as since merging #206 , the private attribute `Case.components[1]._execution_handler._process` is set to a `subprocess.Popen` object after calling `Case.run()`, which includes a `_thread.lock` object, which cannot be pickled. Moving to WIP for now, will attack on Monday.

Issue #6, opened 2024-07-24
Python library for handling git commands


TomNicholas:
> I bet there is some python library for handling git commands that would do this for us

_Originally posted by @TomNicholas in https://github.com/CWorthy-ocean/C-Star/pull/24#discussion_r1673092633_

PR #235, opened 2025-03-31
Unify calls to subprocess.run

### LINKED ISSUE #204: Create helper function to wrap subprocess.run
Currently, the following pattern:

```
        sp_cmd = 'echo 'Hello World"'
        sp_run = subprocess.run(sp_cmd, shell=True, text=True, capture_output=True)
        if sp_run.returncode != 0:
            raise RuntimeError(f"Command {sp_cmd} failed: {sp_run.stderr.strip()}")
        hw = sp_run.stdout.strip()
```

Can be found all over the place in C-Star. 
It would be fairly straightforward to wrap this behaviour in a helper function, reducing each such block to, e.g.:

`hw = run_subprocess('echo "Hello World"')`

ankona:
This PR wraps up the common pattern of calling `subprocess.run` with a utility method. The new method includes:

- always-on logging before start w/message customization
- always-on logging after process term w/message customization
- always-on logging on a non-zero return code w/message customization

> [!NOTE] 
> Out-of-Scope
> Calls to `subprocess.run` related to git repository manipulation are not replaced here. They will be completed in [another ticket](https://github.com/CWorthy-ocean/C-Star/issues/6).

- [x] Closes #204
- [x] Tests updated
- [x] Tests passing
- [x] Full type hint coverage
- [ ] New functions/methods are listed in `api.rst` - `n/a`
- [ ] New functionality has documentation - `n/a`

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -13,6 +13,10 @@ jobs:
   test:
     name: integration-${{ matrix.python-version }}
     runs-on: ubuntu-latest
+
+    # temporarily disable breaking integration tests from blocking a build
+    continue-on-error: true
+
     strategy:
       matrix:
         python-version: ["3.12",]
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -14,9 +14,6 @@ jobs:
     name: integration-${{ matrix.python-version }}
     runs-on: ubuntu-latest
 
-    # temporarily disable breaking integration tests from blocking a build
-    continue-on-error: true
-
     strategy:
       matrix:
         python-version: ["3.12",]
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -13,7 +13,6 @@ jobs:
   test:
     name: integration-${{ matrix.python-version }}
     runs-on: ubuntu-latest
-
     strategy:
       matrix:
         python-version: ["3.12",]
--- a/cstar/base/utils.py
+++ b/cstar/base/utils.py
@@ -378,17 +378,17 @@ def _run_cmd(
 
     Examples:
     --------
-    In: _run_cmd("python foo.py", "Running script", "Script completed")
+    In: _run_cmd("python foo.py", msg_pre="Running script", msg_post="Script completed")
     Out: Running script
          Script completed
 
     In: _run_cmd("python foo.py")
     Out: Running command: python foo.py
-         Command completed successfully. STDOUT: <output of foo.py>
+         Command completed successfully.
 
     In: _run_cmd("python return_nonzero.py")
     Out: Running command: python return_nonzero.py
-         Command python return_nonzero.py failed: <stderror of foo.py>
+         Command `python return_nonzero.py` failed. STDERR: <stderror of foo.py>
     """
     print(msg_pre or f"Running command: {cmd}")
     stdout: str = ""
@@ -421,5 +421,5 @@ def _run_cmd(
 
         print(msg)
 
-    print(msg_post or f"Command completed successfully. STDOUT: {result.stdout}")
+    print(msg_post or "Command completed successfully.")
     return stdout
--- a/cstar/base/utils.py
+++ b/cstar/base/utils.py
@@ -367,8 +367,7 @@ def _run_cmd(
         An overridden message logged after the command is successfully executed.
     msg_err (str | None), default = None):
         An overridden message logged when a command returns a non-zero code. Logs
-        will be supplied with the subprocess output using the name `result`,
-        e.g. `msg_err.format(result=result)`.
+        will automatically append the stderr output of the command.
     raise_on_error (bool, default = False):
         If True, raises a RuntimeError if the command returns a non-zero code.
 
@@ -413,9 +412,9 @@ def _run_cmd(
 
     if result.returncode != 0:
         if msg_err:
-            msg = msg_err.format(result=result)
+            msg = f"{msg_err} STDERR: {result.stderr.strip()}"
         else:
-            msg = f"Command `{cmd}` failed: {result.stderr.strip()}"
+            msg = f"Command `{cmd}` failed. STDERR: {result.stderr.strip()}"
 
         if raise_on_error:
             raise RuntimeError(msg)
--- a/cstar/base/utils.py
+++ b/cstar/base/utils.py
@@ -344,7 +344,7 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
 
 def _run_cmd(
     cmd: str,
-    cwd: PathLike | None = None,
+    cwd: Path | None = None,
     env: dict[str, str] | None = None,
     msg_pre: str | None = None,
     msg_post: str | None = None,
@@ -357,7 +357,7 @@ def _run_cmd(
     -----------
     cmd (str):
        The command to be executed as a separate process.
-    cwd (PathLike, default = None):
+    cwd (Path, default = None):
        The working directory for the command. If None, use current working directory.
     env (Dict[str, str], default = None):
        A dictionary of environment variables to be passed to the command.
--- a/cstar/base/utils.py
+++ b/cstar/base/utils.py
@@ -1,7 +1,9 @@
 import re
+import functools
 import hashlib
 import warnings
 import subprocess
+from os import PathLike
 from pathlib import Path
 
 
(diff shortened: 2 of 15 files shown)


cstar/base/utils.py
```diff
@@ -338,3 +340,87 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
                 tree_str += f"{prefix}{sub_prefix}{item_branch}{item}\n"
 
     return tree_str
+
+
+def _run_cmd(
```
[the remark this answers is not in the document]
ankona: Fixed incorrect examples in docstring

cstar/base/utils.py
```diff
@@ -338,3 +340,87 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
...
+        If True, raises a RuntimeError if the command returns a non-zero code.
+
+    Returns:
+    -------
+    stdout: str
+       The captured standard output of the process.
+
+    Examples:
+    --------
+    In: _run_cmd("python foo.py", "Running script", "Script completed")
```
dafyddstephenson: Mismatch with the function signature

cstar/base/utils.py
```diff
@@ -338,3 +340,87 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
...
+
+    In: _run_cmd("python foo.py")
+    Out: Running command: python foo.py
+         Command completed successfully. STDOUT: <output of foo.py>
+
+    In: _run_cmd("python return_nonzero.py")
+    Out: Running command: python return_nonzero.py
+         Command python return_nonzero.py failed: <stderror of foo.py>
+    """
+    print(msg_pre or f"Running command: {cmd}")
```
dafyddstephenson, line 393: I think the default should be not to print anything unless there's an error - subprocess is called a lot in C-Star and could clutter the screen (though you may be handling this separately in the logging PR)

cstar/base/utils.py
```diff
@@ -338,3 +340,87 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
...
+            msg = msg_err.format(result=result)
+        else:
+            msg = f"Command `{cmd}` failed: {result.stderr.strip()}"
+
+        if raise_on_error:
+            raise RuntimeError(msg)
+
+        print(msg)
+
+    print(msg_post or f"Command completed successfully. STDOUT: {result.stdout}")
```
dafyddstephenson: As with my above comment re. a default print statement, I would suggest against printing STDOUT as the default if nothing has gone wrong, we're hoping to hide all the internals of most things from the user, but again this is probably addressed in the `logging` PR. I was imagining something like logging all these things somewhere an advanced user could go looking for them

cstar/execution/scheduler_job.py
```diff
@@ -686,30 +683,21 @@ def submit(self) -> Optional[int]:
...
 
-        result = subprocess.run(
+        stdout = _run_cmd(
             f"sbatch {self.script_path}",
-            shell=True,
             cwd=self.run_path,
             env=slurm_env,
-            capture_output=True,
-            text=True,
+            msg_err="Non-zero exit code when submitting job. STDERR: \n{result.stderr}",
```
dafyddstephenson: This pattern with `result` feels a little unintuitive to me. I think my instincts say to write whatever message and then have the `STDERR: \n{result.stderr}` part baked into the `_run_cmd` itself and always printed (if there's an error in a subprocess, something's gone sufficiently wrong that they should always get the full STDERR afaics)

cstar/base/utils.py
```diff
@@ -338,3 +340,87 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
...
+        kwargs["cwd"] = cwd
+    if env:
+        kwargs["env"] = env
+
+    result: subprocess.CompletedProcess[str] = fn(**kwargs)
+    stdout = str(result.stdout).strip() if result.stdout is not None else ""
+
+    if result.returncode != 0:
+        if msg_err:
+            msg = msg_err.format(result=result)
```
dafyddstephenson: have made a comment below on this somewhere but I think it's better to replace the `STDERR: {result.stderr}` in `msg_err` in the call itself with a standard pattern of appending `STDERR: {result.stderr}` to _any_ message that is passed through by the call, so we don't have this slightly unintuitive pattern of passing an undefined variable (in the scope of the caller) `result`  from the caller to `_run_cmd` for the latter to handle.

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
```diff
@@ -135,12 +135,12 @@ def test_make_failure(self, marbl_codebase, tmp_path):
 
         ## There are two subprocess calls, we'd like one fail, one pass:
         self.mock_subprocess_run.side_effect = [
-            mock.Mock(returncode=1, stderr="Compiling MARBL failed successfully"),
```
dafyddstephenson, line 138: aw cmon leave the jokes in the unit tests 😂

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
```diff
@@ -135,12 +135,12 @@ def test_make_failure(self, marbl_codebase, tmp_path):
...
+            mock.Mock(returncode=1, stderr="Mocked MARBL Compilation Failure"),
         ]
 
         # Test
         with pytest.raises(
             RuntimeError,
-            match="Error 1 when compiling MARBL. STDERR stream: \n Compiling MARBL failed successfully",
```
dafyddstephenson, line 144: 🥺

dafyddstephenson (changes requested):
Thanks for powering through this Chris! Couple of minor suggestions but the biggest one is I think to print STDERR on error from within `_run_cmd` , rather than passing it through the caller. I'll also run through the examples on a couple of different systems to check if everything's consistent before merging.

cstar/execution/scheduler_job.py
ankona: I agree. I wanted to hear somebody else say the same before I tossed it and limited callers to a static error message, though. Removing...

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
ankona: Looked like a copypasta error!

cstar/base/utils.py
ankona: Adjusted existing success msg.

cstar/base/utils.py
ankona: It should be a `log.debug` in the logging PR. Note that this behavior of logging a preamble pre-dated this PR but was not consistent. If you don't want the ability to add the preamble (and the postamble), I can remove them.

cstar/base/utils.py
ankona: updated examples w/kwargs

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
ankona: Updated to match tweaked log output

.github/workflows/integration_tests.yaml
```diff
@@ -13,6 +13,10 @@ jobs:
   test:
     name: integration-${{ matrix.python-version }}
     runs-on: ubuntu-latest
+
+    # temporarily disable breaking integration tests from blocking a build
```
dafyddstephenson: Is this "temporarily" meant to stay in the PR?

cstar/base/utils.py
dafyddstephenson: this hasn't updated for me

cstar/base/utils.py
dafyddstephenson: I think adding a pre-amble is worthwhile, but the `or` should just be to not print anything.

dafyddstephenson (changes requested):
Hi Chris, looks like nothing new has been pushed since my last review? Unless I'm missing anything.

dafyddstephenson:
Btw, if you merge the current main it should fix the integration test failure :) it's just a hardcoded shasum that's updated

.github/workflows/integration_tests.yaml
ankona: It was meant to stay if I had to merge it. Per your other comment, I'm going to remove it now and 🤞 that main works.

ankona:
> Hi Chris, looks like nothing new has been pushed since my last review? Unless I'm missing anything.

Great catch. I was flipping back and forth between branches and didn't push. It should be updated now w/the fix noted as not updated (example call to _run_cmd) and removal of `continue-on-fail` for integration tests

cstar/base/utils.py
ankona: I'm going to make it `log.debug` in the other task so it's not always happening. I've updated that ticket w/the requirement and resolving this.

cstar/tests/unit_tests/system/test_scheduler.py
```diff
@@ -311,6 +311,7 @@ def test_pbsscheduler_global_max_cpus_per_node_failure(
 
         Captures printed error messages to ensure the failure is logged correctly.
         """
+        capsys.readouterr()  # Clear previous output
```
ankona: removed this unnecessary change

cstar/base/utils.py
```diff
@@ -338,3 +340,86 @@ def _dict_to_tree(input_dict: dict, prefix: str = "") -> str:
...
+        kwargs["env"] = env
+
+    result: subprocess.CompletedProcess[str] = fn(**kwargs)
+    stdout = str(result.stdout).strip() if result.stdout is not None else ""
+
+    if result.returncode != 0:
+        if msg_err:
+            msg = f"{msg_err} STDERR: {result.stderr.strip()}"
+        else:
+            msg = f"Command `{cmd}` failed. STDERR: {result.stderr.strip()}"
```
dafyddstephenson: Looks good, I'd just suggest a `\n` before the `{result.stderr.strip()}`

dafyddstephenson (approved):
Looks good! One minor suggestion

cstar/base/utils.py
ankona: done

PR #250, opened 2025-04-29
Avoid distributed formatting of ~/.cstar.env file

ankona:
Deprecate use of custom `_update_user_dotenv` method for writing client-supplied strings into user's `~/.cstar.env`. Replaced `_update_user_dotenv` with `dotenv.set_key` and updated tests as needed for change to env var setting 

- [x] Closes #CW-808
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/roms/external_codebase.py
```diff
@@ -40,9 +39,12 @@ def _codebase_adjustments(self) -> None:
...
         computing systems.
         """
+        roms_root = Path(
+            cstar_sysmgr.environment.environment_variables[self.expected_env_var]
+        )
         shutil.copytree(
-            Path(os.environ[self.expected_env_var]) / "ci/ci_makefiles/",
-            os.environ[self.expected_env_var],
+            roms_root / "ci/ci_makefiles/",
+            str(roms_root),
```
dafyddstephenson: don't think this needs to be a `str` ?

cstar/marbl/external_codebase.py
```diff
@@ -52,12 +51,7 @@ def get(self, target: str | Path) -> None:
             checkout_target=self.checkout_target,
         )
         # Set environment variables for this session:
-        os.environ["MARBL_ROOT"] = str(target)
-        cstar_sysmgr.environment.environment_variables["MARBL_ROOT"] = os.environ[
-            "MARBL_ROOT"
-        ]
-        env_file_str = f'MARBL_ROOT="{target}"\n'
-        _update_user_dotenv(env_file_str)
+        cstar_sysmgr.environment.set_key(self.expected_env_var, f'"{str(target)}"')
```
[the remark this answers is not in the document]
ankona: leaving alone pending verification that no MARBL paths are expected to have quotes

ScottEilerman (approved):

cstar/marbl/external_codebase.py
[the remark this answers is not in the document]
dafyddstephenson: Looking at this side-by-side with `main` I am unclear as to why `target` needs to be so deeply wrapped here. What happens if you just use `str(target)`?

cstar/roms/external_codebase.py
```diff
@@ -72,17 +75,11 @@ def get(self, target: str | Path) -> None:
...
-        ]
-        os.environ["PATH"] = os.environ.get("PATH", "") + f":{target}/Tools-Roms/"
-        # os.environ["PATH"] += f":{target}/Tools-Roms/"
-        cstar_sysmgr.environment.environment_variables["PATH"] = os.environ["PATH"]
-        env_file_str = (
-            f"ROMS_ROOT={target}" + "\nPATH=${PATH}:" + f"{target}/Tools-Roms\n"
+        cstar_sysmgr.environment.set_key(self.expected_env_var, str(target))
+        curr_path_value = os.environ.get("PATH", "")
+        cstar_sysmgr.environment.set_key(
+            "PATH", f"{curr_path_value}:{target}/Tools-Roms/"
```
dafyddstephenson: I think this may actually be problematic. In the previous version, we set the active `PATH` variable value (via `os.environ`) by expanding it and appending to it, which made sense in the context of the session. Meanwhile, in `.cstar.env`, we wrote the _un-expanded_ string `$PATH:{target}/Tools-Roms/`, such that next time the environment was loaded from the dotenv file, which could be years later, the user's newest value of `PATH` would be appended to. In this version, due to expansion, the current value of `PATH` appears to be immortalised in the dotenv file.

cstar/system/environment.py
```diff
@@ -240,3 +242,17 @@ def load_lmod_modules(self, lmod_file) -> None:
             lmod_list = F.readlines()
             for mod in lmod_list:
                 self._call_lmod(f"load {mod}")
+
+    def set_key(self, key: str, value: str) -> None:
```
dafyddstephenson: I would suggest renaming this - I had to do a triple take to recognise that it wasn't overriding the imported `set_key`, it's a bit confusing.

cstar/tests/integration_tests/test_cstar_test_blueprints.py
```diff
@@ -55,39 +57,52 @@ def test_cstar(
             Logger instance for logging messages during test execution.
         """
 
-        # Regardless of remote or local, if yaml_datasets we need roms-tools support data
-        if "yaml_datasets" in test_config_key:
-            fetch_roms_tools_source_data(symlink_path="roms_tools_datasets")
+        dotenv_path = tmpdir / ".cstar.env"
+        ext_root = tmpdir / "externals"
```
dafyddstephenson, line 61: Appreciate this - was an overdue improvement to the testing framework that I never got around to as it always starts from a clean env on GitHub. Cheers!

dafyddstephenson (changes requested):
Hey, I believe one major change is needed re. expansion, and some more minor changes. Haven't thoroughly reviewed the updated unit tests yet but they look ok and I can take a deeper look on the next pass.

cstar/roms/external_codebase.py
ankona: w00t! fixed.

cstar/marbl/external_codebase.py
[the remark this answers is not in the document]
ankona: removed unnecessary cast and removed all quotes (dotenv is already adding quotes automatically)

cstar/system/environment.py
ankona: renamed to `set_env_var`

cstar/roms/external_codebase.py
ankona: replaced manual expansion by using `dotenv.load_dotenv` after `dotenv.set_key` to ensure variables are correctly expanded in current env

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
```diff
@@ -105,40 +106,59 @@ def test_get_success(self, marbl_codebase, tmp_path):
...
-            local_path=marbl_dir,
-            checkout_target=marbl_codebase.checkout_target,
-        )
-
-        ## Check that _update_user_dotenv was (mock) called correctly
-        env_file_str = f'MARBL_ROOT="{marbl_dir}"\n'
-        self.mock_update_user_dotenv.assert_called_once_with(env_file_str)
+        dotenv_path = tmp_path / ".cstar.env"
+        key = "MARBL_ROOT"
+        value = str(marbl_dir)
```
dafyddstephenson, line 111: Very minor suggestion but for future comprehension I would set these above where `marbl_dir` is set and possibly rename `key -> expected_env_var_name` and `value -> expected_env_var_value`.

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
```diff
@@ -105,40 +106,59 @@ def test_get_success(self, marbl_codebase, tmp_path):
...
+            ),
+            mock.patch.dict(os.environ, {key: "old-value"}),
+        ):
+            # Test
+            ## Call the get method
+            marbl_codebase.get(target=marbl_dir)
+
+            # Assertions:
+            ## Check environment variables
+            cstar_sysmgr.environment.set_env_var(key, value)
```
dafyddstephenson: I don't think we should have to set this manually in the test - `marbl_codebase.get()` should now be doing it after your other modifications? That seems to be how the equivalent `ROMSExternalCodebase` test shakes out

cstar/tests/unit_tests/roms/test_roms_external_codebase.py
```diff
@@ -109,57 +107,79 @@ def teardown_method(self):
...
+        roms_dir = tmp_path / "roms"
+        dotenv_path = tmp_path / ".cstar.env"
+
+        with (
+            mock.patch(
+                "cstar.system.environment.CSTAR_USER_ENV_PATH",
+                dotenv_path,
+            ),
+            mock.patch.dict(os.environ, {}) as mock_env,
+        ):
```
dafyddstephenson: as this covers the whole test it might be better as two decorators?

cstar/tests/unit_tests/roms/test_roms_external_codebase.py
```diff
@@ -109,57 +107,79 @@ def teardown_method(self):
...
+            cfg = dotenv.dotenv_values(dotenv_path)
+
+            # confirm user environment file was updated
+            for k, v in [(k0, v0), (k1, v1)]:
+                actual_value = cfg[k]
+                if ":" in v:
+                    assert len(actual_value.split(":")) >= len(v.split(":"))
+                    assert v.split(":")[1] in actual_value
+                else:
+                    assert v in actual_value
```
dafyddstephenson: this is neat but a bit hard to parse. As there are only two k/v pairs here could we just explicitly check against them?

cstar/tests/unit_tests/system/test_environment.py
```diff
@@ -168,28 +171,127 @@ def test_env_file_loading(self, mock_load_lmod, mock_uses_lmod, tmp_path):
         )
         # Write simulated user .env content with an overriding variable
         user_env_file_path.write_text(
-            "MPIROOT=/override/mock/mpi\n" "CUSTOM_VAR=custom_value\n"
+            "MPIROOT=/override/mock/mpi\nCUSTOM_VAR=custom_value\n"
```
dafyddstephenson: nitpicky but unless a linter did this I'd say it's less readable

cstar/tests/unit_tests/system/test_environment.py
```diff
@@ -168,28 +171,127 @@ def test_env_file_loading(self, mock_load_lmod, mock_uses_lmod, tmp_path):
...
+    @patch.object(
+        cstar.system.environment.CStarEnvironment, "uses_lmod", return_value=True
+    )
+    @patch.object(
+        cstar.system.environment.CStarEnvironment,
+        "load_lmod_modules",
+        return_value=None,
+    )
+    def test_env_file_updating(self, mock_load_lmod, mock_uses_lmod, tmp_path: Path):
+        """Tests that environment variables are updated correctly in the user .env file.
```
dafyddstephenson, line 201: Could you add this one-liner to the `Tests` section of the class-level docstring also?

cstar/tests/unit_tests/system/test_environment.py
```diff
@@ -168,28 +171,127 @@ def test_env_file_loading(self, mock_load_lmod, mock_uses_lmod, tmp_path):
...
+
+        Mocks
+        -----
+        - tmp_path creates temporary, emulated system and user .env files
+        - CStarEnvironment.package_root is patched to the temporary directory to load these .env files
+        - CStarEnvironment.uses_lmod is patched to always be true
```
dafyddstephenson: I don't think this should be needed, was it necessary to pass?

cstar/tests/unit_tests/system/test_environment.py
```diff
@@ -168,28 +171,127 @@ def test_env_file_loading(self, mock_load_lmod, mock_uses_lmod, tmp_path):
...
+            patch.object(
+                cstar.system.environment.CStarEnvironment, "package_root", new=root_path
+            ),
+            patch(
+                "cstar.system.environment.CSTAR_USER_ENV_PATH",
+                user_env_file_path,
+            ),
+        ):
+            # Instantiate the environment to trigger loading the environment variables
+            env = CStarEnvironment("perlmutter", "mpi-prefix", "gnu")
```
dafyddstephenson:
```suggestion
            env = CStarEnvironment(system_name="perlmutter", mpi_exec_prefix="mpi-prefix", compiler="gnu")
```

cstar/tests/unit_tests/system/test_environment.py
```diff
@@ -168,28 +171,127 @@ def test_env_file_loading(self, mock_load_lmod, mock_uses_lmod, tmp_path):
...
+
+            # Confirm variables that are not overriden by the user are not
+            # accessible via CStarEnvironment
+            for key in ["NETCDFHOME", "MPIHOME"]:
+                assert key not in env.environment_variables
```
dafyddstephenson: I think I'm missing something, but why would these not be accessible?
If I copy your above `env = ` line on my laptop I can access both `perlmutter`-specific variables and the contents of my `~/.cstar.env` via `env.environment_variables`, which should be correct.

on my laptop I see both system and user vars, which I believe is correct?

dafyddstephenson (changes requested):
This is a waaaay cleaner implementation, thanks a bunch. Couple of nitpicky comments and one potential red flag in the new unit test that I might need explained to me, otherwise good to go.

cstar/tests/unit_tests/marbl/test_marbl_external_codebase.py
ankona: Good catch - i forgot to go back and remove things tied to the custom writes... Removed

cstar/tests/unit_tests/roms/test_roms_external_codebase.py
ankona: 1. os.environ dict was already patched in `setup_method` so I removed the dupe. 
2. not sure that i can get to the `tmp_path` fixture from a decorator. left `"cstar.system.environment.CSTAR_USER_ENV_PATH"` as-is

cstar/tests/unit_tests/roms/test_roms_external_codebase.py
ankona: I don't know what i was thinking... maybe I had more things to check? Silly code removed.

cstar/tests/unit_tests/system/test_environment.py
ankona: added `# fmt: skip` so it wouldn't look funky when placing two strings onto a single line

cstar/tests/unit_tests/system/test_environment.py
ankona: Added for now, but manually duplicating comments has got to go. I created a ticket to go back and look at why it's there (e.g. is it just a missing documentation generation option). See [CW-812](https://cworthy.atlassian.net/browse/CW-812)

cstar/tests/unit_tests/system/test_environment.py
ankona: Copypasta fail. Removed

cstar/tests/unit_tests/system/test_environment.py
ankona: added keywords

cstar/tests/unit_tests/system/test_environment.py
ankona: `mock_system.env` doesn't match `perlmutter.env` so no system level config is loaded. I wanted to say "hey, i didn't find a system level file so you won't find any variables from there" but it's obtuse.

Changed to just verify system exists, as noted. Also, added some fixtures to avoid repeated creation of `.../.cstar.env`, and `system.env`

cstar/tests/unit_tests/conftest.py
```diff
@@ -8,3 +9,45 @@
...
+def system_dotenv_dir(tmp_path: pathlib.Path) -> pathlib.Path:
+    # A path to a temporary directory for writing system-level
+    # environment configuration file
+    return tmp_path / "additional_files" / "env_files"
+
+
+@pytest.fixture
+def mock_system_name() -> str:
+    # A name for the mock system/platform executing the tests.
+    return "mock_system"
```
dafyddstephenson, line 42: In line with the below, perhaps this should take an optional arg and return `"mock_system"` by default if not supplied?

dafyddstephenson (approved):
all good, let's goo

cstar/tests/unit_tests/conftest.py
ankona: We shouldn't call fixture functions directly, so parameterizing isn't ideal. In the case that we want some other name, I think we can already achieve that by passing directly using `pytest.parametrize` instead of using the fixture.

dafyddstephenson: fair ✅

PR #252, opened 2025-05-01
Partitioning refactor

### LINKED ISSUE #214: Prevent unnecessary repeated partitioning
As part of #158 , several PRs have been opened to better track files handled by C-Star and prevent unnecessary repetition of potentially expensive steps:
- #209 caches file hashes for files obtained by `InputDataset.get()` to prevent them being fetched multiple times
- #211 does the same for `AdditionalCode` 
- #213 prevents `Component.build()` from compiling already compiled executables where it is clear the source code has not changed
- #212 removes the need to run `Case.setup()` again after exiting a python session

These cover the `Case.setup()` and `Case.build()` methods, but `Case.pre_run()` will still start from scratch any time it is called.

After merging #209 and #210 , we should add tracking for files produced by `partition` and introduce a check in `ROMSInputDataset.partition` that skips over unnecessary regeneration of already partitioned files.

### LINKED ISSUE #247: Allow `ROMSInputDataset` to accept partitioned source files
I believe the best way to do this would be to initialise the `ROMSInputDataset` as usual, but add a `partitioned_source: Optional[tuple[int,int]]` kwarg to `__init__`. The tuple would represent ROMS' `np_xi` and `np_eta` parameters, though we could also just have `source_np_xi` and `source_np_eta` directly instead. The `source.location` would still be a single filepath/URL, e.g. `https://www.dodgynetcdffiles2go.ru/totally_legit_roms_grid.nc`, but `ROMSInputDataset.get()` would download `"totally_legit_roms_grid.{tile}.nc" for tile in range(0,np_xi*np_eta)`, similarly to how ROMS currently interprets `roms.in` (is given an unparitioned filepath, but looks for partitioned files).

If one attempts to create a `ROMSSimulation` containing a `ROMSInputDataset` whose source partitioning does not match the simulation partitioning, an error would be raised during `ROMSSimulation.__init__()`.

@ScottEilerman thoughts?

dafyddstephenson:
This PR makes some internal changes to partitioning:

- introduces a `ROMSPartitioning` class 
- replaces `ROMSInputDataset.partitioned_files` with `ROMSInputDataset.partitioning` (an instance of the above class) to track locally partitioned netCDF files more effectively than just listing them
- defines several utility functions to modularise methods that fetch and partition datasets.
- Adds (`Optional[int]`) `source_np_xi` and `source_np_eta` parameters to `ROMSInputDataset.__init__` to specify that the source of a dataset is already partitioned, allowing users to supply pre-partitioned data to instantiate `ROMSInputDataset`s (#247)
- Adds checks that allow C-Star to skip unnecessary repeat calls to `pre_run` or `partition` (#214)
- Catch use cases like attempting to partition data multiple times with different parameters 

- [x] Closes #247 
- [x] Closes #214 
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


cstar/base/input_dataset.py
```diff
@@ -261,39 +261,60 @@ def get(self, local_dir: str | Path) -> None:
...
+        computed_file_hash = self._symlink_or_download_from_source(
+            source_location=self.source.location,
+            location_type=self.source.location_type,
+            expected_file_hash=self.source.file_hash,
+            target_path=target_path,
+            logger=self.log,
+        )
+
+        self.working_path = target_path
+        self._local_file_hash_cache.update({target_path: computed_file_hash})  # 27
```
ScottEilerman, line 273: `# 27` 🤔

cstar/roms/input_dataset.py
```diff
@@ -27,24 +70,49 @@ class ROMSInputDataset(InputDataset, ABC):
...
+            start_date=start_date,
+            end_date=end_date,
+        )
+
+        self.source_np_xi = source_np_xi
+        self.source_np_eta = source_np_eta
+        self.partitioning: Optional[ROMSPartitioning] = None
+
+    @property
+    def source_partitioning(self):
```
ScottEilerman:
```suggestion
    def source_partitioning(self) -> tuple[int,int] | None:
```

cstar/roms/input_dataset.py
```diff
@@ -53,7 +121,9 @@ def __repr__(self) -> str:
 
         return repr_str
 
-    def partition(self, np_xi: int, np_eta: int):
+    def partition(
+        self, np_xi: int, np_eta: int, overwrite_existing_files: bool = False
```
ScottEilerman, line 126: I was going to add a comment that, if `overwrite_existing_files is True`, we should maybe clear out the old files just to make sure we don't end up with a confusing state where some are overwritten but some others may be left over from the previous partitioning. But, now I'm wondering if we should just not support overwriting here at all, and make people manually clean up the old files if they're going to do so? Thoughts?

cstar/roms/input_dataset.py
```diff
@@ -158,10 +235,78 @@ def get(
...
+        else:
             super().get(local_dir=local_dir)
-            return
-        elif self.source.location_type == "path":
+
+    def _get_from_partitioned_source(
+        self, local_dir: Path, source_np_xi: int, source_np_eta: int
+    ) -> None:
+        n_source_partitions = source_np_xi * source_np_eta
+        ndigits = len(str(n_source_partitions))
```
ScottEilerman, line 302: This is genius. I would have done something over-complicated with division and `mod`.

cstar/base/input_dataset.py
```diff
@@ -261,39 +261,60 @@ def get(self, local_dir: str | Path) -> None:
...
+        self._local_file_stat_cache.update({target_path: target_path.stat()})
+
+    @staticmethod
+    def _symlink_or_download_from_source(
```
ScottEilerman, line 277: Unless there's some usage I'm not seeing, it seems like we can drop `@staticmethod`, since we only ever call this from within the class and are often passing it `self` variables for location_type, logger, etc.

ScottEilerman: After looking a little more, I see that in the two calls, some of the variables get `self.something` in one but a different value in the other, so it makes sense that they need to be passed in. It looks like the two I mentioned above are always the same `self` value though and could be dropped.

ScottEilerman: This could wait until a later refactor if we want. In the next blueprint iteration, we will probably split the symlink and download functionality into two subclasses anyway (along the `if location_type` branches you have in here already).

ScottEilerman (commented):
A few cleanup comments, I think the only big decision is whether we bother supporting the overwrite feature (and if so, whether we need to add some safeguards).

cstar/base/input_dataset.py
dafyddstephenson: IYKYK

cstar/roms/input_dataset.py
dafyddstephenson: I considered this, though it is just a tidying thing: if, e.g. we first partition into 25 tiles and then 15, only the first 15 will be used, so 16-25 will just be hanging around doing nothing, in an ugly but ultimately harmless way. I also considered moving the existing files to a temporary location and having a `finally` to return them in case of failure, which covers both cases. WDYT?

ScottEilerman: Yeah, I just imagine it will be easily forgotten and someone will look at the directly later and see 25 files, and not know that 16-25 are now totally incompatible with 1-15.

I guess the question to me is: how often is someone going to hit this "I want to intentionally change partitioning" condition? If it's often enough that we should handle it for them, we should handle it well. My assumption (which could be wrong) is that it will be rare enough that we can just throw an error for now and tell them to clean out the old files themselves.

dafyddstephenson: Agree, I just took the "well they shouldn't be peeking at the files, C-Star is taking care of it for them" line. But either way it's a bit gross and not completely insane to do the temporary move thing, so I'll just add that.

dafyddstephenson: [8ed114f](https://github.com/CWorthy-ocean/C-Star/pull/252/commits/8ed114f075c136225a9ff46911fe466f7dccad4f)

cstar/roms/input_dataset.py
```diff
@@ -136,73 +137,134 @@ def partition(
...
-                raise FileExistsError(
-                    f"The file has already been partitioned into a different arrangement "
-                    f"({self.partitioning.np_xi},{self.partitioning.np_eta}). "
-                    "To overwrite these files, try again with overwrite_existing_files=True"
+                id_files_to_partition = [
+                    self.working_path,
+                ]
+            return id_files_to_partition
+
+        def partition_files(files: list[Path]) -> list[Path]:
```
ScottEilerman, line 198: The files arg here isn't used. It still ends up working because `id_files_to_partition` ends up in scope when it's called. I usually prefer private functions at the same level over inner functions for this reason, because the scope can get confusing. Just a preference though, there's reasons to prefer keeping all of the logic encapsulated as well, if the private functions don't get used anywhere else.

cstar/roms/input_dataset.py
```diff
@@ -230,6 +292,8 @@ def get(
             working_path_parent = self.working_path[0].parent
         else:
             working_path_parent = self.working_path.parent
+        # import pdb
+        # pdb.set_trace()
```
ScottEilerman: Missed some cleanup here.

cstar/roms/input_dataset.py
dafyddstephenson: Good catch - fixed ([bfdbe45](https://github.com/CWorthy-ocean/C-Star/pull/252/commits/bfdbe45b4fc2bbed0cadc1f9de13747729754150)). The main reason for nesting these ones was just organisation/neatness (too much logic to sprawl it all out at once, too exclusive to the parent function to define at the class level). There is one example that's re-used, so is class-level: `_update_partitioning_attribute`.

cstar/roms/input_dataset.py
dafyddstephenson: No idea how this got in here, wasn't me. I always tidy up comments and never have to debug, perfect code flows directly from my mind into the keys. Took it out ([c036584](https://github.com/CWorthy-ocean/C-Star/pull/252/commits/c0365840387410378552824388e5f1e0b65c0d0c)) but we should definitely find who's responsible 🥷

ScottEilerman (approved):

PR #258, opened 2025-05-14
Fix bug where errors may not be displayed to user

ankona:
### Description

The `_run_cmd` utility method introduced a defect where only successful execution outputs are displayed directly to the user. 

See `cstar/base/utils.py`

```
        if raise_on_error:
            raise RuntimeError(msg)
```

This results in the stdout being written to a log but may not be displayed to the user if raise_on_error is False, resulting in a silent failure.

### Request

Update the above code to include a print to `stdout` including the error.

```
        if raise_on_error:
            raise RuntimeError(msg)

        # Exception was not thrown. Report error
        print(stderr)
```

### Acceptance Criteria

- The user is informed of a failure in _run_cmd without looking at logs
- All existing tests must pass

### Checklist

- [X] Closes #CW-821
- [X] Tests passing
- [X] Tests added

### DIFF
--- a/cstar/base/utils.py
+++ b/cstar/base/utils.py
@@ -264,6 +264,7 @@ def _run_cmd(
         if raise_on_error:
             raise RuntimeError(msg)
 
+        print(result.stderr)
         log.error(msg)
 
     log.debug(msg_post or "Command completed successfully.")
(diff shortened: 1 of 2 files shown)


cstar/base/utils.py
```diff
@@ -264,6 +264,7 @@ def _run_cmd(
         if raise_on_error:
             raise RuntimeError(msg)
 
+        print(result.stderr)
         log.error(msg)
```
ScottEilerman, line 268: Doesn't `log.error(msg)` already get this info to the user (in whatever way they have chosen to set up their logging, which should default to stdout)?

ankona: Sounds right. Closing to investigate original conditions that caused me to identify bug more closely.

PR #269, opened 2025-06-16
fix incorrect path in coverage file

ankona:
This PR fixes a minor defect in the `coverage.toml` file. It fixes the path to the tests directory so they are excluded from code coverage metrics.

- [x] Tests passing


dafyddstephenson:
integration tests hung so I'm gonna re-run but should be good

dafyddstephenson (approved):

PR #272, opened 2025-06-16
fix shadowed declaration of files member

ankona:
- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -42,8 +42,6 @@ class AdditionalCode(LoggingMixin):
        Verify whether the files associated with this AdditionalCode instance can be found at `local_dir`
     """
 
-    files: list[str]
-
     def __init__(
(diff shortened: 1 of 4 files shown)


dafyddstephenson (changes requested):
I think the original inclusion of the class-level typehint was to silence the linter everywhere you've added `ignore`s, but most of those lines are now gone, so hopefully after merging the latest `develop` into this branch we can just throw away the duplicated hint without needing any linter overrides

ankona:
> I think the original inclusion of the class-level typehint was to silence the linter everywhere you've added `ignore`s, but most of those lines are now gone, so hopefully after merging the latest `develop` into this branch we can just throw away the duplicated hint without needing any linter overrides

W00t!

PR #274, opened 2025-06-16
fix unnecessary conditional

ankona:
This PR contains a minor tweak to avoid an extraneous compound conditional.

- [x] Tests added


dafyddstephenson:
Should be fine, I'm just re-running the integration tests that timed out first

PR #281, opened 2025-06-16
fix unnecessary keys iteration

ankona:
This PR makes a minor tweak to avoid iterating over all keys in a dictionary when a specific key is not found (by getting the key directly instead of checking membership).

- [x] Tests passing

### DIFF
--- a/cstar/base/external_codebase.py
+++ b/cstar/base/external_codebase.py
@@ -173,12 +173,12 @@ class ExternalCodeBase(ABC, LoggingMixin):
         """
 
         # check 1: X_ROOT variable is in user's env
-        env_var_exists = cstar_sysmgr.environment.environment_variables.get(
+        env_var = cstar_sysmgr.environment.environment_variables.get(
             self.expected_env_var, None
         )
 
         # check 2: X_ROOT points to the correct repository
-        if env_var_exists:
+        if env_var:
             local_root = Path(
                 cstar_sysmgr.environment.environment_variables[self.expected_env_var]
             )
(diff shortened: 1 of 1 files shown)


cstar/base/external_codebase.py
```diff
@@ -175,9 +175,8 @@ def local_config_status(self) -> int:
         """
 
         # check 1: X_ROOT variable is in user's env
-        env_var_exists = (
-            self.expected_env_var
-            in cstar_sysmgr.environment.environment_variables.keys()
+        env_var_exists = cstar_sysmgr.environment.environment_variables.get(
+            self.expected_env_var, None
```
dafyddstephenson: Current change modifies type from bool
```suggestion
env_var_exists = self.expected_env_var in cstar_sysmgr.environment.environment_variables
```

dafyddstephenson (changes requested):

cstar/base/external_codebase.py
ankona: truthiness of the subsequent tests work on null/non-null without the implied iteration of `in`

dafyddstephenson: In that case I'd suggest renaming to `env_var` as `env_var_exists` suggests a bool

dafyddstephenson: in that case I'd suggest renaming to `env_var` as `env_var_exists` suggests a bool

dafyddstephenson (changes requested):

cstar/base/external_codebase.py
ankona: Done

dafyddstephenson (approved):

PR #283, opened 2025-06-16
fix param that fails typecheck

ankona:
This PR updates the signature of the `to_blueprint` method on `Simulation` and its subclasses to allow strings or paths. This modification enables the type checker to pass a test that previously failed.

- [X] Tests passing
- [X] Full type hint coverage

### DIFF
--- a/cstar/roms/simulation.py
+++ b/cstar/roms/simulation.py
@@ -793,7 +793,7 @@ class ROMSSimulation(Simulation):
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


dafyddstephenson (changes requested):
I think the actual fix here is to change the typehint on `Simulation.to_blueprint` and `ROMSSimulation.to_blueprint` to `Path | str`

dafyddstephenson (approved):

PR #284, opened 2025-06-16
Refactor lmod path calculation to one location

ankona:
This PR refactors the creation of the lmod path into a property on `CStarEnvironment` and updates two locations indepdentdently calculating the same path. It also adds some fixtures for mocking the lmod path.

- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage

### DIFF
--- a/cstar/tests/unit_tests/conftest.py
+++ b/cstar/tests/unit_tests/conftest.py
@@ -1,6 +1,9 @@
+import functools
 import logging
-import pathlib
+from collections.abc import Callable
+from pathlib import Path
 
+import dotenv
 import pytest
 
 from cstar.base.log import get_logger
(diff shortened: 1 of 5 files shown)


dafyddstephenson (changes requested):
Changes look good conceptually but unit tests are broken!

cstar/tests/unit_tests/conftest.py
```diff
@@ -43,11 +40,132 @@ def mock_system_name() -> str:
...
+    Returns
+    -------
+    str
+        The complete path to the file
+    """
+    tmp_path.mkdir(parents=True, exist_ok=True)
+
+    path = tmp_path / mock_lmod_filename
+    path.touch()  # CStarEnvironment expects the file to exist & opens it
+    return path
```
dafyddstephenson, line 171: Would suggest adding these to `cstar/tests/unit_tests/system/conftest.py` for cleaner organisation as (I think) they won't be needed at a higher level?

dafyddstephenson (dismissed):

cstar/tests/unit_tests/conftest.py
ankona: I make use of them in the worker unit tests, which is the follow-up to this PR. Leaving here should avoid dupes.

PR #285, opened 2025-06-17
Add class to track metadata of local files

dafyddstephenson:
> WIP! Opening as a draft to solicit feedback on the high-level structure

EDIT: ready to go!
NOTE: this is built on top of the commits in #288 , merge #288 first and then update diffs.

- [ ] Closes #xxxx
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage
- [x] New functions/methods are listed in `api.rst`

As part of consolidating the various `get()` and `get()`-like methods on classes like `AdditionalCode`, `InputDataset`, etc. into a single `DataSource.get()` method, this PR first consolidates related attributes _set_ by the `get()`-like methods on these classes that are used to track local file metadata, e.g. `AdditionalCode._local_file_hash_cache` or `InputDataset._local_file_stat_cache`.

This PR introduces a class `LocalFileStatistics` to track modification times, filesizes, SHA256 hashes, and names for a selection of (or single) co-located files. For now an instance of this class will be assigned to an attribute on the various "getting" classes, and they will keep their existing `get()` methods. In the future, `DataSource.get()` will return a `LocalFileStatistics` instance which can then optionally be assigned to an attribute on the class in question.

Example :

```
In [1]: from cstar.base import AdditionalCode

In [2]: compile_time_code = AdditionalCode(location = "~/Code/my_ucla_roms/Examples/Rivers_real/",
                                   files=["cppdefs.opt",
                                          "flux_frc.opt",
                                          "ocean_vars.opt",
                                          "param.opt",
                                          "river_frc.opt"])

In [3]: print(compile_time_code.local_file_stats)
None

In [4]: compile_time_code.get("tmp/")
[INFO] • Copying cppdefs.opt to /Users/dafyddstephenson/Code/my_c_star/cstar/base/tmp
[INFO] • Copying flux_frc.opt to /Users/dafyddstephenson/Code/my_c_star/cstar/base/tmp
[INFO] • Copying ocean_vars.opt to /Users/dafyddstephenson/Code/my_c_star/cstar/base/tmp
[INFO] • Copying param.opt to /Users/dafyddstephenson/Code/my_c_star/cstar/base/tmp
[INFO] • Copying river_frc.opt to /Users/dafyddstephenson/Code/my_c_star/cstar/base/tmp
[INFO] ✅ All files copied successfully

In [5]: print(compile_time_code.local_file_stats)
LocalFileStatistics
-------------------
Parent directory: /Users/dafyddstephenson/Code/my_c_star/cstar/base/tmp
Files: 

Name                                     Hash         Size (bytes) Modified
cppdefs.opt                              65de992b17…  1202         2025-06-17 16:56:25
flux_frc.opt                             2600e40001…  1068         2025-06-17 16:56:25
ocean_vars.opt                           90fc1cf1e1…  2346         2025-06-17 16:56:25
param.opt                                48cc34acaa…  908          2025-06-17 16:56:25
river_frc.opt                            1f23e50bf4…  728          2025-06-17 16:56:25
```

### DIFF
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -132,7 +132,7 @@ class AdditionalCode(LoggingMixin):
     @property
     def exists_locally(self):
         if not self.local_file_stats:
-            return None
+            return False
         try:
             self.local_file_stats.validate()
         except (FileNotFoundError, ValueError):
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -135,7 +135,7 @@ class AdditionalCode(LoggingMixin):
             return False
         try:
             self.local_file_stats.validate()
-        except (FileNotFoundError, ValueError):
+        except (FileNotFoundError, ValueError, KeyError):
             return False
         return True
 
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -76,7 +76,15 @@ class AdditionalCode(LoggingMixin):
         self.source: DataSource = DataSource(location)
         self.subdir: str = subdir
         self._checkout_target = checkout_target
-        self.files: Optional[list[str]] = [] if files is None else files
+
+        self.files: Optional[list[str]]
+        if files is None:
+            self.files = []
+        elif any((not f) for f in files):
+            raise ValueError("An invalid filename was supplied to AdditionalCode")
+        else:
+            self.files = files
+
         # Initialize object state
         self.local_file_stats: Optional[LocalFileStatistics] = None
 
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -121,18 +121,26 @@ class AdditionalCode(LoggingMixin):
 
     @property
     def checkout_target(self) -> Optional[str]:
+        """If in a repository, the commit hash or tag at which to checkout this
+        additional code."""
         return self._checkout_target
 
     @property
     def working_path(self) -> Optional[Path]:
+        """The current local path where this AdditionalCode exists on the local system,
+        if it has been fetched."""
+
         if self.local_file_stats is None:
             return None
         return self.local_file_stats.parent_dir
 
     @property
-    def exists_locally(self):
+    def exists_locally(self) -> bool:
+        """Determines whether this AdditionalCode instance exists on the local
+        system."""
         if not self.local_file_stats:
             return False
+
         try:
             self.local_file_stats.validate()
         except (FileNotFoundError, ValueError, KeyError):
@@ -212,7 +220,6 @@ class AdditionalCode(LoggingMixin):
                 paths=[local_dir / Path(f).name for f in self.files]
             )
             self.log.info("✅ All files copied successfully")
-            # self.working_path = local_dir
 
         finally:
             if tmp_dir:
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -1,12 +1,13 @@
 import shutil
 import tempfile
 from pathlib import Path
-from typing import Dict, Optional
+from typing import Optional
 
 from cstar.base.datasource import DataSource
 from cstar.base.gitutils import _clone_and_checkout
+from cstar.base.local_file_stats import LocalFileStatistics
 from cstar.base.log import LoggingMixin
-from cstar.base.utils import _get_sha256_hash, _list_to_concise_str
+from cstar.base.utils import _list_to_concise_str
 
 
 class AdditionalCode(LoggingMixin):
@@ -77,8 +78,8 @@ class AdditionalCode(LoggingMixin):
         self._checkout_target = checkout_target
         self.files: Optional[list[str]] = [] if files is None else files
         # Initialize object state
-        self.working_path: Optional[Path] = None
-        self._local_file_hash_cache: Dict = {}
+        # self.working_path: Optional[Path] = None
+        self.local_file_stats: Optional[LocalFileStatistics] = None
 
     def __str__(self) -> str:
         base_str = self.__class__.__name__ + "\n"
@@ -88,8 +89,8 @@ class AdditionalCode(LoggingMixin):
             base_str += f"\nSubdirectory: {self.subdir}"
         if self.checkout_target is not None:
             base_str += f"\nCheckout target: {self.checkout_target}"
-        base_str += f"\nWorking path: {self.working_path}"
-        base_str += f"\nExists locally: {self.exists_locally}"
+        # base_str += f"\nWorking path: {self.working_path}"
+        # base_str += f"\nExists locally: {self.exists_locally}"
         if not self.exists_locally:
             base_str += " (get with AdditionalCode.get())"
         if self.files is not None:
@@ -110,9 +111,11 @@ class AdditionalCode(LoggingMixin):
         repr_str += "\n)"
         # Additional info:
         info_str = ""
-        if self.working_path is not None:
-            info_str += f"working_path = {self.working_path},"
-            info_str += f"exists_locally = {self.exists_locally}"
+
+        # if self.working_path is not None:
+        #     info_str += f"working_path = {self.working_path},"
+        #     info_str += f"exists_locally = {self.exists_locally}"
+
         if len(info_str) > 0:
             repr_str += f"\nState: <{info_str}>"
         return repr_str
@@ -123,20 +126,12 @@ class AdditionalCode(LoggingMixin):
 
     @property
     def exists_locally(self):
-        """Determine whether a local working copy of the AdditionalCode exists at
-        self.working_path (bool)"""
-        if (self.working_path is None) or (self._local_file_hash_cache is None):
+        if not self.local_file_stats:
+            return None
+        try:
+            self.local_file_stats.validate()
+        except (FileNotFoundError, ValueError):
             return False
-
-        for f in self.files:
-            path = self.working_path / f
-            if not path.exists():
-                return False
-
-            current_hash = _get_sha256_hash(path.resolve())
-            if self._local_file_hash_cache.get(path) != current_hash:
-                return False
-
         return True
 
     def get(self, local_dir: str | Path) -> None:
@@ -205,15 +200,18 @@ class AdditionalCode(LoggingMixin):
                 )
                 if src_file_path.exists():
                     shutil.copy(src_file_path, tgt_file_path)
-                    self._local_file_hash_cache[tgt_file_path] = _get_sha256_hash(
-                        tgt_file_path
-                    )
-
+                    # self._local_file_hash_cache[tgt_file_path] = _get_sha256_hash(
+                    #     tgt_file_path
+                    # )
                 else:
                     raise FileNotFoundError(f"Error: {src_file_path} does not exist.")
 
+            self.local_file_stats = LocalFileStatistics(
+                paths=[local_dir / Path(f).name for f in self.files]
+            )
             self.log.info("✅ All files copied successfully")
-            self.working_path = local_dir
+            # self.working_path = local_dir
+
         finally:
             if tmp_dir:
                 shutil.rmtree(tmp_dir)
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -225,7 +225,7 @@ class AdditionalCode(LoggingMixin):
                     raise FileNotFoundError(f"Error: {src_file_path} does not exist.")
 
             self.local_file_stats = LocalFileStatistics(
-                paths=[local_dir / Path(f).name for f in self.files]
+                files=[local_dir / Path(f).name for f in self.files]
             )
             self.log.info("✅ All files copied successfully")
 
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -78,7 +78,6 @@ class AdditionalCode(LoggingMixin):
         self._checkout_target = checkout_target
         self.files: Optional[list[str]] = [] if files is None else files
         # Initialize object state
-        # self.working_path: Optional[Path] = None
         self.local_file_stats: Optional[LocalFileStatistics] = None
 
     def __str__(self) -> str:
@@ -89,8 +88,8 @@ class AdditionalCode(LoggingMixin):
             base_str += f"\nSubdirectory: {self.subdir}"
         if self.checkout_target is not None:
             base_str += f"\nCheckout target: {self.checkout_target}"
-        # base_str += f"\nWorking path: {self.working_path}"
-        # base_str += f"\nExists locally: {self.exists_locally}"
+        base_str += f"\nWorking path: {self.working_path}"
+        base_str += f"\nExists locally: {self.exists_locally}"
         if not self.exists_locally:
             base_str += " (get with AdditionalCode.get())"
         if self.files is not None:
@@ -112,9 +111,9 @@ class AdditionalCode(LoggingMixin):
         # Additional info:
         info_str = ""
 
-        # if self.working_path is not None:
-        #     info_str += f"working_path = {self.working_path},"
-        #     info_str += f"exists_locally = {self.exists_locally}"
+        if self.working_path is not None:
+            info_str += f"working_path = {self.working_path},"
+            info_str += f"exists_locally = {self.exists_locally}"
 
         if len(info_str) > 0:
             repr_str += f"\nState: <{info_str}>"
@@ -123,6 +122,12 @@ class AdditionalCode(LoggingMixin):
     @property
     def checkout_target(self) -> Optional[str]:
         return self._checkout_target
+
+    @property
+    def working_path(self) -> Optional[Path]:
+        if self.local_file_stats is None:
+            return None
+        return self.local_file_stats.parent_dir
 
     @property
     def exists_locally(self):
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -205,9 +205,6 @@ class AdditionalCode(LoggingMixin):
                 )
                 if src_file_path.exists():
                     shutil.copy(src_file_path, tgt_file_path)
-                    # self._local_file_hash_cache[tgt_file_path] = _get_sha256_hash(
-                    #     tgt_file_path
-                    # )
                 else:
                     raise FileNotFoundError(f"Error: {src_file_path} does not exist.")
 
--- a/cstar/base/input_dataset.py
+++ b/cstar/base/input_dataset.py
@@ -8,7 +8,7 @@ import dateutil.parser
 import pooch
 
 from cstar.base.datasource import DataSource
-from cstar.base.local_file_stats import LocalFileStatistics
+from cstar.base.local_file_stats import FileInfo, LocalFileStatistics
 from cstar.base.log import LoggingMixin
 from cstar.base.utils import _get_sha256_hash
 
@@ -196,11 +196,9 @@ class InputDataset(ABC, LoggingMixin):
 
         p = target_path.absolute()
         self.local_file_stats = LocalFileStatistics(
-            paths=[
-                p,
-            ],
-            stats={p: p.stat()},
-            hashes={p: computed_file_hash},
+            files=[
+                FileInfo(path=p, stat=p.stat(), sha256=computed_file_hash),
+            ]
         )
 
     @staticmethod
--- a/cstar/base/input_dataset.py
+++ b/cstar/base/input_dataset.py
@@ -191,7 +191,6 @@ class InputDataset(ABC, LoggingMixin):
             logger=self.log,
         )
 
-        # TODO
         p = target_path.absolute()
         self.local_file_stats = LocalFileStatistics(
             paths=[
@@ -211,7 +210,7 @@ class InputDataset(ABC, LoggingMixin):
     ) -> str:
         if location_type == "path":
             source_location = Path(source_location).expanduser().resolve()
-            # TODO this probably doesn't need to be calculated here
+            # TODO when refactoring get(), avoid calculating hash here
             computed_file_hash = _get_sha256_hash(source_location)
             if (expected_file_hash is not None) and (
                 expected_file_hash != computed_file_hash
--- a/cstar/base/input_dataset.py
+++ b/cstar/base/input_dataset.py
@@ -148,7 +148,7 @@ class InputDataset(ABC, LoggingMixin):
             return False
         try:
             self.local_file_stats.validate()
-        except (FileNotFoundError, ValueError):
+        except (FileNotFoundError, ValueError, KeyError):
             return False
         return True
 
@@ -291,12 +291,13 @@ class InputDataset(ABC, LoggingMixin):
         )
 
         # TODO
+        p = target_path.absolute()
         self.local_file_stats = LocalFileStatistics(
             paths=[
-                target_path,
+                p,
             ],
-            stats={target_path: target_path.stat()},
-            hashes={target_path: computed_file_hash},
+            stats={p: p.stat()},
+            hashes={p: computed_file_hash},
         )
         # self.working_path = target_path
         # self._local_file_hash_cache.update({target_path: computed_file_hash})  # 27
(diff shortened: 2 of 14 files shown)


ankona:
> > WIP! Opening as a draft to solicit feedback on the high-level structure
> > EDIT: ready to go!
> > NOTE: this is built on top of the commits in #288 , merge #288 first and then update diffs.

>Is this "draft only" comment still valid? Did you mean to be in draft mode and just clicked the wrong thing? 

No the only mistake here is the quote block should have ended before "EDIT". The PR was in draft mode until it wasn't

cstar/base/additional_code.py
```diff
@@ -121,22 +123,20 @@ def __repr__(self) -> str:
     def checkout_target(self) -> Optional[str]:
         return self._checkout_target
 
+    @property
+    def working_path(self) -> Optional[Path]:
```
ankona, line 137: docstring

cstar/base/additional_code.py
```diff
@@ -121,22 +123,20 @@ def __repr__(self) -> str:
...
+        if self.local_file_stats is None:
+            return None
+        return self.local_file_stats.parent_dir
+
     @property
     def exists_locally(self):
```
ankona: dostring

cstar/base/additional_code.py
```diff
@@ -205,15 +205,15 @@ def get(self, local_dir: str | Path) -> None:
...
-
                 else:
                     raise FileNotFoundError(f"Error: {src_file_path} does not exist.")
 
+            self.local_file_stats = LocalFileStatistics(
+                paths=[local_dir / Path(f).name for f in self.files]
+            )
             self.log.info("✅ All files copied successfully")
-            self.working_path = local_dir
+            # self.working_path = local_dir
```
ankona: don't forget to remove your commented code

cstar/base/additional_code.py
```diff
@@ -77,8 +78,7 @@ def __init__(
         self._checkout_target = checkout_target
         self.files: Optional[list[str]] = [] if files is None else files
```
ankona: I think we might have a bug in here. Consider writing a test that puts in an empty string in the files list and we'll probably end up with a bad path.

You may end up doing something like:

```py
if any(f is None for f in files):
  raise ValueError("An invalid file path was supplied.")
```

ankona: +1 for a test passing `files = [None, None]` producing an invalid state

cstar/base/additional_code.py
```diff
@@ -121,22 +123,20 @@ def __repr__(self) -> str:
...
+            return None
+        return self.local_file_stats.parent_dir
+
     @property
     def exists_locally(self):
-        """Determine whether a local working copy of the AdditionalCode exists at
-        self.working_path (bool)"""
-        if (self.working_path is None) or (self._local_file_hash_cache is None):
+        if not self.local_file_stats:
+            return False
```
ankona, line 150: total style nitpick... consider a line of whitespace here to differentiate the if and try

cstar/base/input_dataset.py
```diff
@@ -85,94 +84,22 @@ def validate(self):
...
-        elif isinstance(self.working_path, list):
-            local_hash = {
-                path: _get_sha256_hash(path.resolve()) for path in self.working_path
-            }
-        elif isinstance(self.working_path, Path):
-            local_hash = {self.working_path: _get_sha256_hash(self.working_path)}
-
-        self._local_file_hash_cache.update(local_hash)
-        return local_hash
+    def working_path(self) -> Optional[Path | list[Path]]:
```
ankona, line 98: Consider simplifying consumer code by avoiding nullables and varying types.

```python
@property
def working_path(self) -> list[Path]:
    if not self.local_file_stats:
        return []

    return self.local_file_stats.paths
```

In addition... either the method is poorly named (working_**path** implies it's always a _singular_ result) or the return type allowing a list makes no sense.

If you want `working_path` to always return the first path, fine. Just give yourself another property `all_paths -> list[Path]` to expose the other things instead of having this property do double duty.

ankona, line 98: docstring

cstar/base/input_dataset.py
```diff
@@ -74,9 +75,7 @@ def __init__(
         assert self.end_date is None or isinstance(self.end_date, dt.datetime)
 
         # Initialize object state:
-        self.working_path: Optional[Path | List[Path]] = None
-        self._local_file_hash_cache: Dict = {}
-        self._local_file_stat_cache: Dict = {}
+        self.local_file_stats: Optional[LocalFileStatistics] = None
```
ankona, line 78: Consider implementing `LocalFileStatistics` so you can avoid null value checks in client code. Using the same pattern as below, we can use it later without first checking for none...

Consider instantiating with:

```python
self.local_file_stats = LocalFileStatistics(
            paths=[],
            stats={},
            hashes={},
        )
```

You might need to tweak `LocalFileStatistics.is_valid` but you'll like using this better  if you can just do

```python
try:
  return self.local_file_stats.is_valid():
except ...:
   return False

...
```

cstar/base/input_dataset.py
```diff
@@ -283,6 +210,7 @@ def _symlink_or_download_from_source(
     ) -> str:
         if location_type == "path":
             source_location = Path(source_location).expanduser().resolve()
+            # TODO when refactoring get(), avoid calculating hash here
```
ankona, line 216: don't forget your todo. consider creating a ticket if it's not getting done in this PR

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+        """Initialize the LocalFileStatistics object.
+
+        Stores a list of local file paths and optionally precomputed metadata
+        such as file stat results and SHA-256 hashes. All paths must be located
+        within the same parent directory. Metadata is lazily computed if not
+        provided.
+
+        Parameters
+        ----------
+        paths : list of Path
```
ankona: I just created a new ticket about these docstrings... The resulting documentation looks really bad and these parameters and properties are already covered in the documentation.

We should stop copy/pasting every docstring into the class level docstring. 

See [this](https://cworthy.atlassian.net/browse/CW-875) for an example of how poorly docs render when included in the class docstring.

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+        Raises
+        ------
+        ValueError
+            If the provided paths are not all in the same directory or if the keys
+            of `stats` or `hashes` do not match the resolved paths.
+        """
+
+        self.paths = [p.absolute() for p in paths]
+
+        if not all([d.parent == self.paths[0].parent for d in self.paths]):
```
ankona:
```suggestion
        if not all(d.parent == self.paths[0].parent for d in self.paths):
```
unnecessary comprehension.

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+    >>> tracker.stats[files[0]]
+    os.stat_result(...)
+    >>> tracker.hashes[files[0]]
+    'e3b0c44298fc1c149afbf4c8996fb924...'
+    """
+
+    def __init__(
+        self,
+        paths: list[Path],
+        stats: Optional[dict[Path, os.stat_result]] = None,
```
ankona: Consider creating purpose-built methods for instantiating from pre-existing data. The constructor arguments are reduced and it makes it easier to understand.

```python
@classmethod
def hydrate(self, files: ..., stats: ..., hashes: ...) -> Self:
    if not files:
        raise ValueError("no files provided.")
    if not stats:
        raise ValueError("no stats provided.")
    if not hashes:
        raise ValueError("no hashes provided.")
    
    return LocalFileStats(files, stats, hashes)
```

Now, we have a method that doesn't allow nulls and it's really easy to separate when we should check for empty sets.

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+
+            current_hash = _get_sha256_hash(f)
+            if current_hash != self.hashes[f]:
+                raise ValueError(
+                    f"Computed SHA256 hash for {f} ({current_hash}) does not "
+                    f"match value in cache ({self.hashes[f]})"
+                )
+
+    def __str__(self) -> str:
+        base_str = self.__class__.__name__ + "\n"
```
ankona, line 174: these repeated string concatenations are poor form. Consider how much easier it is to see the intended formatting when avoiding doing your layout manually with `\n` and `+=`:

```python
return textwrap.dedent("""\
    {self.__class__.__name__}
    {"-" * (len(base_str) - 1)}
    Parent directory:
    {self.parent_dir}
    Files:
    
    {self.as_table()}""")
```

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+        base_str += "-" * (len(base_str) - 1) + "\n"
+        base_str += "Parent directory: "
+        base_str += f"{self.parent_dir}\n"
+        base_str += "Files: \n\n"
+        base_str += self.as_table()
+        return base_str
+
+    def __repr__(self) -> str:
+        repr_str = self.__class__.__name__ + "("
```
ankona, line 183: ditto on avoiding repetitive concatenations

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+        self.paths = [p.absolute() for p in paths]
+
+        if not all([d.parent == self.paths[0].parent for d in self.paths]):
+            raise ValueError(
+                "LocalFileStatistics requires paths to exist in a common directory"
+            )
+
+        self.parent_dir = paths[0].parent
+
+        if stats is not None:
```
ankona: Same deal with nulls. Look for a way to process additions immediately to avoid having a null stats and hashes member (same as comment about instantiating the LocalFileStats to an empty version).

If we're exposing a list, like "self.files = []` and allowing people to modify it at-will, we put ourselves in this position where we can't trust our member variables. If we hide that as `self._files = []` we can expose 

```python
def add_files(self, flist: list[Path]) -> None:
   if not validate(f):
      raise ValueError
   self._files.extend(f.absolute() for f in flist)
```

When I notice that I'm making myself do repeated checks for nulls, i always start to ask "how can I avoid doing this everywhere?"

cstar/base/local_file_stats.py
```diff
@@ -0,0 +1,229 @@
...
+        return base_str
+
+    def __repr__(self) -> str:
+        repr_str = self.__class__.__name__ + "("
+        repr_str += "paths="
+        repr_str += f"{(',\n'+' '*len(repr_str)).join([str(p) for p in self.paths])}"
+        repr_str += ")"
+        return repr_str
+
+    def as_table(self, max_rows: int = 20) -> str:
```
ankona, line 189: As table is a "presentation class" and should not be included on this class, outright. Consider creating another class just to contain this in a more "UI-focused" package:

```
class Tabularizer:
  """A contract to spit out data that can be rendered in a table"""

  """A contract for entities that can be rendered in a tabular format"""
  def as_table(self) -> list[list[str]]:
    # this method returns lists of lists on purpose so i can separate how
    # i render and what i render
    ...
  def headers(self) -> list[str]:
    ...

class LocalFileStatisticsBriefTabularizer:
   """Tabularizer allows me to implement a specific data extraction""""
    def __init__(self, lfs: LocalFileStatistics) -> None:
        self._lfs = lfs

    def as_table(self) -> list[list[str]]:
        return [["a", "b"] for x in self._lfs]
    def headers(self) -> list[str]:
      return ["A HEAD","B HEAD"]

class LocalFileStatisticsFullTabularizer:
   """This Tabularizer allows me to avoid logic for another desired rendering...""""
    def __init__(self, lfs: LocalFileStatistics) -> None:
        self._lfs = lfs

    def as_table(self) -> list[list[str]]:
        return [["a", "b", "c", "d"] for x in self._lfs]  
    def headers(self) -> list[str]:
      return ["A HEAD","B HEAD","C HEAD","D HEAD"]    

class TableRenderer:
    """THe magic! Now, I can render ANY Tabularizer with one method..."""
    @classmethod
    def _csv(self, tt: Tabularizer) -> str:
        data = tt.as_table(). # i don't care what's in there anymore...
        shead = ','.join(tt.headers)
        sdata = ','.join(x for x in data)
        return f"{shead}\n{sdata}"

    @classmethod
    def _json(self, tt: Tabularizer) -> str:
        data = tt.as_table(). # i don't care what's in there anymore...
        header = tt.headers()
        d = dict(zip(header, x) for x in data)
        return json.dumps(d)

    @classmethod
    def render(self, format: Literal['json','csv'], tt: Tabularizer) -> str:
        fmt_map = {"csv": self._csv, "json": self._json}
        return fmt_map[format](tt)


.... 

# somewhere else, when i want to use it...
print(TableRenderer.render(LocalFileStatisticsFullTabularizer(lfs)))
        
```

We now have a really nice separation of concerns for building tabular data (even in multiple layouts) and actually rendering it. It's more testable, and we can avoid repeating table rendering code everywhere (improving output quality because it's more consistent).

cstar/roms/input_dataset.py
```diff
@@ -268,21 +268,15 @@ def get(
...
-        elif isinstance(self.working_path, list):
-            working_path_parent = self.working_path[0].parent
-        else:
-            working_path_parent = self.working_path.parent
-
-        if (self.exists_locally) and (working_path_parent == local_dir):
-            self.log.info(f"⏭️ {self.working_path} already exists, skipping.")
-            return
+        if self.exists_locally:
+            assert self.local_file_stats is not None
```
ankona: assertion here is not documented in exceptions and is probably a bad choice. (also, consider rebasing off latest develop and using either `CStarError` or a new type).

You should raise a more descriptive error here than an AssertionError to inform the client code what went wrong

cstar/roms/input_dataset.py
```diff
@@ -268,21 +268,15 @@ def get(
...
-            working_path_parent = None
-        elif isinstance(self.working_path, list):
-            working_path_parent = self.working_path[0].parent
-        else:
-            working_path_parent = self.working_path.parent
-
-        if (self.exists_locally) and (working_path_parent == local_dir):
-            self.log.info(f"⏭️ {self.working_path} already exists, skipping.")
-            return
+        if self.exists_locally:
```
ankona, line 275: Consider inverting to avoid having nested ifs..

```python
if not self.exists_locally:
    self.warning(...)
    raise ValueError

if self.local_stats.parent_dir == local_dir:
    self.log.info(...)
    return
   ...

...
```

cstar/roms/input_dataset.py
```diff
@@ -321,10 +315,11 @@ def _get_from_partitioned_source(
         self._update_partitioning_attribute(
             new_np_xi=source_np_xi, new_np_eta=source_np_eta, parted_files=parted_files
         )
-        self.working_path = parted_files
+        # self.working_path = parted_files
```
ankona: don't forget your commented code

cstar/roms/input_dataset.py
```diff
@@ -321,10 +315,11 @@ def _get_from_partitioned_source(
...
         assert self.partitioning is not None
-        self._local_file_stat_cache.update(self.partitioning._local_file_stat_cache)
-        self._local_file_hash_cache.update(self.partitioning._local_file_hash_cache)
+        self.local_file_stats = LocalFileStatistics(
+            paths=[p.resolve() for p in parted_files]
```
ankona: you should not be required to know that you have to `p.resolve` before passing data to the LFS object. this implies that you understand the internal workings of that object and, in fact, must know that in every place it's created (see all the .resolve calls below). 

Make LFS call resolve on what it receives if it wants that?

cstar/roms/simulation.py
```diff
@@ -1592,9 +1592,8 @@ def restart(self, new_end_date: str | datetime) -> "ROMSSimulation":
         new_sim.initial_conditions = new_ic
 
         # Reset cached data for input datasets
+
         for inp in new_sim.input_datasets:
-            inp._local_file_hash_cache = None
-            inp._local_file_stat_cache = None
-            inp.working_path = None
+            inp.local_file_stats = None
```
ankona: Consider adding a `clear` method and maybe a `empty` property so we can avoid null checks.

`inp.clear()`

Same thought process here... you are making a `Simulation` responsible for what it means to be "empty" inside the input dataset, and that is not sustainable. Let the input_dataset own what clearing means.

cstar/base/additional_code.py
dafyddstephenson: 43f911d

cstar/base/additional_code.py
dafyddstephenson: 43f911d

cstar/base/additional_code.py
dafyddstephenson: 43f911d

cstar/base/additional_code.py
dafyddstephenson: 43f911d

cstar/base/input_dataset.py
dafyddstephenson: #291

cstar/base/local_file_stats.py
dafyddstephenson: API docs have been in need of a tidy-up for a while, so that would be great. I made a half-hearted effort once but didn't get far enough quick enough and bailed. If you could also add something to the contributing guide on how to write docstrings that get rendered nicely in the RTD that'd be great.

cstar/base/input_dataset.py
dafyddstephenson: I considered designing it as a `Collection` which would implicitly cover this, but in the end the PR was getting a bit big and taking a lot of time considering it's just a stepping stone to something else. I'll file a ticket once it's merged and revisit it once the subsequent PR closing #291 is in.

cstar/base/additional_code.py
dafyddstephenson: 017a2c8

cstar/base/input_dataset.py
dafyddstephenson: Opened #292 to address this. Bit of lore as to where this inconsistency came from in the first place in there.

dafyddstephenson: a318d90

cstar/base/local_file_stats.py
dafyddstephenson: This choice was made to allow conditionals in the `__str__` construction. There aren't any here, but in many other cases, lines are added or not depending on whether certain attributes are set, making a cleaner final `__str__` representation.

cstar/base/local_file_stats.py
dafyddstephenson: I think in this case it is simpler to maintain a single entry point with optionals for pre-computed values, rather than expanding the API with another constructor for a relatively uncommon case. Note also that in the current implementation, the `Optional`s aren't left as `None` in the final state.

cstar/base/local_file_stats.py
dafyddstephenson: See above comment

cstar/base/local_file_stats.py
dafyddstephenson: There are other instances like this in the code and a dedicated issue/PR pair to address it would be worthwhile, but I'm going to leave that as separate to this PR.

cstar/roms/input_dataset.py
dafyddstephenson: 68888fb

cstar/roms/input_dataset.py
dafyddstephenson: 97769b4

cstar/roms/simulation.py
dafyddstephenson: c6c099f

cstar/roms/input_dataset.py
dafyddstephenson: I don't think a guard clause makes sense here as `not self.exists_locally` doesn't require any special handling (like `raise`) - it's the expected situation, the correct result is to proceed without doing anything. I previously had the conditional as a one-liner (no nesting):

```
        if (self.exists_locally) and (self.local_file_stats.parent_dir == local_dir):  # type: ignore[union-attr]
            self.log.info(f"⏭️ {self.working_path} already exists, skipping.")
            return
```
... but find the current form more human readable, despite the single-level nest.

cstar/roms/input_dataset.py
dafyddstephenson: [901abc0](https://github.com/CWorthy-ocean/C-Star/pull/285/commits/901abc02b0a80200838fa2db908540c383d6c9af) - This was made redundant a number of commits ago and I just forgot to tidy it up

cstar/base/input_dataset.py
```diff
@@ -269,9 +194,14 @@ def get(self, local_dir: str | Path) -> None:
...
-        self._local_file_hash_cache.update({target_path: computed_file_hash})  # 27
-        self._local_file_stat_cache.update({target_path: target_path.stat()})
+        p = target_path.absolute()
+        self.local_file_stats = LocalFileStatistics(
+            paths=[
+                p,
+            ],
+            stats={p: p.stat()},
+            hashes={p: computed_file_hash},
+        )
```
ScottEilerman, line 202: I'm not sure if this is the same discussion as above that mentioned a Collection, but: seeing the usage here makes me wonder if there should be a `LocalFileStatistics` object for every path, and here we would hold a list of `LocalFileStatistics` objects. Then there's no need to have dictionaries for stats/hashes that you have to match with the list of paths, it's just 1:1:1 for every object.

I'll keep reading and see if I stick with this suggestion or if other usage will make me reconsider...

cstar/base/local_file_stats.py
ScottEilerman: For what it's worth, something like `f"{'foobar\n' if True else ''}"` is totally valid, though if you use a full dedent block like above, I guess you'd still end up with an empty line unless you pack optional stuff on the line above it which would get messy.

I'm also wondering if a lot of these str/reprs will be largely unused or unseen once we move away from interactive notebooks and more toward automated processes, so maybe not worth sinking a lot of time into until we see whether we really want them later.

cstar/base/input_dataset.py
ScottEilerman: After reading further, I do think this would be a good refactor, as it would simplify the `LocalFileStatistics` logic a lot by getting rid of some loops and potential null situations. Though you'd still need a collection construct to relocate those loops. Optional for now, it'll probably work fine as it is.

cstar/base/additional_code.py
```diff
@@ -110,33 +118,41 @@ def __repr__(self) -> str:
...
+        return self.local_file_stats.parent_dir
 
-            current_hash = _get_sha256_hash(path.resolve())
-            if self._local_file_hash_cache.get(path) != current_hash:
-                return False
+    @property
+    def exists_locally(self) -> bool:
+        """Determines whether this AdditionalCode instance exists on the local
+        system."""
+        if not self.local_file_stats:
```
ScottEilerman, line 149: With regards to @ankona's comments about nulls throughout, I wonder how much use we have for the program state where we haven't done a `.get()` yet but don't want to just go ahead and do it. If we were to lazily `get()` anything whenever we end up needing it, it might reduce a lot of this "have I gotten it locally yet" logic because the answer would always be yes.

Food for thought for now. Maybe it will become apparent in the next `get()` PR if we should go down this road or not.

ScottEilerman (approved):
A few remarks but no blockers on my end.

dafyddstephenson:
@ScottEilerman I've refactored the class a little so it doesn't rely so heavily on cross-checking a bunch of dictionaries anymore, there's now a dataclass `FileInfo` that holds a single one:one:one path/stat/hash combination. I think having `LocalFileStatistics` manage multiple related files is best, but this lightweight dataclass kinda does the same thing as what you were suggesting with a one-instance-per-file arrangement. `LocalFileStatistics` now also has a `__getitem__` so you can do lookup via `my_local_file_stats[my_path]` and get back a `FileInfo` instance. I think it's a lot neater - thanks for the suggestion.

cstar/base/input_dataset.py
dafyddstephenson: See below comment and 3a9846d6f578991a9c9df9ce321c446ce6687f3c

dafyddstephenson:
@ScottEilerman merged your branch and tidied up - thanks again!

ScottEilerman (approved):

PR #286, opened 2025-06-18
Improved CStarEnvironment loading behavior

ankona:
This PR includes the following features:
1. Expose the `system_env_path` property consistent w/`lmod_path` and `user_env_path`
2. Avoids unnecessary reloading of environment files and package root

- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage

### DIFF
--- a/cstar/system/environment.py
+++ b/cstar/system/environment.py
@@ -157,6 +157,9 @@ class CStarEnvironment:
     def _find_package_root(cls) -> Path:
         """Determine the root directory containing the source for this package.
 
+        Uses `importlib.util.find_spec` to locate the package directory, enabling
+        access to additional configuration files within the package structure.
+
         Returns
         -------
         Path
--- a/cstar/system/environment.py
+++ b/cstar/system/environment.py
@@ -3,7 +3,7 @@ import os
 import platform
 from pathlib import Path
 
-from dotenv import dotenv_values, load_dotenv, set_key
+from dotenv import dotenv_values, set_key
 
 from cstar.base.utils import _run_cmd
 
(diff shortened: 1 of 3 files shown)


cstar/system/environment.py
```diff
@@ -59,11 +59,23 @@ def __init__(
...
+            The name of the hosting platform
+        mpi_exec_prefix : str
+            The MPI prefix
+        compiler : str
+            The compiler to use for builds
+        """
         self._system_name = system_name
         self._mpi_exec_prefix = mpi_exec_prefix
         self._compiler = compiler
+        self._package_root: Path = self._find_package_root()
```
ankona:
```suggestion
        self._PACKAGE_ROOT: Path = self._find_package_root()
```

Consider renaming to be consistent w/other constant.

cstar/system/environment.py
```diff
@@ -113,15 +125,55 @@ def __repr__(self):
...
+        top_level_package_name = __name__.split(".")[0]
+        spec = importlib.util.find_spec(top_level_package_name)
+        if spec is not None and isinstance(spec.submodule_search_locations, list):
+            return Path(spec.submodule_search_locations[0])
+
+        raise ImportError(f"Top-level package '{top_level_package_name}' not found.")
+
     @property
     def package_root(self) -> Path:
         """Identifies the root directory of the top-level package.
```
dafyddstephenson: suggest moving the _next_ two lines to `_find_package root`, as they are describing logic that is now located there. Also suggest `Identifies` -> `Returns`

dafyddstephenson (approved):
Approve so we don't iterate but I did have one picky comment re. a docstring

ScottEilerman (approved):
no objections here!

cstar/system/environment.py
ankona: Renamed

cstar/system/environment.py
ankona: Swapped to return

ankona: Fixed docstring copypasta fail by moving detail description to `_find_package_root`

PR #287, opened 2025-06-18
Expose default_externals_root as property on ExternalCodeBase

ankona:
This PR extracts the calculation of the directory where external codebases will be written. It is now exposed as a property on the `ExternalCodeBase` class. 

This property enables the path to be re-used internally and externally without hardcoding copies of the magic string `"externals"`.

- [X] Tests passing
- [X] Full type hint coverage

### DIFF
--- a/cstar/base/external_codebase.py
+++ b/cstar/base/external_codebase.py
@@ -153,8 +153,8 @@ class ExternalCodeBase(ABC, LoggingMixin):
         MARBL_ROOT."""
 
     @property
-    def externals_root(self) -> Path:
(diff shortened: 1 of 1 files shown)


dafyddstephenson (changes requested):
Cool, makes total sense - only suggestion is to rename the property to something like `default_externals_root` or sth similar, as users do have the option of overriding it.

ankona:
> Cool, makes total sense - only suggestion is to rename the property to something like `default_externals_root` or sth similar, as users do have the option of overriding it.

Done

dafyddstephenson (approved):

PR #289, opened 2025-06-21
Manager refactor

ankona:
This PR refactors system-specific logic out of the `CStarSystemManager` by implementing configuration classes for the known platforms. By extracting config into individual dataclasses, it reduces the code churn necessary for the platform-specific additions or modifications. 

The `HostNameEvaluator` class was added to encapsulate logic for identifying the current platform and further focus the responsibilities of `CStarSystemManager`.

These changes are laying groundwork for the upcoming worker PR.

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/cstar/system/manager.py
+++ b/cstar/system/manager.py
@@ -71,8 +71,8 @@ class HostNameEvaluator:
     def name(self) -> str:
         """Determine the system name that will be used by C-Star.
 
-        Makes use of environment variables and platform details. Prioritizes an LMOD-
-        based system name first, then falls back to
+        Prioritizes the LMOD-based system name. When LMOD-specific environment
+        variables are not present, the name is determiend by the OS.
 
         Raises
         ------
--- a/cstar/system/manager.py
+++ b/cstar/system/manager.py
@@ -128,13 +128,8 @@ def register_sys_context(
     return _inner()
 
 
-def _get_system_context(name: str) -> _SystemContext:
+def _get_system_context() -> _SystemContext:
     """Retrieve a system context from the context registry.
-
-    Parameters
-    ----------
-    name : str
-        The name of the system to retrieve a context for
 
     Returns
     -------
@@ -146,10 +141,12 @@ def _get_system_context(name: str) -> _SystemContext:
     CStarError
         If the supplied name has not been registered.
     """
-    if type_ := _registry.get(name):
+    namer = HostNameEvaluator()
+
+    if type_ := _registry.get(namer.name):
         return type_()
 
-    raise CstarError(f"Unknown system requested: {name}")
+    raise CstarError(f"Unknown system requested: {namer.name}")
 
 
 @register_sys_context
@@ -281,8 +278,8 @@ class CStarSystemManager:
         Initialize the system manager by determining the system name and initializing
         the environment and scheduler based on that name.
         """
-        namer = HostNameEvaluator()
-        self._context = _get_system_context(namer.name)
+
+        self._context = _get_system_context()
         """A context object configured for the current system."""
         self._environment = CStarEnvironment(
             system_name=self._context.name,
--- a/cstar/system/manager.py
+++ b/cstar/system/manager.py
@@ -1,8 +1,10 @@
+import functools
 import os
-import platform
-from enum import Enum
-from typing import Optional
+import platform as platform_
+from dataclasses import dataclass, field
+from typing import ClassVar, Protocol
 
+from cstar.base.exceptions import CstarError
 from cstar.system.environment import CStarEnvironment
 from cstar.system.scheduler import (
     PBSQueue,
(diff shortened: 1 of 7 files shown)


cstar/system/manager.py
```diff
@@ -14,166 +16,337 @@
...
-        Otherwise, constructs a system name using `platform.system()` and
-        `platform.machine()`.
+        """Determine the system name that will be used by C-Star.
 
-        Returns
-        -------
-        str
-            The system's name in lowercase.
+        Makes use of environment variables and platform details. Prioritizes an LMOD-
+        based system name first, then falls back to
```
ScottEilerman: missing the end of this sentence

cstar/system/manager.py
```diff
@@ -14,166 +16,337 @@
...
+        )
+
+
+@register_sys_context
+@dataclass(frozen=True)
+class _DerechoSystemContext(_SystemContext):
+    """The contextual dependencies for the Derecho system."""
+
+    name: ClassVar[str] = "derecho"
+    """The unique name identifying the Derecho system."""
```
ScottEilerman, line 189: for my own edification: if we didn't specify these attribute docstrings for each class, would they inherit the docstring from the parent? It feels redundant but maybe it's necessary.

ScottEilerman (approved):
LGTM overall. We should keep in mind that a lot of the queue management will probably be moving to the server/orchestrator, whereas the runner should be seeing a fairly consistent container-internal environment. That said, this looks like a good refactor to get us from now to then, and probably a lot of these pieces will still be relevant even after we make that shift.

We also don't plan to use Derecho or Expanse anytime soon. The effort to maintain support here seems minimal, so by all means let's do so if it's easy. But if it becomes a bigger lift down the road, we can consider dropping them in the short term and revisiting once things are more solidified.

cstar/tests/unit_tests/system/test_hostnameevaluator.py
```diff
@@ -0,0 +1,204 @@
...
+    lmod_syshost = "sys-host"
+    lmod_sysname = "sys-name"
+    with patch.dict(
+        os.environ,
+        {
+            HostNameEvaluator.ENV_LMOD_SYSHOST: lmod_syshost,
+            HostNameEvaluator.ENV_LMOD_SYSNAME: lmod_sysname,
+        },
+    ) as _:
+        yield _
```
ScottEilerman, line 26: edification Q: is adding the `as _` / `yield _` functionally different from 

```python
with patch.dict(...):
    yield
```

or just a style choice?

dafyddstephenson:
> We also don't plan to use Derecho or Expanse anytime soon. The effort to maintain support here seems minimal, so by all means let's do so if it's easy. But if it becomes a bigger lift down the road, we can consider dropping them in the short term and revisiting once things are more solidified.

The philosophy for supporting Derecho and Expanse is mostly to keep us on our toes - the idea is every HPC system has quirks and if we only write for Perlmutter we will unknowingly bake in some system-specific stuff that should have been more general. These two HPCs are just ones a lot of us at CW have access to, but the more the merrier IMHO!

cstar/system/manager.py
ankona: NO, I'm

ankona: Fixed

cstar/system/manager.py
ankona: The documentation generates correctly without the duped attribute docstrings, but intellisense in vs code and nvim.cmp doesn't work correctly. I'd like to leave them for the developer experience.

cstar/tests/unit_tests/system/test_hostnameevaluator.py
ankona: The _ tells the linter it's ok to ignore it and yielding `_` ensures the typechecker is also satisfied. 

Omitting the `mock._patch` from the typehint would also make it unnecessary until the linter decides "if you're not yielding anything, return instead of yield." In the end, I really like to have the typehints if that fixture is used for `env_full_lmod.called` or whatever, so `yield _` wins

cstar/system/manager.py
ankona: I met you halfway. The property docstrings could be removed and doc-gen/autocomplete all work, so those are gone!

cstar/tests/unit_tests/system/test_hostnameevaluator.py
ScottEilerman: Reasonable! Thanks for explaining.

dafyddstephenson:
Sorry for the silence on this one, I have been checking everything runs on different HPC systems as part of reviewing it. Conceptually looks good though

cstar/tests/unit_tests/system/test_manager.py
```diff
@@ -116,146 +48,84 @@ class TestEnvironmentProperty:
...
-            system = CStarSystemManager()
-            scheduler = system.scheduler
-
-            # Verify that the scheduler is of the expected type
-            assert isinstance(scheduler, expected_scheduler_type)
-
-            # Verify that the queue names match
-            assert set(q.name for q in scheduler.queues) == set(expected_queue_names)
+        mock_environment_vars: dict[str, str],  # noqa: ARG002
+    ) -> None:
```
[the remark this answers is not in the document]
ankona: Done

cstar/tests/unit_tests/system/test_hostnameevaluator.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+@patch("platform.machine", return_value=DEFAULT_MOCK_MACHINE_NAME)
+@patch("platform.system", return_value=DEFAULT_MOCK_HOST_NAME)
+@patch.dict(os.environ, {})
+def test_no_lmod_in_env(
+    mock_system: mock.MagicMock, mock_machine: mock.MagicMock
+) -> None:
+    """Verify that the an environment that specifies no lmod environment variables
```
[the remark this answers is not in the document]
ankona: reworded for clarity and to fix above.

cstar/tests/unit_tests/system/test_hostnameevaluator.py
```diff
@@ -0,0 +1,204 @@
...
+        {},
+    ) as _:
+        yield _
+
+
+@patch("platform.machine", return_value=DEFAULT_MOCK_MACHINE_NAME)
+@patch("platform.system", return_value=DEFAULT_MOCK_HOST_NAME)
+@patch.dict(os.environ, {})
+def test_no_lmod_in_env(
+    mock_system: mock.MagicMock, mock_machine: mock.MagicMock
```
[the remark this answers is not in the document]
ankona: @patch injections are passed in reverse order to the patches, this is ok.

dafyddstephenson (changes requested):
Looks great on the whole, a couple suggestions that I'll just make here instead of line by line:
- I think it's generally clearer (and PEP8-suggested) to avoid trailing underscores for potential clashes (here `platform_` and `cls_`) by instead opting for more explicit variable names (e.g. `target_class` or similar instead of `cls_` in `register_sys_context`)
- This may actually be better off going into `main` and then merging back into `develop`, as it makes a lot of improvements to the existing feature set.

ankona:
> Looks great on the whole, a couple suggestions that I'll just make here instead of line by line:
> 
> * I think it's generally clearer (and PEP8-suggested) to avoid trailing underscores for potential clashes (here `platform_` and `cls_`) by instead opting for more explicit variable names (e.g. `target_class` or similar instead of `cls_` in `register_sys_context`)
> * This may actually be better off going into `main` and then merging back into `develop`, as it makes a lot of improvements to the existing feature set.

- Agreed on `xxx_` naming. Done!
- Renamed member vars to avoid the name collision with `platform` and get rid of `platform_`.
- Swapped PR base to `main`

dafyddstephenson (approved):

PR #298, opened 2025-07-16
Move source file retrieval to `datasource.py`

### LINKED ISSUE #291: Centralise get-like calls
We should move the logic of `AdditionalCode.get()` , `InputDataset.get()` and _possibly_ `ExternalCodeBase.get()` to `DataSource.get()`, i.e. the calls become `AdditionalCode.source.get()` (possibly with a wrapper method that allows continued use of `AdditionalCode.get()`). After #285, this method could then return a `LocalFileStatistics` instance which could be handled by the class in question when it comes to setting any related attributes.

Note to self - #285 leaves a TODO related to this in `base/input_dataset.py`

dafyddstephenson:
Closes #291 

⚠️ WORK IN PROGRESS - feedback requested on proposal (below), not code
⚠️ BUILT ON #285; MERGE #285 FIRST AND RELOAD DIFFS

---
Putting this out there for feedback before I start building. There is no code to review yet, the first commit was just to get something on this branch to issue a draft PR, but the new Enums from that commit will be refactored as part of this.

I've been thinking how to refactor `DataSource` so as to handle fetching and also maintain the current structure of the various primary classes (`AdditionalCode` , `InputDataset`, etc.).

Currently, `InputDataset` is fairly straight forward as typically there is a single `source` file (that may correspond to many local files, e.g. in the case of a roms-tools `yaml` recipe).

`AdditionalCode` is more complex - its `source` is usually a repository or local directory, but the `AdditionalCode` class also takes parameters `subdir` and `files` - if we look at the [example case](https://github.com/CWorthy-ocean/cstar_blueprint_roms_marbl_example), there are two associated `AdditionalCode` instances - one for compile-time code (`subdir = "roms_compile_time_code", files = ["bgc.opt", ... , "tracers.opt"]`) and one for run-time code (`subdir = "roms_runtime_code", files = ["roms.in", "marbl_in", "marbl_tracer_output_list", "marbl_diagnostic_output_list"]`). 

The idea here was that, while ocean model input datasets are typically a melange of stuff that scientists find down the back of each others sofa's, where each file can be very large, and not all of them are strictly necessary (say if you need a shorter run, you can safely drop later forcing files), the additional code related to a model configuration is almost always negligible in size (all plaintext), interdependent (you can't drop any of it), and colocated (usually all in one directory, to be compiled into, or read by, an executable).

The current `DataSource` implementation tracks a single location (e.g. a local netCDF file or remote repository). As `AdditionalCode` expects a subset of files from a single directory, it also has the `subdir` and `files` parameters. `DataSource` is unaware of these downstream expectations, it just tracks e.g. the repo and checkout target. This means we can't just copy the logic from `AdditionalCode.get()` (which is aware of the required file subset) to `DataSource.get()` (which is not).

To solve this I'd propose doing something similar to #285, which introduces a `FileInfo` class for a single file that has been fetched and tracked by C-Star, and a `LocalFileStatistics` class for a collection of colocated files*.

In this model, `DataSource` would similarly be split into `SourceFile` and `SourceFileCollection` (or similar). The `SourceFileCollection` could just be a sequence of colocated `SourceFile` instances, or it could be instantiated similarly to `AdditionalCode`, with a `location` (e.g. a repo), a `subdir` (e.g. within that repo, optional), and a list of `files` (e.g. within that subdir, optional). If the latter parameters aren't supplied, the repo (assuming no subdirectories) would be taken as a whole. This could later be extended to buckets, zip archives, anywhere a bunch of files are slapped together in one directory. 

My thought is to additionally hanve an ABC `Retriever` with an abstract `get` method and a series subclasses related to different combinations of attributes on `SourceFile` and `SourceFileCollection` (`RemoteBinaryRetriever` uses pooch, requires a `file_hash`, but `RemoteTextRetriever` just streams plaintext files using `requests`, etc.) - `SourceFile{,Collection}.get()` would then use the appropriate retriever.

\* #285  hasn't been merged so these classes can still be renamed or modified

@ScottEilerman @ankona thoughts?

### DIFF
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -132,7 +132,7 @@ class AdditionalCode(LoggingMixin):
     @property
     def exists_locally(self):
         if not self.local_file_stats:
-            return None
+            return False
         try:
             self.local_file_stats.validate()
         except (FileNotFoundError, ValueError):
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -135,7 +135,7 @@ class AdditionalCode(LoggingMixin):
             return False
         try:
             self.local_file_stats.validate()
-        except (FileNotFoundError, ValueError):
+        except (FileNotFoundError, ValueError, KeyError):
             return False
         return True
 
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -76,7 +76,15 @@ class AdditionalCode(LoggingMixin):
         self.source: DataSource = DataSource(location)
         self.subdir: str = subdir
         self._checkout_target = checkout_target
-        self.files: Optional[list[str]] = [] if files is None else files
+
+        self.files: Optional[list[str]]
+        if files is None:
+            self.files = []
+        elif any((not f) for f in files):
+            raise ValueError("An invalid filename was supplied to AdditionalCode")
+        else:
+            self.files = files
+
         # Initialize object state
         self.local_file_stats: Optional[LocalFileStatistics] = None
 
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -121,18 +121,26 @@ class AdditionalCode(LoggingMixin):
 
     @property
     def checkout_target(self) -> Optional[str]:
+        """If in a repository, the commit hash or tag at which to checkout this
+        additional code."""
         return self._checkout_target
 
     @property
     def working_path(self) -> Optional[Path]:
+        """The current local path where this AdditionalCode exists on the local system,
+        if it has been fetched."""
+
         if self.local_file_stats is None:
             return None
         return self.local_file_stats.parent_dir
 
     @property
-    def exists_locally(self):
+    def exists_locally(self) -> bool:
+        """Determines whether this AdditionalCode instance exists on the local
+        system."""
         if not self.local_file_stats:
             return False
+
         try:
             self.local_file_stats.validate()
         except (FileNotFoundError, ValueError, KeyError):
@@ -212,7 +220,6 @@ class AdditionalCode(LoggingMixin):
                 paths=[local_dir / Path(f).name for f in self.files]
             )
             self.log.info("✅ All files copied successfully")
-            # self.working_path = local_dir
 
         finally:
             if tmp_dir:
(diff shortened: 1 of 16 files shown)


ScottEilerman:
@dafyddstephenson 
Here are a few notes in Jira from when we chatted a few months ago:


> Contrary to design doc, we think we want to make datasources a separate “provider” type of class that just knows how to fetch things to the working directory. This is similar to the current design actually, we just need to move the .get() logic from other classes to datasource, and then split that up into different subclasses for different retrieval methods.
> 
> These can get reused on any number of other classes, like InputDataset, or AdditionalCode, or anything else that needs to move data.
> 

> “DataSource” can be renamed if it makes sense.
> 
> Ones we need include (feel free to rename):
> 
> * RemoteHTTP
> * RemoteGitFiles
> * LocalPath (symlinks)
> 
> Partitioning is related to ROMS InputDatasets and not the data source.

I think this aligns pretty well with your Retriever concept. And then re: ExternalCodebase vs AdditionalCode, we had:

> Contrary to what the design doc says, we are keeping a distinction between ExternalCodebase (will be very similar to current) and AdditionalCode (will look more like a variation on InputDataset, with a DataSource that can get files from wherever). 
> 
>   *  ExternalCodebase remains mostly as-is; we could implement full git checkouts in a DataSource if it makes sense, or just leave that logic where it is
>   *  Make a GitFileRetriever DataSource that can .get() select files, to be used by AdditionalCode (or other classes if desired)


Overall, I think the direction I'd recommend is to think about what a clean `Retriever` class and attached `.get()` method should look like, independent of the current entanglements with the code and input dataset classes. I think the simplest form is: data is somewhere else, and it needs to put it somewhere local for the model to use. Then build out different subclasses for each type of "somewhere else." Then see how we can refactor the existing constructs to use these.

Happy to chat more, but those are my initial notes and thoughts!

dafyddstephenson:
Thanks @ScottEilerman, As I feared, exploring this has led me to totally rethink the `LocalFileStatistics` class, and I think I have a much cleaner plan for the whole thing now. I've put together a skeleton of this in a [fresh PR](https://github.com/CWorthy-ocean/C-Star/pull/300). Sorry for all the nonsense around this, hopefully this is the end of the teething pain, hindsight is 20:20 etc. 😬

PR #300, opened 2025-07-17
Data retrieval refactor design outline

dafyddstephenson:
First off, apologies for the mayhem around this. I think a good step back was needed to try and do this cleanly, and this is that. 

I've made a new branch on the upstream for this refactor and will try to issue small PRs to that branch as the changes get implemented, probably one module or class at a time.

This is the first, basically outlining a skeleton of how I'm planning to refactor C-Star's handling of tracking any external data: how the source is tracked, how the data are retrieved, and how the retrieved version is tracked. In my previous attempts I came to realise that effectively all C-Star cares about is:
1. Where something comes from
2. How to retrieve it from that place
3. Where it is after retrieval
4. Whether it has been changed since retrieval

All the other stuff floating around all over the repo right now (checksums, `os.stat_result`s, checkout targets, etc.) are just used to answer those four questions.

In the design presented here:

### 1.  Where something comes from

A `SourceData` (`retrieval/source_data.py`) instance is instantiated from a minimal set of parameters (only `location` is required). This class is suspiciously similar to `DataSource` from before, inferring via properties as much as possible about the source, but has a `_retriever` attr pointing to an appropriate subclass of a new ABC: `Retriever`.

### 2. How to retrieve it from that place

... `Retriever` (`retrieval/retriever.py`) is a class with a single abstract method, `retrieve()`, allowing each subclass (e.g. `RepositoryRetriever`) to use its own logic to get whatever data from the source. `Retriever.retrieve()` returns a subclass instance of a new ABC: `RetrievedData`

### 3. Where it is after retrieval and 4. whether it has changed since retrieval

... `RetrievedData` (`retrieval/retrieved_data.py`) is a class with two abstract properties: `paths` (`list[Path]`), and `changed_from_source` (`bool`). A `RetrievedFile` can verify its current checksum against a cached one, a `RetrievedRepository` can verify it's checked out at the right target and isn't dirty. The proposed abstract `reset` method could be useful to restore the correct state if `changed_from_source` is true.

I think this design can be made to work with the existing internals, and a lot of the existing fetching and tracking logic can just be reshuffled into the above structure.


cstar/io/source_data.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Inferred data characteristics
+
+    @property
+    def _location_as_path(self) -> Path:
+        """Return self.location as a Path for parsing"""
+        if self.location_type is LocationType.HTTP:
+            return Path(urlparse(self.location).path)
+        elif self.location_type is LocationType.PATH:
+            return Path(self.location)
```
ScottEilerman: consider adding `.resolve()` or `expanduser()` here?

cstar/io/source_data.py
```diff
@@ -0,0 +1,249 @@
...
+
+        sources = [
+            SourceData(location=loc, file_hash=fh, checkout_target=ct)
+            for loc, fh, ct in zip(locations, file_hashes, checkout_targets)
+        ]
+        return cls(sources)
+
+    @property
+    def sources(self) -> list[SourceData]:
+        return list(self._sources)
```
ScottEilerman: I think the list() call here is redundant with init?

cstar/io/staged_data.py
```diff
@@ -0,0 +1,191 @@
...
+    def __init__(
+        self,
+        source: "SourceData",
+        path: Path,
+        sha256: str | None = None,
+        stat: os.stat_result | None = None,
+    ):
+        super().__init__(source, path)
+
+        self._path = path
```
ScottEilerman: already done in init

cstar/io/staged_data.py
```diff
@@ -0,0 +1,191 @@
...
+            self.unstage()
+            self._clear_cache()
+            self.source.stage(target_dir=self.path)
+
+
+class StagedRepository(StagedData):
+    def __init__(self, source: "SourceData", path: Path):
+        super().__init__(source, path)
+        self._source = source
+        self._path = path
```
ScottEilerman: these get done by super()

cstar/io/staged_data.py
```diff
@@ -0,0 +1,191 @@
...
+        self._checkout_hash = _run_cmd(
+            cmd="git rev-parse HEAD", cwd=self.path, raise_on_error=True
+        )
+
+    @property
+    def changed_from_source(self) -> bool:
+        """Check if the current repo is dirty or differs from a given commit hash."""
+        # Check existence
+        if not self.path.exists():
+            return True
```
ScottEilerman, line 109: Per conversation, this could cause reset to fail

cstar/io/stager.py
```diff
@@ -0,0 +1,85 @@
...
+class Stager(ABC):
+    @abstractmethod
+    def stage(self, target_dir: Path, source: "SourceData") -> "StagedData":
+        """Stage this data using an appropriate strategy."""
+
+
+class RemoteBinaryFileStager(Stager):
+    # Used for e.g. a remote netCDF InputDataset
+    def stage(self, target_dir: Path, source: "SourceData") -> "StagedFile":
+        """Stage a remote binary file with hash verification using Pooch."""
```
ScottEilerman: re docstring: no longer using pooch

cstar/io/stager.py
```diff
@@ -0,0 +1,85 @@
...
+            sha256=(source.file_hash or None),
+            stat=None,
+        )
+
+
+class LocalBinaryFileStager(Stager):
+    # Used for e.g. a local netCDF InputDataset
+    def stage(self, target_dir: Path, source: "SourceData") -> "StagedFile":
+        """Create a local symlink to a binary file on the current filesystem."""
+        target_dir.symlink_to(source.location)
```
ScottEilerman: Per conversation, we actually want to make a target_path out of target_dir and the file name, and use that to symlink.

cstar/io/staged_data.py
```diff
@@ -0,0 +1,191 @@
...
+                )
+
+    def __len__(self) -> int:
+        return len(self._items)
+
+    def __getitem__(self, idx: int) -> StagedData:
+        return self._items[idx]
+
+    def __iter__(self) -> Iterator[StagedData]:
+        return iter(self._items)
```
ScottEilerman, line 170: @ankona I know you mentioned using generics to get some of this behavior for free. I know I've done it in the past but am blanking on what exactly that looks like. Do we just subclass `MyCollection(list[StagedData])`? It can't just be a type definition since we do want to add custom methods that call through to the collected objects (e.g. the methods below this comment).

cstar/io/source_data.py
```diff
@@ -0,0 +1,249 @@
...
+    BINARY = "binary"
+
+
+class SourceData:
+    def __init__(
+        self, location: str | Path, file_hash: str | None, checkout_target: str | None
+    ):
+        self._location = str(location)
+        self._file_hash = file_hash
+        self._checkout_target = checkout_target
```
ScottEilerman: consider a factory approach for these -- similarly, it's a little confusing to have checkout_target for files and file_hash on repositories, but not sure at the moment how to separate that, or if they should be separate classes altogether.

cstar/io/source_data.py
ScottEilerman: And in general, if it can be done on init of these sorts of objects, vs needing to remember to do it in methods later.

cstar/io/source_data.py
```diff
@@ -0,0 +1,307 @@
...
+
+@dataclass
+class SourceCharacteristics:
+    source_type: SourceType
+    location_type: LocationType
+    file_encoding: FileEncoding
+
+
+class SourceClassification(SourceCharacteristics, Enum):
+    REMOTE_TEXT_FILE = SourceType.FILE, LocationType.HTTP, FileEncoding.TEXT
```
ScottEilerman: do these enum values need `SourceCharacteristics()` around the arguments, or is there some magic with inheriting from `SourceCharacteristics` that automatically passes the tuple as args?

ScottEilerman: Ok, I think I see, down in .classify... this could be the cause of your mypy ignore conundrum too. I fiddled and might have a fix, will push a branch in a minute.

cstar/io/staged_data.py
```diff
@@ -0,0 +1,191 @@
...
+            self._sha256 = sha256
+        else:
+            self._sha256 = _get_sha256_hash(self.path)
+
+        if stat:
+            self._stat = stat
+        else:
+            self._stat = os.stat(self.path)
+
+    # Abstract
```
ScottEilerman: I think this comment  might be out of sync with whatever it was originally commenting on?

ScottEilerman (commented):
Looking pretty good! I left one comment about the enum and mypy error that is maybe not essential, but while playing with the code to understand that, I got some circular import errors. I have a fix (I think) where I split out some of the enums and basic types into a module (makes it easy to import everywhere since it has no dependencies), and also got rid of the io module's __init__. I'm not sure if getting rid of the latter was necessary or not. I tend to not forward imports up to the __init__ because I find it harder to understand where things really are coming from, and in my experience it becomes _more_ problematic in terms of these sorts of circular imports, but I could be entirely wrong because I know it's fairly common practice, so open to counter-arguments.

Here's my commit with some possible changes/fixes: https://github.com/ScottEilerman/C-Star/commit/b64cba8d9a69fcf9ec58c0ece402c2fd24c68ed3#r163466961

ScottEilerman (commented):
Looking pretty good! I left one comment about the enum and mypy error that is maybe not essential, but while playing with the code to understand that, I got some circular import errors. I have a fix (I think) where I split out some of the enums and basic types into a module (makes it easy to import everywhere since it has no dependencies), and also got rid of the io module's __init__. I'm not sure if getting rid of the latter was necessary or not. I tend to not forward imports up to the __init__ because I find it harder to understand where things really are coming from, and in my experience it becomes _more_ problematic in terms of these sorts of circular imports, but I could be entirely wrong because I know it's fairly common practice, so open to counter-arguments.

Here's my commit with some possible changes/fixes: https://github.com/ScottEilerman/C-Star/commit/b64cba8d9a69fcf9ec58c0ece402c2fd24c68ed3#r163466961

dafyddstephenson:
@ScottEilerman , cherry-picked your commit and did some tidying following comments. Thanks! Anything else?

ScottEilerman:
> @ScottEilerman , cherry-picked your commit and did some tidying following comments. Thanks! Anything else?

Nope, go for it! 🚀

ScottEilerman (approved):

PR #302, opened 2025-07-18
New linting rules (`main`)

dafyddstephenson:
- Extends existing `ruff` rules from defaults & `I` to include `F, E, W, D, UP`, with 10 ignores to address later (c5d6b8735ec8140a6726486aae8857133e9f7714) 
- Applies this new ruleset and fixes issues (47f7082b55236213d755b25d57c1c4a89df4f4d4) 
- Fixes two syntactic breakages in the unit tests (478ea16b023ba8b79753de54ccc586e0a4f55f81)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 files: '^cstar/'
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     
     hooks:
       - id: trailing-whitespace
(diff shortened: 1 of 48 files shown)


.pre-commit-config.yaml
```diff
@@ -38,10 +39,10 @@ repos:
...
       - id: absolufy-imports
         args: [--application-directories=.]
-
-  - repo: https://github.com/PyCQA/docformatter
-    rev: v1.7.5  
-    hooks:
-      - id: docformatter
-        language: python
-        args: ["--in-place", "--recursive", "--wrap-summaries=88", "--wrap-descriptions=88"]
+# TODO: figure out the contradiction, or determine if ruff is doing everything we need here
```
ScottEilerman: Based on the relatively few things that [docformatter](https://github.com/PyCQA/docformatter) is doing, I am pretty sure ruff is doing the same, so we can probably go ahead and drop docformatter. Can always add it back later if for some reason we think we're missing something.

ScottEilerman (approved):

Issue #251, opened 2025-05-01
Try ZenHub for issue tracking


ScottEilerman:
This issue is being used to test out various integrations with ZenHub. Change description.

ScottEilerman:
Comment on github issue from zenhub

ScottEilerman:
Comment on github issue from github.

ScottEilerman:
Comment from Github Project

ScottEilerman:
![](https://github.trello.services/images/mini-trello-icon.png) [My Github issue](https://trello.com/c/sVGUZKEw/2-my-github-issue)

JoeGaines:
Comment from Joe.

JoeGaines:
Close this one?

PR #305, opened 2025-07-21
  Change project key for Jira integration

ScottEilerman:
Change project key for Jira integration

### DIFF
--- a/.github/workflows/create_jira_tickets.sh
+++ b/.github/workflows/create_jira_tickets.sh
@@ -42,7 +42,7 @@ if [[ -n "$JIRA_ASSIGNEE_ID" ]]; then
   cat > payload.json <<EOF
 {
   "fields": {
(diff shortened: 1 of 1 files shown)


dafyddstephenson (approved):
can't see there being a problem but we won't know until we try it out anyway!

ScottEilerman:
> can't see there being a problem but we won't know until we try it out anyway!

I'm going to test it out here before doing the same in the other repos, stand by!

ScottEilerman:
> > can't see there being a problem but we won't know until we try it out anyway!
> 
> I'm going to test it out here before doing the same in the other repos, stand by!

Used [this issue](https://github.com/CWorthy-ocean/C-Star/issues/251), looks like it worked!

dafyddstephenson:
I also just learned from this that you can add the jira label long after the issue is made to make the ticket - cool!

PR #307, opened 2025-07-22
CW-801: Add MVP container files

ankona:
Add MVP container files

- [x] Closes #CSD-121


ci/docker/build.sh
```diff
@@ -0,0 +1,130 @@
...
+if [ ! -z "$BUILD_ARGS" ]; then
+    log_info "Building parameterized $VTAG in $CONTEXT_PATH from source \n$CONTEXT_DIR\n$BUILD_ARGS"
+    $RUNTIME_ENGINE build -t "$VTAG" --no-cache --build-arg "$BUILD_ARGS" "$CONTEXT_PATH" 
+else
+    log_info "Building $VTAG in $CONTEXT_PATH from source \n $CONTEXT_DIR"
+    $RUNTIME_ENGINE build -t "$VTAG" --no-cache "$CONTEXT_PATH"
+fi
+
+log_info "Migrating $VTAG"
+$RUNTIME_ENGINE migrate "$VTAG"
```
ScottEilerman: We'll want these behind an "if on perlmutter" check. Down below, migration is checking for podman-hpc first, but thinking ahead to possibly using podman-hpc on AWS, I guess we really want the check to be perlmutter-specific. Hopefully we can use whatever system inspection logic we use on the python side of things.

ci/docker/build.sh
```diff
@@ -0,0 +1,130 @@
...
+
+if [ -z "$REPO" ]; then
+    REPO=$REPO_DEFAULT
+    log_info "Using default repo: $REPO" 
+fi
+
+ENGINES=("podman-hpc" "podman" "docker")
+RUNTIME_ENGINE=""
+
+for item in "${ENGINES[@]}"; do
```
ScottEilerman: As noted in chat, something in this loop didn't work for me on mac when I didn't have podman-hpc. It just exited out instead of moving on to podman, and I had to hard-code podman on line 70. I am bad at bash so can't offer too much more insight into what may be wrong here...

ci/docker/build.sh
```diff
@@ -0,0 +1,130 @@
...
+    log_error "Dockerfile not found in $CONTEXT_PATH"
+    exit 1
+fi
+
+cp "$CONTEXT_PATH/../entrypoint.sh" $CONTEXT_PATH/entrypoint.sh
+chmod a+r $CONTEXT_PATH/entrypoint.sh
+
+if [ ! -z "$BUILD_ARGS" ]; then
+    log_info "Building parameterized $VTAG in $CONTEXT_PATH from source \n$CONTEXT_DIR\n$BUILD_ARGS"
+    $RUNTIME_ENGINE build -t "$VTAG" --no-cache --build-arg "$BUILD_ARGS" "$CONTEXT_PATH"
```
ScottEilerman: Do we actually want `--no-cache` here and in line 107, or was this just for initial dev?

ci/docker/condabase/Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG IMG_BUILDBASE="ankona/cstar-buildbase:latest"
+
+FROM $IMG_BUILDBASE as main
+
+ARG MINICONDA_SCRIPT="https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh" 
+ARG INSTALL_BASE_DIR="/opt"
+ARG CSTAR_REPO="ankona/C-Star"
+ARG CSTAR_HASH="CW-799"
```
ScottEilerman: I guess we should update these defaults to the primary repo, and... `develop`?

ci/docker/condabase/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ARG CONDA_HOME="$INSTALL_BASE_DIR/miniconda3"
```
ScottEilerman: two-part reminder from our earlier conversation: consider a no-conda build, and/or get rid of the `x86_64-conda-linux-gnu` leftover thing in the miniconda3 folder.

ci/docker/Makefile
```diff
@@ -0,0 +1,96 @@
+CI_DIR      := $(patsubst %/,%, $(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
+ACCOUNT     := ankona
```
ScottEilerman: not sure if there's a convention for default account, maybe `$whoami`? probably not going to be the same for everyone's accounts but maybe the best starting place.

similarly, we should consider how to handle throughout the PR (ctrl-f ankona, default build args, image names, etc.)

ci/docker/roms/Dockerfile
```diff
@@ -0,0 +1,40 @@
+ARG IMG_NCDF="ankona/cstar-netcdf:latest"
+
+FROM $IMG_NCDF
+
+
+ARG SRC_DIR="/install"
+ARG INSTALL_BASE_DIR=/opt
+ENV ROMS_ROOT=$INSTALL_BASE_DIR/ucla-roms 
+ENV ROMS_PATH=$ROMS_ROOT/Tools-Roms/
+ARG REPO_ROMS="https://github.com/dafyddstephenson/ucla-roms/archive/refs/heads/update_makefiles.tar.gz"
```
ScottEilerman: This is probably due for an update too.

ScottEilerman (commented):
Great work on this @ankona ! It's going to be so nice to have these small modular images going forward. And there are so many nitty gritty build details in here that I hope I'll never need to think about.

As we discussed in chat, with a few small fixes to the build script, I was able to build everything on my macbook (which was ill-advised, since though the base image is for amd64 and not arm64, so it didn't end up working at the end, but that's my fault and the fact that it built at all is still notable).

Three high-level questions/suggestions in addition to the comments:
- Do we want to embrace podman/OSS and rename Dockerfile -> Containerfile everywhere?
- Even if this isn't the final or permanent version of the build script, spending a quick 10 mins on a readme with some basic instructions (`make pull` if you want to start from existing stuff, `make build all` and/or `make clean` if you want to rebuild from scratch, what env-vars or arguments you might need to pass/set, etc.) will be handy for future users. In my whirlwind "can I build this" session, I just changed the defaults at the top of the Makefile, which is not what we should encourage. 🙈 
- I got things to work with my personal docker.io account, is it worth a test with either our github registry or NERSC's Harbor registry? Could easily be a future task too.

ci/docker/build.sh
ankona: Added conditional to only migrate if using `-hpc` engine.

ci/docker/build.sh
ankona: Removed --no-cache

ci/docker/condabase/Dockerfile
ankona: yes! `CW-799` was the only branch that had the worker script during dev. good catch. updated to `cworthy-ocean/c-star` and `develop`.

ci/docker/build.sh
ankona: Swapped to `command -v $item` to correct broken behavior.

ci/docker/condabase/Dockerfile
ankona: Added a non-conda image and updated runner to use it, instead. Image is ~half the size.

ci/docker/Makefile
ankona: You should be able to pass the account like a make variable: `make runner ACCOUNT=foo`, but that won't achieve the goal due to not passing parameters into each image build.

I've created a follow-up [ticket](https://cworthy.atlassian.net/browse/CSD-179) to parameterize those steps properly

ankona:
> * Do we want to embrace podman/OSS and rename Dockerfile -> Containerfile everywhere?

1. Updated `Dockerfile` to `Containerfile`
2. Working on a `README.md`
3. Created follow-up [ticket](https://cworthy.atlassian.net/browse/CSD-180) to test pushing to alternative container repositories.

ci/docker/roms/Dockerfile
ankona: To avoid running into issues with building the fork, I've created a [follow-up ticket](https://cworthy.atlassian.net/browse/CSD-181).

dafyddstephenson: I don't know where this originated from but yeah we should be way ahead of this

ScottEilerman (approved):
🙌🙌🙌

PR #311, opened 2025-07-31
Control max nodes based on hyperthreads

### LINKED ISSUE #310: Max CPUs on Perlmutter is giving us logical cpus instead of physical cores
In our logic for Slurm jobs, we ask the system [for the max cpus per node](https://github.com/CWorthy-ocean/C-Star/blob/main/cstar/system/scheduler.py#L416) to determine how many nodes to ask for given the number of partitions / tasks in a job. On perlmutter, this returns 256, which is the [number of logical cpus / hyperthreads](https://docs.nersc.gov/jobs/affinity/#cpu-only-nodes). For ROMS, because it is CPU-bound, the number we actually want is 128, the number of physical cores. This means we're not spreading out to the number of nodes we expect when using C-Star.

My proposed fix for this is to simply divide the number returned by scontrol by 2, but I wanted to run it by folks first. In theory other applications might want to make use of the hyperthreads, so this feels sus to put into the general implementation. I also don't want to do a ton of additional coding for a theoretical use case though.

As I type that, I think the answer is a quick param to choose whether you want hyperthreads, and default it to False. I'll make a PR with that in mind.

ScottEilerman:
- [x] Closes #310 
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/cstar/execution/scheduler_job.py
+++ b/cstar/execution/scheduler_job.py
@@ -178,7 +178,6 @@ class SchedulerJob(ExecutionHandler, ABC):
         queue_name: str | None = None,
         send_email: bool | None = True,
         walltime: str | None = None,
-        use_hyperthreads: bool = False,
     ):
         """Initialize a SchedulerJob instance.
 
@@ -218,10 +217,6 @@ class SchedulerJob(ExecutionHandler, ABC):
         walltime : str, optional
             The maximum walltime for the job, in the format "HH:MM:SS". If not provided,
             it defaults to the queue's maximum walltime.
-        use_hyperthreads : bool, optional
-            Whether to consider hyperthreads when calculating the node distribution, as opposed
-             to only using physical cores, which is usually the desired behavior for CPU-bound models.
-             Defaults to False (physical cores only).
 
         Raises
         ------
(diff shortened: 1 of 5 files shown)


cstar/execution/scheduler_job.py
```diff
@@ -314,17 +319,21 @@ def __init__(
                     + "your system's CPUs per node automatically and cannot continue"
                 )
 
-            nnodes, ncpus = self._calculate_node_distribution(
-                cpus, scheduler.global_max_cpus_per_node
+            max_cpus = (
+                scheduler.global_max_cpus_per_node
+                if use_hyperthreads
+                else scheduler.global_max_cpus_per_node // 2
```
dafyddstephenson: will this reliably be 2? And will we ever use hyperthreading? Perhaps nipping this at the property def level and only supporting physical cores is the move ?

ScottEilerman: Just did a quick google search and it says it's effectively always two.

Definitely open to just moving it to the property. Maybe something like partitioning is more IO bound and could use the hyperthreads fine? Though you could also just specify your nodes/cpus_per_node then. I'm fine either way if you have a preference.

ScottEilerman: I guess it _is_ a real possibility that an HPC has only one HT and the value should be one. Hmm.

dafyddstephenson: I'd say the consideration re. partitioning is sufficiently far away that we can address it as and when rather than adding complexity here _but_ you've already written the code and it works and it's literally 20 lines so 🤷‍♂️ ultimately I'd love to solve this whole automatic process distribution thing some other way, the current implementation makes me a bit sad, so perhaps with some cleverness that previously eluded me this problem can go altogether. This works great for now

dafyddstephenson: The only thing I could think of before was sending a trivial test job with the same characteristics to the queue to see how it responded, then getting the stats and immediately cancelling it. Which was even more offensive to me than hardcoding 128 🙃

ScottEilerman: I decided to go in the direction of hard-coding 128 from the Perlmutter system context.

We can revisit whenever we get to doing system configs for MVP -- the orchestrator will generally be the thing that becomes concerned with this since it will be submitting the jobs, but I'm not quite sure what this looks like yet.

dafyddstephenson (approved):

PR #312, opened 2025-08-08
Jira integration: Add tasks to current sprint and fix subtasks

ScottEilerman:
Makes several modifications to the Jira integration:
- adds created stories to the current sprint
- fixes subtask field ID (it was different on the new Jira board)
- moves user ids and other magic vars to org-wide runner variables 

I tested individual API calls and some of the `jq` logic locally but it's hard to test the action itself without merging.


dafyddstephenson (approved):
Don't know if I'm clever enough to spot any issues with this without it running, might as well let 'er rip

PR #320, opened 2025-08-13
Fix image building defect on ARM64 platform

ankona:
This PR fixes build errors that occur when building and running container images on the `ARM64` platform. The changes include:

- Install `python3-dev` and `gdal` dependencies to allow setuptools to build packages correctly during c-star installation
- Use correct naming convention for an env file (rename file, replace "-" with "_")
- Add `_SystemContext` subclass for `ARM64`
- Copy runner script to image

- [x] Closes #CSD-223
- [x] Tests passing


ScottEilerman (approved):
Was able to build on my mac, and test by bashing into the container, running python, and importing cstar.

PR #321, opened 2025-08-13
Use consistent naming for action target and logic

ankona:
This is a small quality-of-life improvement for the container `Makefile`. It improves the clarity of the "fetch" action by renaming variables and functions to use "pull," instead.


ankona:
Switching to appropriately named branch to trigger Jira integration.

PR #325, opened 2025-08-15
CSD-101 - Mitigate nullable type and __init__-based member declaration mypy errors

ankona:
Qualify of life type-hinting improvements to fix some issues encountered when consuming `Simulation` classes from the worker.

- Shift member variable declarations out of `__init__` so they are discoverable
- Wrap conditionals around use of nullable fields
- Convert `discretization` field to generic parameter  

### Simulation.discretization shadowing

image

### Simulation.runtime_code shadowing

image

### Checklist

- [x] Tests passing

### DIFF
--- a/cstar/roms/simulation.py
+++ b/cstar/roms/simulation.py
@@ -296,7 +296,8 @@ class ROMSSimulation(Simulation):
 
         Returns
         -------
-        str : The file path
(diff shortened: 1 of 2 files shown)


ankona:
not implementing...

PR #328, opened 2025-08-18
CSD-101 - Enable per-stage execution in worker

ankona:
This change parameterizes the stages to be executed by the worker.

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


ankona:
Moving to correctly named branch per subtask.

PR #329, opened 2025-08-18
CSD-226 - Enable parameterized execution of simulation stages

ankona:
This change parameterizes the stages to be executed by the worker.

- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


ScottEilerman (approved):
LGTM! 

Part of me thinks we should consider moving all of the stages into `_on_iteration`, and leave `_on_start` and `_on_shutdown` to only be about the worker process itself? But for now it'll also work as-is since we effectively go through the sequence once anyway. In the spirit of iterating when we know more, we can go forward as-is and change later if we need to, unless you 100% agree and wanna do it now.

PR #336, opened 2025-08-27
Centralise class-instance-creating unit test fixtures and establish naming conventions

dafyddstephenson:
This PR:
- Moves all unit test fixtures that generate an instance of a class to `unit_tests/conftest.py`
- Moves all unit-test-specific subclass definitions of `ABC`s to `unit_tests/fake_abc_subclasses.py` for import (e.g. `FakeExternalCodeBase` which subclasses `ExternalCodeBase` to represent a generic subclass of this ABC)
- Establishes consistent naming of the form `<instance_type>_<class>_<nature>`, for fixtures generating class instances e.g. `fake_roms_input_dataset_netcdf_local`, where:
   -  `<instance_type>` is e.g. `mock`, `fake`, `dummy`, following e.g. [Meszaros](https://martinfowler.com/bliki/TestDouble.html)
   - `<class>` is the class of which an instance is being generated
   - `<nature>` is any extra adjectivey information

This PR does not:
- add, remove, or substantially change any existing unit tests.
- define any new fixtures or classes

Happy to discuss the conventions, but hoping to get at least the centralisation part over the line as it'll really help with the other refactor I'm working on. Cheers :)


cstar/tests/unit_tests/fake_abc_subclasses.py
```diff
@@ -0,0 +1,129 @@
...
+from cstar.roms import ROMSInputDataset
+from cstar.roms.runtime_settings import ROMSRuntimeSettingsSection
+
+
+class FakeExternalCodeBase(ExternalCodeBase):
+    """A mock subclass of the `ExternalCodeBase` abstract base class used for testing
+    purposes.
+    """
+
+    # def __init__(self, log: logging.Logger):
```
ankona: This commented code can be removed

cstar/tests/unit_tests/fake_abc_subclasses.py
```diff
@@ -0,0 +1,129 @@
...
+    simulation execution.
+
+    Attributes
+    ----------
+    default_codebase : ExternalCodeBase
+        Returns an instance of `FakeExternalCodeBase`.
+
+    Methods
+    -------
+    from_dict(simulation_dict, directory)
```
ankona: Adding accurate method descriptions (e.g. no-op) is a good call but placing them here in the class docstring leaves tooling to display the parent class's method docstrings when in use. 

These should probably move to the actual methods. Here are a couple examples showing the incorrect hints when in use:

image

neovim example:

image

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    - `files`: A list of files to be included from the local directory
+
+    This fixture can be used in tests that involve handling or manipulating
+    code that resides on the local filesystem.
+
+    Returns:
+        AdditionalCode: An instance of the AdditionalCode class with preset
+        local directory details.
+    """
+    return AdditionalCode(
```
ankona, line 86: ditto on dummy. i'm torn on being more descriptive.

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    in order to test the `ROMSRuntimeSettings.to_file` and `from_file` methods.
+
+    Paths do not correspond to real files.
+
+    Yields
+    ------
+    ROMSRuntimeSettings
+       The example ROMSRuntimeSettings instance
+    """
+    yield ROMSRuntimeSettings(
```
ankona, line 126: not really fake... see main discussion comment.

cstar/tests/unit_tests/fake_abc_subclasses.py
```diff
@@ -0,0 +1,129 @@
...
+
+    floats: list[float]
+    paths: list[Path]
+    others: list[str | int]
+    floatval: float
+    pathval: Path
+    otherval: str
+
+
+class FakeROMSRuntimeSettingsSectionEmpty(ROMSRuntimeSettingsSection):
```
ankona, line 42: Do we need this at all? Couldn't we avoid the `Fake` and use the original class to avoid the maintenance?

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    - `files`: A list of files to be included from the repository
+
+    This fixture can be used in tests that involve handling or manipulating code
+    fetched from a remote Git repository.
+
+    Returns:
+        AdditionalCode: An instance of the AdditionalCode class with preset
+        remote repository details.
+    """
+    return AdditionalCode(
```
ankona, line 58: Since this returns a real `AdditionalCode` instance without any patching/subclassing, this feels like it doesn't meet the criteria of a fake. 

It would break if attempts were made to use it without knowing (breaking that "working implementations" rule. Should this be named `dummy_add_code_rem`?

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+
+
+@pytest.fixture
+def marbl_codebase():
+    """Fixture providing a configured instance of `MARBLExternalCodeBase` for
+    testing.
+    """
+    source_repo = "https://github.com/marbl-ecosys/MARBL.git"
+    checkout_target = "marbl0.45.0"
+    return MARBLExternalCodeBase(
```
ankona: should this fixture follow the naming convention?

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+        source_repo=source_repo, checkout_target=checkout_target
+    )
+
+
+@pytest.fixture
+def roms_codebase():
```
ankona: should this fixture follow the naming convention?

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    return ROMSExternalCodeBase(
+        source_repo=source_repo, checkout_target=checkout_target
+    )
+
+
+################################################################################
+# ROMSRuntimeSettings
+################################################################################
+@pytest.fixture
+def fake_roms_runtime_settings():
```
ankona: Should the naming convention keep class names as a single element in snake-case to make it clearer where the type name ends (e.g. `fake_romsruntimesettings_local`)?

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    up the following attributes:
+
+    - `location`: The path to the local directory containing the code
+    - `subdir`: A subdirectory within the local directory where the files are located
+    - `files`: A list of files to be included from the local directory
+
+    This fixture can be used in tests that involve handling or manipulating
+    code that resides on the local filesystem.
+
+    Returns:
```
ankona: inconsistent trailing `:` and missing `-------`

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+
+    - `location`: The URL of the remote repository
+    - `checkout_target`: The specific branch, tag, or commit to checkout
+    - `subdir`: A subdirectory within the repository where files are located
+    - `files`: A list of files to be included from the repository
+
+    This fixture can be used in tests that involve handling or manipulating code
+    fetched from a remote Git repository.
+
+    Returns:
```
ankona: inconsistent trailing `:` and missing `-------`

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    """Fixture providing a `FakeSimulation` instance for testing.
+
+    This fixture sets up a minimal `FakeSimulation` instance with a mock external
+    codebase, runtime and compile-time code, and basic discretization settings.
+    The temporary directory (`tmp_path`) serves as the working directory for the
+    simulation.
+
+    Yields
+    ------
+    tuple[FakeSimulation, Path]
```
ankona: Does this need to return a tuple since the path can be accessed via `sim.directory`?

cstar/tests/unit_tests/conftest.py
```diff
@@ -1,11 +1,533 @@
...
+    temporary directory (`tmp_path`) is used as the working directory.
+
+    Yields
+    ------
+    tuple[ROMSSimulation, Path]
+        A tuple containing:
+        - `ROMSSimulation` instance with fully configured attributes.
+        - The temporary directory where the simulation is stored.
+    """
+    directory = tmp_path
```
ankona, line 446: can we remove this unnecessary rename/assignment?

cstar/tests/unit_tests/fake_abc_subclasses.py
```diff
@@ -0,0 +1,129 @@
...
+    """No actual values defined, just used to test some of the formatting/joining
+    functions.
+    """
+
+
+class FakeInputDataset(InputDataset):
```
ankona, line 50: ditto on usefulness. I'm not sure the description as a mock is valid (since it doesn't change any behavior)

cstar/tests/unit_tests/fake_abc_subclasses.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Since InputDataset is an abstract base class, this mock class is needed to allow
+    instantiation for testing purposes. It inherits from InputDataset without adding any
+    new behavior, serving only to allow tests to create and manipulate instances.
+    """
+
+    pass
+
+
+class FakeROMSInputDataset(ROMSInputDataset):
```
ankona, line 61: ditto on usefulness. I'm also not sure the description as a minimal example makes sense since it's implementation matches the non-minimal implementation of `ROMSInputDataset`

cstar/tests/unit_tests/fake_abc_subclasses.py
```diff
@@ -0,0 +1,129 @@
...
+    """A minimal example subclass of the ROMSInputDataset abstract base class."""
+
+    pass
+
+
+class FakeSimulation(Simulation):
```
ankona: Would this meet the definition of `Stub` instead of `Fake` since it doesn't have a working implementation and ignores user inputs?

cstar/tests/unit_tests/roms/test_roms_input_dataset.py
```diff
@@ -147,20 +27,20 @@ class TestStrAndRepr:
...
 
-    def test_str_with_partitioned_files(self, local_roms_netcdf_dataset):
+    def test_str_with_partitioned_files(self, fake_roms_input_dataset_netcdf_local):
         """Test the ROMSInputDataset string representation has correct substring for
         partitioned_files.
 
         Fixtures:
         ---------
-        - local_roms_netcdf_dataset: Provides a ROMSInputDataset with a local NetCDF source.
+        - fake_roms_input_dataset_netcdf_local: Provides a ROMSInputDataset with a local NetCDF source.
```
ankona: inconsistent `:` usage in docstring (looks like 100+ usage without `:` and only 18 with. Can we make these consistent while we're already modifying the docstrings?

cstar/tests/unit_tests/roms/test_roms_input_dataset.py
```diff
@@ -230,15 +110,15 @@ def test_repr_with_partitioned_files_and_working_path(
 
         Fixtures:
         ---------
-        - `local_roms_netcdf_dataset`: Provides a mock ROMSInputDataset object
+        - `fake_roms_input_dataset_netcdf_local`: Provides a mock ROMSInputDataset object
```
ankona: inconsistent `:` usage in docstring (looks like 100+ usage without `:` and only 18 with. Can we make these consistent while we're already modifying the docstrings?

cstar/tests/unit_tests/roms/test_roms_input_dataset.py
```diff
@@ -381,7 +261,7 @@ def test_get_grid_from_remote_yaml(
 
         Fixtures:
         ---------
-        - `remote_roms_yaml_dataset`: Provides a ROMSInputDataset instance with a remote YAML source.
+        - `fake_roms_input_dataset_yaml_remote`: Provides a ROMSInputDataset instance with a remote YAML source.
```
ankona: inconsistent `:` usage in docstring (looks like 100+ usage without `:` and only 18 with. Can we make these consistent while we're already modifying the docstrings?

cstar/tests/unit_tests/roms/test_roms_input_dataset.py
```diff
@@ -451,7 +335,7 @@ def test_get_surface_forcing_from_local_yaml(
 
         Fixtures:
```
ankona: inconsistent `:` usage in docstring

cstar/tests/unit_tests/roms/test_roms_simulation.py
```diff
@@ -858,22 +777,22 @@ def test_str(
 
         Parameters
         ----------
-        example_roms_simulation : fixture
+        fake_roms_simulation : fixture
```
ankona: Do we need to duplicate the fixture name and description? 
- If yes, we're missing it in a lot of places (but I don't think this is the answer)
- If no, let's remove.

cstar/tests/unit_tests/base/test_additional_code.py
```diff
@@ -79,16 +24,17 @@ class TestInit:
         when optional attributes are not provided.
     """
 
-    def test_init(self, remote_additional_code):
+    def test_init(self, fake_additional_code_remote):
```
ankona: I know this is the right type under the hood, but I find myself considering the oddity of validating behavior of a fake implies that the Fixture is incorrectly named. 

Alternatively, this looks like a mistake I've made where I used a fixture but this test really should create it's own concrete instance and not receive a Fake/Stub/whatever.

ankona (changes requested):
Looks pretty good, overall. Quibbles include:

- The `stub` or `dummy` naming might be appropriate in a couple cases; see individual notes.
- Since we're moving code around, some docstring inconsistencies/nitpicks should be handled.

The only thing that I dislike here is the creation of subtypes that add no value.
  - e.g. `FakeROMSInputDataset` and  `FakeROMSRuntimeSettingsSectionEmpty` have no behavioral changes and calling them `fake` is misleading.

It could be a follow-up ticket, but since we're already messing with updating docstrings and moving a lot of things around, this is a missed opportunity to add missing typehints, even if incomplete. For example, the fixtures returning fakes should include the expected contract of the fake (e.g. `-> InputDataset`)

cstar/tests/unit_tests/conftest.py
dafyddstephenson: Think this should be considered fake as all the paths in it point to nowhere 🤔

cstar/tests/unit_tests/fake_abc_subclasses.py
dafyddstephenson: The parent class is an ABC, so we can't create an instance of it. Did you mean something else?

cstar/tests/unit_tests/conftest.py
dafyddstephenson: I believe not - it is "fake" as it contains paths that lead nowhere. I don't think we can call it a "dummy" here as methods on the class are actually called - it's not just brought into existence to satisfy a need for it to exist.

cstar/tests/unit_tests/fake_abc_subclasses.py
ankona: Nope, I missed it. Carry on, nothing to see here!

cstar/tests/unit_tests/conftest.py
dafyddstephenson: this is actually a real instance with legit attribute values. Have just rewritten the tests that use it to avoid it being a fixture at all.

cstar/tests/unit_tests/conftest.py
dafyddstephenson: ditto above

cstar/tests/unit_tests/fake_abc_subclasses.py
dafyddstephenson: `InputDataset` is an ABC so we can't create an instance without subclassing. Updated docstring.

cstar/tests/unit_tests/fake_abc_subclasses.py
dafyddstephenson: See above

dafyddstephenson:
> Looks pretty good, overall. Quibbles include:
> 
>     * The `stub` or `dummy` naming might be appropriate in a couple cases; see individual notes.
> 
>     * Since we're moving code around, some docstring inconsistencies/nitpicks should be handled.
> 
> 
> The only thing that I dislike here is the creation of subtypes that add no value.
> 
>     * e.g. `FakeROMSInputDataset` and  `FakeROMSRuntimeSettingsSectionEmpty` have no behavioral changes and calling them `fake` is misleading.
> 
> 
> It could be a follow-up ticket, but since we're already messing with updating docstrings and moving a lot of things around, this is a missed opportunity to add missing typehints, even if incomplete. For example, the fixtures returning fakes should include the expected contract of the fake (e.g. `-> InputDataset`)

Thanks for the review! All comments addressed in 01d243e4042c2590ceb66e70146233a71682623e
Thought it'd be a quick one but probably should have split into multiple commits, sorry!

ankona (approved):
LGTM!

PR #337, opened 2025-09-09
CSD-39: Add initial workplan and step models

ankona:
Add the initial versions of the models for `WorkPlan` and `Step`

- [x] Closes #CSD-39
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/cstar/orchestration/models.py
+++ b/cstar/orchestration/models.py
@@ -119,4 +119,17 @@ class WorkPlan(BaseModel):
     @field_validator("steps", mode="before")
     @classmethod
     def _deep_copy_steps(cls, value: list[Step]) -> list[Step]:
+        """Ensure the steps provided are deep copied to avoid external change propagation.
+
+        Parameters
+        ----------
+        value : list[Step]
+            The list of steps assigned to the instance
+
+        Returns
+        -------
+        list[Step]
+            The deep-copied step list
+
+        """
         return deepcopy(value)
--- a/cstar/orchestration/models.py
+++ b/cstar/orchestration/models.py
@@ -1,0 +1,95 @@
+import typing as t
+from enum import StrEnum, auto
+
+from pydantic import BaseModel, Field, FilePath, StringConstraints
+
+RequiredString: t.TypeAlias = t.Annotated[
+    str,
+    StringConstraints(strip_whitespace=True, min_length=1),
+]
+
+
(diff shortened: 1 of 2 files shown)


cstar/orchestration/models.py
```diff
@@ -0,0 +1,135 @@
...
+    name: RequiredString
+    """The user-friendly name of the workplan."""
+    description: RequiredString
+    """A user-friendly description of the workplan."""
+    steps: t.Annotated[
+        t.Sequence[Step],
+        Field(
+            default_factory=list,
+            min_length=1,
+            frozen=True,
```
ScottEilerman: This is probably _pedantic_ (and also pydantic), but isn't a frozen list a tuple? I doubt it actually matters, but this is how you know I actually looked at the PR and didn't just rubber stamp it.

cstar/tests/unit_tests/orchestration/test_models.py
```diff
@@ -0,0 +1,1099 @@
...
+        dumper: yaml.Dumper,
+        data: pathlib.PosixPath,
+    ) -> yaml.ScalarNode:
+        return dumper.represent_scalar("tag:yaml.org,2002:str", str(data))
+
+    def workplanstate_representer(
+        dumper: yaml.Dumper,
+        data: WorkPlanState,
+    ) -> yaml.ScalarNode:
+        return dumper.represent_scalar("tag:yaml.org,2002:str", str(data))
```
ScottEilerman, line 40: Are these representers necessary for builtins like Paths and Enums? I know they would be for custom classes, but I don't feel like I've needed these ones before.

ScottEilerman (approved):

cstar/tests/unit_tests/orchestration/test_models.py
ankona: I am playing around with the serialization in the tests to check behavior of deserialization after removing extra yaml tags that might look confusing to a user. I will refine the "real" serialization code further, but this should be ok for my testing purposes.

cstar/orchestration/models.py
ankona: I'm trying to avoid changing the references here (e.g. `foo = Model(bar=[a,b]); foo.bar = [x,y]  # <-- raises on assignment`. I may swap to a tuple as well as I iterate.

PR #338, opened 2025-09-10
Modify container builds after Perlmutter MPI testing

ScottEilerman:
This PR collects the work I've done trying to get ROMS to run in a container across two nodes on Perlmutter, using podman-hpc. After doing some cleanup, I'm rebuilding and retesting one last time to make sure I didn't break anything, and will leave this in Draft mode until I confirm it works.

There are effectively three build routes one could take:

1. You can hop into the all_w_nersc_libs directory and follow the README instructions to do a minimal build on top of nersc-optimized MPI/HDF5/NETCDF (rather than building those from source). This works, and ran in 1/3 the time of the next option
2. You can run `make all` to do the multistage build, recreating everything from source, using a NERSC-approved installation of MPICH, and run it via instructions in ci/containers/README.md.
3. In several of the Containerfiles, some lines are commented out, corresponding to our attempt to build and use OpenMPI instead of MPICH. I never got OpenMPI to work completely. On one hand, we could parameterize the builds further, rather than leaving a bunch of opaque locations where you'd need to toggle comments. On the other hand, I don't want to put that effort in ( 😅 ), especially for something that doesn't work. Open to commentary / criticism if you think we should clean this up a bit.

When I say "you can run it" for the first two, note that I also did a somewhat hacky thing of putting compile-time .opt files into my `/runner/` directory during the container build, so that I didn't have to re-build at runtime. You'd need to do the same for whatever domain you're trying to run.


ci/containers/buildbase/Containerfile
```diff
@@ -24,12 +25,16 @@ RUN apt-get update && \
         software-properties-common \
 	    wget \
         zlib1g-dev \
-        --no-install-recommends && \
+        libfabric-dev \
```
ankona: let's keep these one-per-line and alphabetized

ci/containers/runner/Containerfile
```diff
@@ -45,9 +53,20 @@ COPY --from=stage_marbl $INSTALL_BASE_DIR/marbl0.45.0/ $INSTALL_BASE_DIR/marbl0.
...
+# RUN bash -c "source /etc/profile &&  \
+#     cd $INSTALL_BASE_DIR/work_internal/compile_time_code  \
+#     && make compile_clean &&  \
+#     make tools-roms &&  \
+#     make"
+
+# add these to end of roms compile if you need debug outputs:
+# BUILD_MODE=debug KEEP_PPSRC=true"
 
-COPY entrypoint.sh /entrypoint.sh
```
ankona, line 49: `run.sh` doesn't ensure the scripts from the profile directory are loaded. Do we want to get rid of `entrypoint.sh`?

ci/containers/all_w_nersc_libs/README.md
```diff
@@ -0,0 +1,31 @@
...
+module load cray-netcdf/4.9.0.9
+module load PrgEnv-gnu/8.5.0
+
+podman-hpc build -v /opt/cray:/opt/cray . -t runner_built_w_nersc_libs:latest
+podman-hpc migrate runner_built_w_nersc_libs:latest
+```
+
+And then at runtime:
+
+```
```
ankona: Do you want to add `commandline` here to ensure rendering consistency?

ci/containers/README.md
```diff
@@ -55,3 +55,40 @@ To retrieve the latest version of all images, run:
...
+
+Two different versions of an MPICH build have succeeded: one that links in NERSC's optimized libraries at build time (see [here](ci/containers/all_w_nersc_libs)), and one that builds our own versions of mpich, hdf5, and netcdf from source (see the Makefile in this directory, culminating in the [runner container image](ci/containers/runner)). Instructions for building and running the NERSC-linked container are in a [separate README](ci/containers/all_w_nersc_libs/README.md).
+
+For the multistage, built-from-source version, you should be able to clone this repo onto Perlmutter, and run `make all` from this directory. In initial tests, simulation-specific compile-time .opt files were added into the runner build directory and built in at container image build-time, so that we didn't need to recompile roms at runtime. 
+
+With a built and migrated runner image, the following command worked at runtime:
+
+```
```
ankona: Do you want to add `commandline` here to ensure rendering consistency?

ankona (changes requested):
Looks great. Main notes are:

- I have a minor change request for organizing/formatting dependencies. 
- The `README.md`'s are fantastic.
- I'm not sure about the removal of the `entrypoint.sh` script on the runner.

ci/containers/runner/Containerfile
ScottEilerman: I can't remember if I intended this, or if it was an accident. I had an additional entrypoint-build script that would rebuild roms and do some other things, that I eventually got rid of. Anyway, I have no objection to adding this back.

ScottEilerman (commented):
Thanks! Addressed all comments.

ankona (approved):
w00t! push it to production.

PR #342, opened 2025-09-27
Implementation of `cstar.io` on `ExternalCodeBase`

dafyddstephenson:
Following #300 , this implements the new `io` methods to the `ExternalCodeBase` class and its subclasses.

`ExternalCodeBase` was always considered a bit of an outlier in that unlike `InputDataset` and `AdditionalCode` an instance is not strictly tied to a `Simulation`. As such, the implementation here allows a `StagedRepository` to be created within `ExternalCodeBase.__init__` by just checking if the root environment variable associated with the codebase is already defined, and seeing if the codebase found at that location is already correctly configured. This led to some complications in getting the existing unit tests to pass.

Also problematic were tests creating a `ROMSSimulation` instance (e.g. as part of a dict or blueprint roundtrip test). The changes implemented in #336 helped a lot with testing existing instances (covering most tests), but the handful of tests that _create_ instances of more complex classes will likely require a bit more work to run cleanly without sprawling lists of patches and mocks. It would be nice to find some time for a second pass at the changes to the unit tests here to tidy them up, group related patches, define custom context managers, etc. Please keep in mind if reviewing!

The rest of the implementation here is I think as planned. A few tweaks were made to `cstar.io`, but nothing major.

### DIFF
--- a/cstar/base/external_codebase.py
+++ b/cstar/base/external_codebase.py
@@ -30,6 +30,8 @@ class ExternalCodeBase(ABC, LoggingMixin):
     TODO : update here
     """
 
+    _working_copy: StagedRepository | None = None  # updated by self.get()
+
     def __init__(
         self, source_repo: str | None = None, checkout_target: str | None = None
     ):
@@ -55,8 +57,15 @@ class ExternalCodeBase(ABC, LoggingMixin):
         if not checkout_target:
             checkout_target = self._default_checkout_target
         self._source = SourceData(location=source_repo, identifier=checkout_target)
-
-        self._working_copy: StagedRepository | None = None  # updated by self.get()
+        if not self.is_configured:
+            self._working_copy = None
+        else:
+            root_dir = str(
+                cstar_sysmgr.environment.environment_variables.get(self.root_env_var)
+            )
+            self._working_copy = StagedRepository(
+                source=self.source, path=Path(root_dir)
+            )
 
     def __str__(self) -> str:
         base_str = f"{self.__class__.__name__}"
@@ -105,9 +114,9 @@ class ExternalCodeBase(ABC, LoggingMixin):
 
     @property
     @abstractmethod
-    def expected_env_var(self) -> str:
-        """Environment variable associated with the external codebase, e.g.
-        MARBL_ROOT.
+    def root_env_var(self) -> str:
+        """Environment variable pointing to the location of the codebase when configured, e.g.
+        MARBL_ROOT
         """
 
     @property
@@ -117,9 +126,8 @@ class ExternalCodeBase(ABC, LoggingMixin):
     def get(self, target_dir: Path | None = None) -> None:
         """Retrieve and stage this ExternalCodeBase"""
         if self.working_copy:
-            raise ValueError(
-                f"ExternalCodeBase is already staged at {self.working_copy.path}. "
-                "Consider ExternalCodeBase.working_copy.reset or ExternalCodeBase.uninstall"
+            self.log.info(
+                f"ExternalCodeBase is already staged at {self.working_copy.path}. Skipping get() call"
             )
 
         if not target_dir:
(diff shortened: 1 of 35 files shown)


cstar/base/external_codebase.py
```diff
@@ -62,257 +41,145 @@ def __init__(
...
         """
-        # check 1: X_ROOT variable is in user's env
-        env_var_exists = (
-            self.expected_env_var
-            in cstar_sysmgr.environment.environment_variables.keys()
-        )
+        if self.working_copy:
+            self.log.info(
+                f"ExternalCodeBase is already staged at {self.working_copy.path}. Skipping get() call"
+            )
```
ScottEilerman, line 142: Should there be a `return` here if it's going to skip?

cstar/base/external_codebase.py
```diff
@@ -62,257 +41,145 @@ def __init__(
...
-                head_hash = _get_repo_head_hash(local_root)
-                head_hash_matches_checkout_hash = head_hash == self.checkout_hash
-                if head_hash_matches_checkout_hash:
-                    return 0
-                else:
-                    return 2
 
-        else:  # env_var_exists False (e.g. ROMS_ROOT not defined)
-            return 3
+        staged_repo = self.source.stage(target_dir=target_dir)
```
ScottEilerman, line 154: you love to see it

cstar/base/external_codebase.py
```diff
@@ -62,257 +41,145 @@ def __init__(
...
+        if not isinstance(staged_repo, StagedRepository):
+            raise ValueError(
+                f"Could not retrieve {self.source.location} as a local repository"
+            )
```
ScottEilerman: Should `self.source.stage` be raising this if it failed? I'll need to go back and look at what we did there...

ScottEilerman: Per discussion, since this is just to satisfy mypy, we can probably switch to an assert for now, or just leave as-is.

ScottEilerman (commented):
Initial set of comments that we looked through. Will take a look at the rest.

cstar/roms/simulation.py
```diff
@@ -1066,7 +1064,10 @@ def setup(
 
         for codebase in filter(lambda x: x is not None, self.codebases):
             self.log.info(f"🔧 Setting up {codebase.__class__.__name__}...")
-            codebase.handle_config_status()
+            # codebase.handle_config_status()
+            codebase.setup()
+            # codebase.get()
+            # codebase.configure()
```
ScottEilerman: A few leftover comments here.

cstar/tests/unit_tests/conftest.py
```diff
@@ -443,14 +628,16 @@ def fake_romssimulation(
...
         name="ROMSTest",
         directory=directory,
         discretization=ROMSDiscretization(time_step=60, n_procs_x=2, n_procs_y=3),
-        codebase=ROMSExternalCodeBase(
-            source_repo="http://my.code/repo.git", checkout_target="dev"
-        ),
+        codebase=fake_romsexternalcodebase,
+        # codebase=ROMSExternalCodeBase(
+        #     source_repo="http://my.code/repo.git", checkout_target="dev"
+        # ),
```
ScottEilerman: more leftover comments here and below

cstar/tests/unit_tests/roms/test_roms_external_codebase.py
```diff
@@ -1,50 +1,63 @@
...
-        1. clones ROMS from `ROMSExternalCodeBase.source_repo`
-        2. checks out the correct commit from `ROMSExternalCodeBase.checkout_target`
-        3. Sets environment variable ROMS_ROOT and appends $ROMS_ROOT/Tools-Roms to PATH
-        4. Compiles the NHMG library
-        5. Compiles the Tools-Roms package
 
-    Tests
-    -----
+class TestROMSExternalCodeBaseConfigure:
+    """TODO Add tests for new .get() and .configure() methods. Previous tests:
```
ScottEilerman, line 60: note TODO here

ScottEilerman (commented):
A few cleanup things in the rest but otherwise ok to go forward whenever you're satisfied!

ScottEilerman (approved):

PR #344, opened 2025-09-30
Add compatibility for Anvil

### LINKED ISSUE #343: Need to set up lmod/env files for Anvil
Although ROMS is ported to Anvil, C-Star is not! I have a PR to address this momentarily.

ScottEilerman:
This PR makes a few modifications to allow C-Star to run on Anvil:

- Adds .env and lmod files based on the work @dafyddstephenson already did during porting
- Adds a check for an RCAC-specific env-var to determine you're on Anvil. Unfortunately they don't use the same LMOD env-vars that the other HPCs do. I opted for a simple check in the namer rather than adding another layer of possible things to check to the overall class -- if someone thinks this is too much of a shortcut let's discuss.
- Added handling for the fact that `sinfo` on Anvil seems to say that the allowed walltime is `infinite` even though their help docs say otherwise. For now I sort of bypass our handling around the check and treat it as an "unknown" walltime which displays a warning message but lets you move forward.
- Somewhat unrelated, but I also added handling for the PENDING slurm job status. During my testing, my C-Star flow broke a bit because I expected my .updates() call to block until the job passed or failed but instead it checked, got PENDING, and returned immediately (then tried to move on to post-run which crashed because the job wasn't done yet).

Overall this was a little quick and dirty but seems to work ok for me. Might need some input from folks who designed/tested these pieces for other HPCs though.

- [x] Closes #343 
- [x] Tests added
- [x] Tests passing


cstar/system/manager.py
```diff
@@ -182,6 +185,36 @@ def create_scheduler(cls) -> Scheduler | None:
         )
 
 
+@register_sys_context
+@dataclass(frozen=True)
+class _AnvilSystemContext(_SystemContext):
+    """The contextual dependencies for the Perlmutter system."""
```
ankona:
```suggestion
    """The contextual dependencies for the Anvil system."""
```

cstar/system/manager.py
```diff
@@ -182,6 +185,36 @@ def create_scheduler(cls) -> Scheduler | None:
...
+    compiler: ClassVar[str] = "gnu"
+    """The compiler used on Anvil."""
+    mpi_prefix: ClassVar[str] = "srun"
+    """The MPI prefix used on Anvil."""
+    docs: ClassVar[str] = "https://www.rcac.purdue.edu/knowledge/anvil/architecture"
+    """URI for documentation of the Anvil system."""
+
+    @classmethod
+    def create_scheduler(cls) -> Scheduler | None:
+        per_regular_q = SlurmPartition(name="wholenode")
```
ankona: naming clarity nitpick - what does `per_` mean in this context?

could these be `regular_q`, `shared_q`, and `debug_q` and not lose information?

cstar/execution/scheduler_job.py
```diff
@@ -256,7 +256,7 @@ def __init__(
                 + f"and C-Star cannot default to the max walltime for the queue {queue_name} "
                 + " as it cannot be determined"
             )
-        elif self.queue.max_walltime is None:
+        elif self.queue.max_walltime is None or self.queue.max_walltime == "infinite":
```
ankona: we have the magic string `"infinite"` both here and the scheduler. Can we shift that to a constant, somewhere?

Is `infinite` globally valid or only on Anvil? 
- If global, we might want to also provide a reference to it where users configure walltime, e.g. `class Queue` might say in the docstring that `infinite` is a valid option.
- If Anvil-only, we might document this valid value in the `AnvilSystemContext`

----

On further review, I see that we're not even accepting this value. Would it be easier (no need to specify valid constants per HPC, like above note) if this avoided checking for the constant, altogether? 

Consider that replacing this conditional with an attempt to parse a walltime string would fail and the same error can be logged, while keeping platform-specific value-parsing out of the `SchedulerJob` class.

```python
def _try_parse_walltime(queue: Queue) -> bool:
    if queue.max_walltime is None:
        self.log.debug("Rejecting null walltime")
        return False

    fmt_regex = "...HH:mm:ss..." # however this is done...
    if not re.match(queue.max_walltime, fmt_regex):
        self.log.debug(f"Rejecting invalid walltime format: {queue.max_walltime}")
        return False
    
    return True
```

If platform specific regex are required, those could be added to the system context.

cstar/system/scheduler.py
```diff
@@ -88,6 +88,8 @@ def _parse_walltime(self, walltime_str) -> str:
         str
             The formatted walltime string in the "HH:MM:SS" format.
         """
+        if walltime_str == "infinite":
```
ankona: Consider avoiding use of anvil-specific `"infinite"` constant here and using walltime parsing to decide when to return the input:

```python
if not {":","-"}.intersection(set(walltime_str)):
    # return the raw string when a non-time-spec string is encountered.
    return walltime_str
```

ankona (changes requested):
Looks pretty good. I have a couple questions and a minor docstring fix for you.

cstar/system/manager.py
ScottEilerman: No clue, I copied them from Perlmutter. Maybe they mean... Perlmutter? I'll rename.

ankona: that makes sense :rofl:

cstar/execution/scheduler_job.py
ScottEilerman: `infinite` is coming from slurm's `sinfo` [which seems to always have a fixed format](https://slurm.schedmd.com/sinfo.html#OPT_TIMELIMIT) of `"days-hours:minutes:seconds"` with the one exception being `infinite`.

Another option I considered was translating `infinite` to some arbitrarily large time limit like `9999-00:00:00` such that it will serve our purpose and let the rest of the code flow. Though, that might actually hide the real concern -- according to Anvil's docs, the time limit isn't _actually_ infinite, they do have limits, they just don't seem to be enforcing them here in slurm I guess? It's unclear. I guess that's why I had it go down the route of our "unknown" code branch and print a warning saying "welp we tried, it's your problem if your job times out."

Ok all that said, I'm open to whatever here. `infinite` does seem to be the only magic value so maybe making a constant for it is fine. Regex is fine too, though then I feel obligated to make a bunch of unit tests to make sure my regex is right. Or we could just do `strptime` and use that syntax.

ankona: Since it's a SLURM-provided value, it makes sense that it would be evaluated here. Nothing to see.

... except to nitpick that the first if block raises an exception, so the second clause (`elif self.queue_max_walltime is None or ...`) should just be a new `if`

cstar/system/scheduler.py
ankona: per convo above, this could be pushed to a classvar/constant on the `Queue` to simplify comparisons. Throwing in the property to let the Queue class own answering "is this the maximum."

```python
class Queue:
  MAX_WALLTIME_CONSTANT: t.ClassVar[str] = "infinite"
  ...
  def _parse_walltime(...):
    if walltime_str == Queue.MAX_WALLTIME_CONSTANT:
      return walltime_str

  @property
  def allow_maximum_walltime(self):
    return self.actual_walltime == Queue.MAX_WALLTIME_CONSTANT

class JobScheduler:
  ...
  def __init__(...):
    ...
    elif not self.queue.max_walltime or self.queue.allow_maximum_walltime:
      ...
```

cstar/execution/scheduler_job.py
ScottEilerman: Disregard this thread -- I discovered the max walltime command for our QOS class actually works here (using sacctmgr instead of sinfo) and returns the correct value instead of infinite. but, other parts of the process treat these as partitions. I'm refactoring now to separate the method for max walltime to be settable instead of tied to queue/partition.

ScottEilerman: Ok, updated. I'm not super pleased with how I did it, but it works. I wanted to give the option to set a different walltime query method while defaulting to the original behavior.

cstar/system/scheduler.py
ScottEilerman: now irrelevant, no more `inifinites` 🙃

cstar/execution/scheduler_job.py
ankona: w00t! great catch.

cstar/system/scheduler.py
ankona: Am I officially Mr. Irrelevant, now?

ScottEilerman: hey if you hadn't poked at it I wouldn't have figured out that there was a second way to query the walltime!

ankona (approved):
LGTM

PR #348, opened 2025-10-02
Fix unit test fixture

ScottEilerman:
I kept getting this fake_romsruntimesettings.in file popping up in my fixtures directory when running tests. I thought it must not be getting deleted after the tests finish, but actually it looks like a possible find/replace error in some of the test refactoring from a while back. I believe this commit gets everything back in order.

- [x] Tests passing


dafyddstephenson (approved):
I may have done a very lazy replace here by the looks 😅

PR #349, opened 2025-10-02
Patch fixture for ROMSSimulation.__init__

dafyddstephenson:
This PR creates a new fixture `patch_romssimulation_init_sourcedata` that returns a context manager patching any `SourceData` instances necessary to create a `ROMSSimulation` instance inside a test, for example in `test_from_blueprint_url`. 

This already drops 132 lines of repeated `mock.patch` calls in `test_roms_simulation.py` that were made necessary by the `io` implementation on `ExternalCodeBase` and would continue sprawling as `io` reaches more parts of the code.

It also tidies up some remaining missing typehints.

### DIFF
--- a/cstar/tests/unit_tests/roms/conftest.py
+++ b/cstar/tests/unit_tests/roms/conftest.py
@@ -1,0 +1,434 @@
+from collections.abc import Generator
+from pathlib import Path
+from unittest import mock
+
+import numpy as np
+import pytest
+
+from cstar.base import AdditionalCode
+from cstar.base.datasource import DataSource
+from cstar.roms import ROMSDiscretization, ROMSExternalCodeBase, ROMSSimulation
+from cstar.roms.input_dataset import (
+    ROMSBoundaryForcing,
+    ROMSForcingCorrections,
+    ROMSInitialConditions,
+    ROMSModelGrid,
+    ROMSRiverForcing,
+    ROMSSurfaceForcing,
+    ROMSTidalForcing,
+)
+from cstar.roms.runtime_settings import ROMSRuntimeSettings
+from cstar.tests.unit_tests.fake_abc_subclasses import (
+    FakeROMSInputDataset,
+)
+
+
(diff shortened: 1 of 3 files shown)


cstar/tests/unit_tests/roms/conftest.py
```diff
@@ -446,3 +447,42 @@ def fake_romssimulation_dict_no_forcing_lists(
...
+            ),
+            mock.patch(
+                "cstar.roms.simulation.ROMSExternalCodeBase.is_configured",
+                new_callable=mock.PropertyMock,
+                return_value=False,
+            ),
+        ):
+            yield
+
+    return _context
```
ScottEilerman, line 488: This is fine as-is with how you're calling it in the tests, but FYI you can also drop the def/return and just have the fixture yield out of the context manager like this:
```python

    with (
            mock.patch(
                "cstar.base.external_codebase.SourceData",
                side_effect=[
                    mock_externalcodebase_sourcedata,
                    mock_marbl_externalcodebase_sourcedata,
                ],
            ),
            mock.patch(
                "cstar.roms.simulation.ROMSExternalCodeBase.is_configured",
                new_callable=mock.PropertyMock,
                return_value=False,
            ),
        ):
            yield

```

and then simply passing the fixture will have the entire test execute within the patch context, without needing to call it explicitly in the test. That's the more common way I see mock/patch fixtures done, unless you really need parts of the tests to _not_ be in the mock context.

ScottEilerman (approved):
one suggestion/alternative but ok as-is if you prefer!

cstar/tests/unit_tests/roms/conftest.py
dafyddstephenson: Right, this was my first approach and how I'm patching things elsewhere, but I don't think I actually want it to run for the entire test. It's working with two `ROMSSimulation` instances, so the patch interferes with the first one, when I just want to patch the creation of the second (one line)

PR #340, opened 2025-09-24
CSD-39: add initial models for serialization of workplans

ankona:
Add initial models for workplan orchestration

- [x] Closes #CSD-39
- [x] Closes #CSD-140
- [x] Closes #CSD-136
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage

PR #350, opened 2025-10-06
merge main to develop and fix conflicts

ScottEilerman:
Takes @dafyddstephenson 's branch from #347, adds on the unit test fix (#348), and then fixes _more_ merge conflicts with the branch we just merged into develop (#340). Re-adds some cdr pieces to the new fixtures from main. Finally, reverted a `return` to a `yield` because mypy didn't like it.


cstar/tests/unit_tests/roms/conftest.py
```diff
@@ -266,6 +267,12 @@ def fake_tidal_forcing() -> ROMSTidalForcing:
     return ROMSTidalForcing(location="http://my.files/tidal.nc", file_hash="345")
 
 
+@pytest.fixture
+def fake_cdr_forcing() -> ROMSCdrForcing:
+    """Provides a ROMSTidalForcing instance with fake attrs for testing"""
```
dafyddstephenson: docstring

cstar/tests/unit_tests/roms/conftest.py
```diff
@@ -43,7 +43,7 @@ def fake_romsexternalcodebase() -> Generator[ROMSExternalCodeBase, None, None]:
 # ROMSRuntimeSettings
 ################################################################################
 @pytest.fixture
-def fake_romsruntimesettings() -> ROMSRuntimeSettings:
```
dafyddstephenson: The thing that "fixed" mypy was the removal of this typehint - when you add a new typehint (`Generator[ROMSRuntimeSettings,None,None]`) to match the `yield` you get the same list of `arg-type` errors. Would suggest keeping it as a `return` (as no complex cleanup is needed) and ... well idk what I'd suggest for the errors tbh 😂

dafyddstephenson (changes requested):

cstar/tests/unit_tests/roms/conftest.py
dafyddstephenson: `mypy` passes on my original branch but it looks like `roms/runtime_settings.py` got shaken up in eb5f4a32ebab6eb0abeb1ef06abed77d2b2fbc27

dafyddstephenson: looks like the fixture is duplicated between `unit_tests/conftest.py` and `unit_tests/roms/conftest.py` also? (deleting the first one doesn't fix this tho)

ScottEilerman: Good catch on the duplicate. Deleted that. I reverted the `yield` to `return` and just left off the typehint for now. I think this is related to the thing we did in ROMSRuntimeSettings with pydantic to let you init it from primitives instead of the pydantic type. I think Chris and I found a better way to allow that when doing the pydantic models for blueprints, but I don't want to try to adopt that here right now just to make mypy happy if the code works fine. I think that class will get naturally cleaned up once we eventually move to a runtime namelist from roms anyway.

PR #351, opened 2025-10-08
Support new blueprints in worker and tests

ScottEilerman:
This PR looks big but a the diffs are a bit inflated by blueprint yamls and some indent-level changes in tests.

The primary purpose of this PR is to enable the worker to use the new blueprint schema, and to bypass batch script generation/submission on HPCs if RomsSimulation is being orchestrated by a separate process. The latter was relatively easily achieved by using a new convention of setting a `CSTAR_ORCHESTRATED` env-var (name open to discussion).

The PR spiraled a bit as I began updating tests to match the worker changes for the new blueprint. Originally, the worker received the runtime params (start date, end date, and output_dir) via command line parsing, but the newer design incorporates these into the blueprint itself. I then began updating fixtures and realized all of the unit tests and integration tests were still using old-style blueprints, and now was as good a time as any to swap them out and cut over the codebase to the new blueprints for real. In addition to a lot of test refactoring, that led to a bit of code refactor to move the primary blueprint adapter into the `RomsSimulation.from_blueprint()` (partly to maintain some existing functionality/API, and partly to avoid circular imports by reaching back to `adapter.py` from `simulation.py`).

As I was removing old blueprints, I took the opportunity to remove functionality related to running ROMS-tools yamls to generate inputs on the fly from the RomsSimulation object. We intend to later support similar automation of ROMS-tools input generation via its own blueprint and "application" framework.

I temporarily disabled (via a `NotImplementedError`) `RomsSimulation.to_blueprint()`. In the current paradigm, I don't envision anyone constructing `RomsSimulation` objects as code and then saving them out, it is intended to mainly be a one-way yaml -> code process. I could see potentially wanting to save out `RomsSimulation` objects back to blueprints as part of an audit trail process, but re-implementing the feature at this moment isn't a priority.

I think we can also get rid of `from_dict` and `to_dict`, as their primary purpose was to facilitate to/from blueprint methods, but I stopped short of removing those for the time being, primarily because they are "load bearing" in a lot of the tests, and I know Dafydd's ongoing refactor touches those same tests, so if we do think we can nix them, it'll be cleaner after that refactor merges into develop.

I additionally did a bit of minor cleanup relating to things that were bugging me or knock-on effects of other changes:
- removed test patches related to the user's `.cstar.env` path in favor of a global auto-use patch. A few tests that needed that patch didn't have it and it kept wrecking my local environment. The new patch guarantees that we never touch a user's real `.cstar.env` while running tests.
- changed a function signature on log.register_file_handler to utilize module defaults; changed where we set the log file in the worker flow, since the output directory isn't easily readable at the same place it used to be
- I removed functionality in `RomsSimulation` that makes sure it is running in an empty directory. In my initial POC flow, I create the working directory and drop the sbatch script in it right from the get-go. This functionality also exists in `worker.py` -- that functionality was changed slightly to check the "unique" directory instead of the parent. I think this general category of check (make sure user isn't overwriting something or re-entering an existing run in a weird way) check will either become unnecessary or change shape quite a bit in a fully orchestrated setting, so the main focus was clearing hurdles for now and we can revisit this check when the time is right.


dafyddstephenson:
Just a note to see please let me review before merging!

ScottEilerman:
Note that there's a parallel PR #355 that goes from this branch to a feature branch, and comments that started there are also resulting in changes here.

ScottEilerman:
Closing this, it has been merged into `orchestration` already, and develop is getting merged into orchestration, so moving this into develop will just complicate things further when we eventually bring orchestration back into develop.

@dafyddstephenson this may still be the cleanest place at the moment to look at some of these changes, separate from the rest of the orchestration work. However, some pieces of this are already being modified/updated to line up with the IO refactor, so we may just want to walk through some of it together after travels and conferences.

PR #353, opened 2025-10-11
`cstar.io` implementation for `AdditionalCode` 

dafyddstephenson:
home stretch!

### DIFF
--- a/cstar/base/additional_code.py
+++ b/cstar/base/additional_code.py
@@ -1,11 +1,10 @@
-import shutil
-import tempfile
 from pathlib import Path
 
-from cstar.base.datasource import DataSource
-from cstar.base.gitutils import _clone_and_checkout
+from cstar.base.gitutils import git_location_to_raw
 from cstar.base.log import LoggingMixin
-from cstar.base.utils import _get_sha256_hash, _list_to_concise_str
+from cstar.io.constants import SourceClassification
+from cstar.io.source_data import SourceDataCollection, _SourceInspector
+from cstar.io.staged_data import StagedDataCollection
 
 
 class AdditionalCode(LoggingMixin):
@@ -47,8 +46,8 @@ class AdditionalCode(LoggingMixin):
         self,
         location: str,
         subdir: str = "",
-        checkout_target: str | None = None,
-        files: list[str] | None = None,
+        checkout_target: str = "",
+        files: list[str] = [],
     ):
         """Initialize an AdditionalCode object from a DataSource  and a list of code
         files.
@@ -71,23 +70,29 @@ class AdditionalCode(LoggingMixin):
         AdditionalCode
             An initialized AdditionalCode object
         """
-        self.source: DataSource = DataSource(location)
-        self.subdir: str = subdir
-        self._checkout_target = checkout_target
-        self.files: list[str] | None = [] if files is None else files
+        if (
+            _SourceInspector(location).classify()
+            == SourceClassification.REMOTE_REPOSITORY
+        ):
+            source = SourceDataCollection.from_locations(
+                locations=[
+                    git_location_to_raw(location, checkout_target, f, subdir)
+                    for f in files
+                ]
+            )
+        else:
+            source = SourceDataCollection.from_locations(
+                locations=[f"{location}/{subdir}/{f}" for f in files]
+            )
+        self.source: SourceDataCollection = source
         # Initialize object state
-        self.working_path: Path | None = None
-        self._local_file_hash_cache: dict = {}
+        self.working_copy: StagedDataCollection | None = None
 
     def __str__(self) -> str:
         base_str = self.__class__.__name__ + "\n"
         base_str += "-" * (len(base_str) - 1)
-        base_str += f"\nLocation: {self.source.location}"
-        if self.subdir is not None:
-            base_str += f"\nSubdirectory: {self.subdir}"
-        if self.checkout_target is not None:
-            base_str += f"\nCheckout target: {self.checkout_target}"
-        base_str += f"\nWorking path: {self.working_path}"
+        base_str += f"\nLocations: {self.source.locations}"
+        base_str += f"\nWorking copy: {self.working_copy}"
         base_str += f"\nExists locally: {self.exists_locally}"
         if not self.exists_locally:
             base_str += " (get with AdditionalCode.get())"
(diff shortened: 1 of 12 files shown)


cstar/base/additional_code.py
```diff
@@ -61,159 +48,91 @@ def __init__(
...
+            source = SourceDataCollection.from_locations(
+                locations=[
+                    git_location_to_raw(location, checkout_target, f, subdir)
+                    for f in files
+                ]
+            )
+        else:
+            source = SourceDataCollection.from_locations(
+                locations=[f"{location}/{subdir}/{f}" for f in files]
+            )
```
ScottEilerman, line 73: This if/else branch "feels" like it should be handled by `SourceDataCollection` itself. Importing the private `_SourceInspector` is one reason why. I see the problem though with needed to formulate the locations based on git repos.

One solution is to add parameters to `from_locations` and just pass everything into it and let it sort it out. I think that's how I'd lean at the moment, in lieu of any better ideas. You could potentially have a separate helper function `resolve_location(location, subdir, checkout_target)` but that will still need to inspect/classify the source and should probably live in `source_data.py` as well.

cstar/io/source_data.py
```diff
@@ -115,7 +115,7 @@ def source_type(self) -> SourceType:
             if self._is_repository:
                 self._source_type = SourceType.REPOSITORY
             elif self.location_type is LocationType.HTTP:
-                if (not self._http_is_html) and (self.suffix):
+                if not self._http_is_html:  # and (self.suffix):
```
ScottEilerman, line 118: maybe already in your cleanup pass, but noting it here in case you missed it

cstar/io/source_data.py
```diff
@@ -315,7 +315,7 @@ def _validate(self):
                 SourceType.REPOSITORY,
             ]:
                 raise TypeError(
-                    f"Cannot create SourceDataCollection with data of source type '{s.source_type.value}'"
+                    f"Cannot create SourceDataCollection with data of source type '{s._classification.value.source_type.value}'"
```
ScottEilerman, line 318: I was about to say, maybe we should surface this as a property, but you already have a #TODO in source_data to do that. 👍

cstar/roms/simulation.py
```diff
@@ -707,20 +709,23 @@ def roms_runtime_settings(self) -> ROMSRuntimeSettings:
...
-            f in self.runtime_code.files
+            f in runtime_code_filenames
             for f in [
                 "marbl_in",
                 "marbl_tracer_output_list",
                 "marbl_diagnostic_output_list",
             ]
         ):
+            # TODO [0] below isn't the best
+            runtime_code_path = self.runtime_code.working_copy[0].path.parent
```
ScottEilerman, line 722: I guess you could lightly refactor this section to identify that there is exactly one path in `self.runtime_code.source` for each of the three file names, then pass those full paths into the constructor below. Probably going to be the same outcome in 99.9% of cases, because why would you have these in different directories...

cstar/roms/simulation.py
```diff
@@ -1085,14 +1090,18 @@ def tree(self):
...
 
         if hasattr(self, "compile_time_code") and (self.compile_time_code is not None):
             simulation_tree_dict.setdefault("compile_time_code", {})
             simulation_tree_dict["compile_time_code"] = [
-                compile_time_code.split("/")[-1]
-                for compile_time_code in self.compile_time_code.files
+                codefile.basename
+                for codefile in self.compile_time_code.source
+                # compile_time_code.split("/")[-1]
+                # for compile_time_code in self.compile_time_code.files
```
ScottEilerman, line 1104: as before, noting for cleanup pass, though you're probably already on it

cstar/simulation.py
```diff
@@ -450,34 +450,12 @@ def to_dict(self) -> dict:
...
-            runtime_code_info["location"] = runtime_code.source.location
-            if runtime_code.subdir is not None:
-                runtime_code_info["subdir"] = runtime_code.subdir
-            if runtime_code.checkout_target is not None:
-                runtime_code_info["checkout_target"] = runtime_code.checkout_target
-            if runtime_code.files is not None:
-                runtime_code_info["files"] = runtime_code.files
-
-            simulation_dict["runtime_code"] = runtime_code_info
+        if hasattr(self, "runtime_code") and self.runtime_code is not None:
```
ScottEilerman, line 453: nit, here and below, you could reduce to:
```if getattr(self, "runtime_code", None) is not None:```

In addition to being shorter, it'll chop a few microseconds off of our weeks of runtime.

**Review by ScottEilerman — APPROVED** [review]
A few suggestions, but overall looking good! Preemptively approving, resolve as you see fit.

File: cstar/base/additional_code.py (hunk 1 above)
Comment (dafyddstephenson) [reply cstar/base/additional_code.py:73]: I'm a couple branches down from this currently so I've added this PR comment with a link in a `TODO` on my current branch and will address it after the next PR. 

I considered having a `SourceDataCollection.from_repository` but it felt like I was kinda crowbarring for a niche outcome. The issue I think is that the conversion of `subdir`, `files` etc. into `locations` feels like it has to happen on the `AdditionalCode` side. I considered making `git_location_to_raw` more broad so it would accept the path construction as well but that felt like more crowbarring. We can discuss! I feel like whatever solution will still be in the neighbourhood of what is here, so no stress.

File: cstar/roms/simulation.py (hunk 4 above)
Comment (dafyddstephenson) [reply cstar/roms/simulation.py:722]: I'd already mentally solved this one I was just trying to hammer this through as it works and come back to it later - I was going to have a `StagedDataCollection.common_parent` property that takes all the files in the collection and finds their nearest shared parent directory. In the case of `AdditionalCode` they're all in the same dir, so applies here too, but there are other places in the code that prop would be useful

File: cstar/simulation.py (hunk 6 above)
Comment (dafyddstephenson) [reply cstar/simulation.py:453]: yeah this is old code but shortening it reminded me that this was the only voodoo that satisfied `mypy`. Weird janky stuff like this is usually a sacrificial offering to pre-commit.

File: cstar/base/additional_code.py (hunk 1 above)
Comment (ScottEilerman) [reply cstar/base/additional_code.py:73]:
```
SourceDataCollection.from_locations(locations=[resolve_location(root? parent? location?, subdir, checkout_target, filename) for filename in files])
```

PR #355, opened 2025-10-20
Support new blueprints in worker and tests (feature branch)

ScottEilerman:
See description in https://github.com/CWorthy-ocean/C-Star/pull/351


cstar/base/log.py
```diff
@@ -8,9 +8,9 @@
 
 def register_file_handler(
     logger: logging.Logger,
-    level: int,
-    fmt: str,
     filename: str | Path | None,
+    level: int = DEFAULT_LOG_LEVEL,
```
ankona, line 12: good catch on using the defaults

cstar/entrypoint/worker/worker.py
```diff
@@ -116,14 +110,19 @@ def __init__(
...
-            start_date=request.start_date,
-            end_date=request.end_date,
+            self._blueprint_uri
+        )
+        log_file = (
+            self._simulation.directory
+            / LOGS_DIRECTORY
+            / WORKER_LOG_FILE_TPL.format(datetime.now(timezone.utc))
         )
+        register_file_handler(self.log, log_file, level=service_cfg.log_level)
```
ankona: The request probably shouldn't perform this logging configuration here. I think it might make more sense to have a flag in the request, e.g. log_to_file: bool, that the worker handling the request would use to execute the desired log-file configuration steps.

This way may also result in a bug where the file handler is registered for the code that creates the request, but not the code handling it.

cstar/entrypoint/worker/worker.py
```diff
@@ -116,14 +110,19 @@ def __init__(
...
+        self.log.setLevel(service_cfg.log_level)
+        self._output_root = self._simulation.directory.expanduser()
+        self._output_dir = self._get_unique_path(self._output_root)
```
ankona, line 119: Consider making `output_dir` a property (or renaming `_get_unique_path`)

cstar/entrypoint/worker/worker.py
```diff
@@ -116,14 +110,19 @@ def __init__(
...
+            self._blueprint_uri
+        )
+        log_file = (
+            self._simulation.directory
+            / LOGS_DIRECTORY
+            / WORKER_LOG_FILE_TPL.format(datetime.now(timezone.utc))
         )
+        register_file_handler(self.log, log_file, level=service_cfg.log_level)
+        self.log.setLevel(service_cfg.log_level)
+        self._output_root = self._simulation.directory.expanduser()
```
ankona, line 118: It's best practice to declare your member variables at the class level. If you want to keep/rename things, you could have updated the removed `output_dir: pathlib.Path` line instead of removing it (and adding `_output_root: pathlib.Path`).

cstar/entrypoint/worker/worker.py
```diff
@@ -165,7 +164,7 @@ def _prepare_file_system(self) -> None:
             None,
         )
 
-        if self._output_root.exists() and outputs:
+        if self._output_dir.exists() and outputs:
```
ankona, line 161: I am convinced this method shouldn't be on this request and should be done by the code handling it. It's not your problem, but maybe we should create a follow-up ticket to just shift it, later.

cstar/orchestration/models.py
```diff
@@ -358,6 +367,11 @@ def _model_validator(self) -> "RomsMarblBlueprint":
 
         return self
 
+    @property
+    def cpus_needed(self) -> int:
+        """Number of CPUS needed for ROMS is derived from the partitioning parameters."""
```
ankona: super nit: don't capitalize `S` or it becomes part of the acronym.

This docstring feels like it could use a minor tweak. As-is, we state an interesting fact but it doesn't seem right. Consider just dropping "is" or using an extended description:

`Number of CPUs needed for ROMS (derived from partitioning parameters).`

or

```python
"""
Number of CPUs needed for ROMS.

The count is computed from partitioning parameters.
"""
```

cstar/roms/simulation.py
```diff
@@ -979,10 +1003,9 @@ def to_blueprint(self, filename: str | Path) -> None:
         from_blueprint : Creates a `ROMSSimulation` instance from a YAML blueprint.
         to_dict : Converts the instance into a dictionary representation.
         """
-        with open(filename, "w") as yaml_file:
-            yaml.dump(
-                self.to_dict(), yaml_file, default_flow_style=False, sort_keys=False
-            )
+        raise NotImplementedError(
```
ankona: Consider updating the docstring, too. Forcing the user to run the code and let it break feels funny. If they see `THIS METHOD HAS BEEN DEPRECATED` in the docstring, they may avoid it.

Python 3.13 introduces `typing.deprecated`, so someday we'll be able to solve more elegantly.

cstar/roms/simulation.py
```diff
@@ -1408,7 +1431,12 @@ def run(
             f"{final_runtime_settings_file}"
         )
 
-        if cstar_sysmgr.scheduler is not None:
+        # If this simulation is already being orchestrated or there is no scheduler,
```
ankona: Consider letting the scheduler tell the blueprint that it's active, instead.

If we add something like this:

```python
# cstar.system.scheduler.py

class Queue:
  ...
  @property
  def is_subtask(self) -> bool:
     raise NotImplemented

...

class SlurmQueue:
  
  ...
  @property
  def is_subtask(self) -> bool:
     return os.getenv("SLURM_JOB_ID")
```

your client here doesn't need to know specifics and we might be able to avoid the extra ENV var `CSTAR_ORCHESTRATED`

```python
if cstar_sysmgr.scheduler and cstar_sysmgr.scheduler.is_subtask:
   ...
```

cstar/tests/integration_tests/blueprints/fixtures.py
```diff
@@ -5,7 +5,9 @@
 
 
 @pytest.fixture
-def modify_template_blueprint(tmpdir) -> Callable[[Path | str, dict[str, str]], Path]:
+def modify_template_blueprint(
+    tmpdir,
```
ankona: consider adding your typehint
```suggestion
    tmpdir: pathlib.Path,
```

cstar/tests/integration_tests/test_fixtures.py
```diff
@@ -18,7 +15,7 @@ def test_mock_input_fixture(mock_user_input):
         assert input("Enter your choice: ") == "yes"
 
 
-def test_modify_template_blueprint(modify_template_blueprint, tmpdir):
+def test_modify_template_blueprint(modify_template_blueprint, tmp_path):
```
ankona: nit: typehints

cstar/tests/unit_tests/conftest.py
```diff
@@ -290,23 +307,11 @@ def system_dotenv_dir(tmp_path: Path) -> Path:
 
 
 @pytest.fixture
-def mock_system_name() -> str:
+def mock_system_name(scope="session") -> str:
```
ankona: Is `scope="session"` meant to be here or in `@pytest.fixture(scope="session")`?

cstar/tests/unit_tests/roms/test_roms_simulation.py
```diff
@@ -1131,15 +1131,7 @@ def test_to_blueprint(self, fake_romssimulation):
...
-    @patch("pathlib.Path.exists", return_value=True)
-    @patch(
-        "builtins.open",
-        new_callable=mock_open,
-        read_data="name: TestROMS\ndiscretization:\n  time_step: 60",
-    )
-    def test_from_blueprint_valid_file(
-        self, mock_open_file, mock_path_exists, tmp_path, fake_romssimulation_dict
-    ):
+    def test_from_blueprint_valid_file(self, blueprint_path):
```
ankona: nit
```suggestion
    def test_from_blueprint_valid_file(self, blueprint_path: Path) -> None:
```

cstar/tests/unit_tests/roms/test_roms_simulation.py
```diff
@@ -1206,15 +1184,17 @@ def test_from_blueprint_invalid_filetype(
...
                 )
 
     @patch("requests.get")
     @patch("pathlib.Path.exists", return_value=True)
     def test_from_blueprint_url(
-        self, mock_path_exists, mock_requests_get, tmp_path, fake_romssimulation_dict
+        self,
+        mock_path_exists,
+        mock_requests_get,
+        tmp_path,
```
ankona: nit: typehints

pyproject.toml
```diff
@@ -7,7 +7,9 @@ authors = [
     { name = "Dafydd Stephenson", email = "dafydd@cworthy.org" },
     { name = "Thomas Nicholas", email = "tom@cworthy.org" },
     { name = "Nora Loose", email = "nora@cworthy.org" },
-    { name = "Matt Long", email = "matt@cworthy.org" }
+    { name = "Matt Long", email = "matt@cworthy.org" },
+    { name = "Scott Eilerman", email = "scott.eilerman@cworthy.org" },
+    { name = "Chris McBride", email = "christopher.mcbride@cworthy.org" },
```
ankona: i am `chris@cworthy.org`!

ankona (commented):
looks pretty good. there are a couple places where things _might_ be better located, but nothing that I'm hard-set against.

cstar/entrypoint/worker/worker.py
ScottEilerman: I think I'm just going to drop the file handler for the moment, and we can reconsider where to add it when things settle. In practice, we already have a log file in the form of the slurm.out dump of stdout/sterr. I think there's some additional log adjustments we'll want to do too in terms of timestamps, format, etc.

cstar/entrypoint/worker/worker.py
ScottEilerman: `_output_root` and `_output_dir` were already defined in the `__init__` (which I assumed was ok since they're private), I just moved them down and changed how they get set. I think I'm going to punt in general on the output directory handling here, because it's going to need some refactoring anyway when we introduce the workplan-derived path structures in the next week or two.

cstar/entrypoint/worker/worker.py
ScottEilerman: Per above, let's clean this up when we redo the path assignment.

cstar/entrypoint/worker/worker.py
ScottEilerman: Agreed, we have this sort of handling in a few places and it should get consolidated somewhere.

cstar/orchestration/models.py
ScottEilerman: done! I got the S right on the superclass, at least.

cstar/roms/simulation.py
ScottEilerman: good idea! somehow even after all that container MPI business where I was fussing over these slurm env-vars for weeks, I completely forgot they existed and could be used here.

cstar/tests/unit_tests/conftest.py
ScottEilerman: lol. oops.

doesn't even really matter here. thought i'd shave off 0.2ns by not redeclaring a string every test.

pyproject.toml
ScottEilerman: i'm the only one that doesn't have a single first name email. :(

cstar/roms/simulation.py
ScottEilerman: per slack convo, just going to delete it for now. we may want it again someday, and this was a sort of reminder, but we'll know if we want or need it without this.

ScottEilerman (commented):
I think I addressed most of them except the output_dir stuff (see comments).

Might go back and add more tests around the active allocation stuff, we are only testing the "old" path at the moment. Though we could also handle that as a future PR to not hold up this one.

cstar/entrypoint/worker/worker.py
ankona: 👍

cstar/entrypoint/worker/worker.py
ankona: it looks like there was a pre-existing bug in failing to set the value of `self.output_dir` instead of creating a new value.

or...

i meant to keep the original value around _somewhere_ but forgot to declare my private member vars

either way, you can solve this minor issue by adding:

```python
    ...
    stages: tuple[...]
    """..."""
    
    _output_root: Path
    """An expanded version of the user-supplied output root."""
    _output_dir: Path
    """A unique output directory name used to avoid collisions on repeated runs."""
```

cstar/entrypoint/worker/worker.py
ankona: 👍

cstar/entrypoint/worker/worker.py
ankona: per offline convo, 👍

cstar/entrypoint/worker/worker.py
ScottEilerman: https://cworthy.atlassian.net/browse/CSD-316

cstar/entrypoint/worker/worker.py
ScottEilerman: https://cworthy.atlassian.net/browse/CSD-316

cstar/orchestration/models.py
ankona: w00t! most important issue to resolve, right here!

cstar/roms/simulation.py
ankona: 👍

ankona (approved):
Let's get this into the integration branch!

PR #358, opened 2025-10-20
Workplan POC

ScottEilerman:
The main body of this code in dag_runner.py is POC only. Dunno if you wanted to merge this for now and ditch it later, or I can take it out of the PR and I'll shamefully pass it to you at a dead drop like the illicit material it is.

Other notable changes to enable that workflow:
* raise errors from worker so that exit codes and slurm dependencies work closer to correct (see note in deficiencies section)
* split slurm job status check to a function to make it more easily callable from "outside" the simulation structure (e.g. we don't persist the SlurmJob object, we just need the id)
* add judicious mkdir calls
* add slurm job dependencies
* do some hacky things to get around errors related to external codebase directories 

Some known deficiencies (_not_ comprehensive, just a few I wrote down or know about):
* the check job task doesn't fail if the job itself failed
* if a job fails, the slurm jobs that depend on it end up in an eternal PENDING state, not a failed state. they also don't show up in sacct while pending, but you can find them with `jobinfo` (may be anvil only?) or `squeue`. `jobinfo` does give some info that could be used to fail them. though fixing the above bullet would negate the check anyway.
* no tests, docstrings, etc.
* not only that, I broke a bunch of unit tests and never even ran them to see that till now, because I was only focused on whether my example case ran or not.
* as we discussed, the main loop and traversal structure is probably a broken POC (piece of crap, not proof of concept)

### DIFF
--- a/cstar/entrypoint/worker/worker.py
+++ b/cstar/entrypoint/worker/worker.py
@@ -163,12 +163,12 @@ class SimulationRunner(Service):
             raise ValueError(msg)
 
         # leftover external code folder causes non-empty repo errors; remove.
-        externals_path = cstar_sysmgr.environment.package_root / "externals"
-        if externals_path.exists():
-            msg = f"Removing existing externals dir: {externals_path}"
-            self.log.debug(msg)
-            shutil.rmtree(externals_path)
-        externals_path.mkdir(parents=True, exist_ok=False)
+        # externals_path = cstar_sysmgr.environment.package_root / "externals"
+        # if externals_path.exists():
+        #     msg = f"Removing existing externals dir: {externals_path}"
+        #     self.log.debug(msg)
+        #     shutil.rmtree(externals_path)
+        # externals_path.mkdir(parents=True, exist_ok=False)
 
         # create a clean location to write outputs.
         if not self._output_dir.exists():
--- a/cstar/entrypoint/worker/worker.py
+++ b/cstar/entrypoint/worker/worker.py
@@ -5,7 +5,6 @@ import enum
 import logging
 import os
 import pathlib
-import shutil
 import sys
 from datetime import datetime, timezone
 from typing import Final, override
@@ -15,7 +14,6 @@ from cstar.base.log import get_logger
 from cstar.entrypoint.service import Service, ServiceConfiguration
 from cstar.execution.handler import ExecutionHandler, ExecutionStatus
 from cstar.roms import ROMSSimulation
-from cstar.system.manager import cstar_sysmgr
 
 DATE_FORMAT: Final[str] = "%Y-%m-%d %H:%M:%S"
 WORKER_LOG_FILE_TPL: Final[str] = "cstar-worker.{0}.log"
@@ -188,9 +186,13 @@ class SimulationRunner(Service):
         if disposition == ExecutionStatus.COMPLETED:
             self.log.info("Simulation completed successfully.")
         elif disposition == ExecutionStatus.FAILED:
-            self.log.error("Simulation failed.")
+            msg = "Simulation failed."
+            self.log.error(msg)
+            raise CstarError(msg)
         else:
-            self.log.warning(f"Simulation ended with status: {disposition}")
+            msg = f"Simulation ended with status: {disposition}."
+            self.log.warning(msg)
+            raise CstarError(msg)
 
     @override
     def _on_start(self) -> None:
--- a/cstar/entrypoint/worker/worker.py
+++ b/cstar/entrypoint/worker/worker.py
@@ -160,6 +160,12 @@ class SimulationRunner(Service):
             msg = f"Output directory {self._output_root} is not empty."
             raise ValueError(msg)
 
+        # this kept tripping up my runs because it is checking the "default" path
+        # that is derived from package_root, rather than the path set by the user.
+        # probably we should just remove this and handle it elsewhere (as noted in
+        # previous PR, this whole method maybe belongs elsewhere), but for the moment,
+        # it's commented out til we think on it.
+
         # leftover external code folder causes non-empty repo errors; remove.
         # externals_path = cstar_sysmgr.environment.package_root / "externals"
         # if externals_path.exists():
--- a/cstar/entrypoint/worker/worker.py
+++ b/cstar/entrypoint/worker/worker.py
@@ -193,11 +193,9 @@ class SimulationRunner(Service):
             self.log.info("Simulation completed successfully.")
         elif disposition == ExecutionStatus.FAILED:
             msg = "Simulation failed."
-            self.log.error(msg)
             raise CstarError(msg)
         else:
             msg = f"Simulation ended with status: {disposition}."
-            self.log.warning(msg)
             raise CstarError(msg)
 
     @override
--- a/cstar/execution/scheduler_job.py
+++ b/cstar/execution/scheduler_job.py
@@ -668,6 +668,7 @@ class SlurmJob(SchedulerJob):
             submission output.
         """
         self.save_script()
+        self.run_path.mkdir(parents=True, exist_ok=True)
         # remove any slurm variables in case submitting from inside another slurm job
         env_vars_to_exclude = []
         for k in os.environ.keys():
--- a/cstar/execution/scheduler_job.py
+++ b/cstar/execution/scheduler_job.py
@@ -468,7 +468,7 @@ class SchedulerJob(ExecutionHandler, ABC):
         Writes the generated job script to the file specified by the `script_path` attribute.
         The file can then be used to submit the job to the scheduler.
         """
-        self.script_path.mkdir(parents=True, exist_ok=True)
+        self.script_path.parent.mkdir(parents=True, exist_ok=True)
         with open(self.script_path, "w") as f:
             f.write(self.script)
 
--- a/cstar/execution/scheduler_job.py
+++ b/cstar/execution/scheduler_job.py
@@ -19,6 +19,25 @@ from cstar.system.scheduler import (
 )
 
 
+def get_status_of_slurm_job(job_id: str) -> ExecutionStatus:
+    sacct_cmd = f"sacct -j {job_id} --format=State%20 --noheader"
+    msg_err = f"Failed to retrieve job status using {sacct_cmd}."
+    stdout = _run_cmd(sacct_cmd, msg_err=msg_err, raise_on_error=True)
+    # Map sacct states to ExecutionStatus enum
+    sacct_status_map = {
+        "PENDING": ExecutionStatus.PENDING,
+        "RUNNING": ExecutionStatus.RUNNING,
+        "COMPLETED": ExecutionStatus.COMPLETED,
+        "CANCELLED": ExecutionStatus.CANCELLED,
+        "FAILED": ExecutionStatus.FAILED,
+    }
+    for state, status in sacct_status_map.items():
+        if state in stdout:
+            return status
+    # Fallback if no known state is found
+    return ExecutionStatus.UNKNOWN
+
+
 def create_scheduler_job(
     commands: str,
     account_key: str,
@@ -32,6 +51,7 @@ def create_scheduler_job(
     queue_name: str | None = None,
     send_email: bool | None = True,
     walltime: str | None = None,
+    depends_on: list[str] | None = None,
 ) -> "SchedulerJob":
     """Create a scheduler job for either SLURM or PBS based on the system's active
     scheduler.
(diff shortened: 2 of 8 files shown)


cstar/entrypoint/worker/worker.py
```diff
@@ -188,9 +192,13 @@ def _log_disposition(self) -> None:
         if disposition == ExecutionStatus.COMPLETED:
             self.log.info("Simulation completed successfully.")
         elif disposition == ExecutionStatus.FAILED:
-            self.log.error("Simulation failed.")
+            msg = "Simulation failed."
+            self.log.error(msg)
+            raise CstarError(msg)
```
ankona, line 196: I don't think we need to have this message logged twice (here + exception handler). Consider getting rid of the `self.log.error` here and let the handler take care of it.

cstar/entrypoint/worker/worker.py
```diff
@@ -188,9 +192,13 @@ def _log_disposition(self) -> None:
...
         else:
-            self.log.warning(f"Simulation ended with status: {disposition}")
+            msg = f"Simulation ended with status: {disposition}."
+            self.log.warning(msg)
+            raise CstarError(msg)
```
ankona, line 199: Ditto on logging the same message twice

cstar/execution/scheduler_job.py
```diff
@@ -19,6 +19,33 @@
 )
 
 
+def get_status_of_slurm_job(job_id: str) -> ExecutionStatus:
+    """
```
ankona: supernit: I think the first line of docstring should have description & not a newline.

cstar/execution/scheduler_job.py
```diff
@@ -19,6 +19,33 @@
...
+    Check the status of a Slurm job using sacct.
+
+    :param job_id: the job_id to check
```
ankona: nit:
```suggestion
    Parameters
    ----------
    :param job_id: the job_id to check
```

cstar/execution/scheduler_job.py
```diff
@@ -19,6 +19,33 @@
...
+    sacct_status_map = {
+        "PENDING": ExecutionStatus.PENDING,
+        "RUNNING": ExecutionStatus.RUNNING,
+        "COMPLETED": ExecutionStatus.COMPLETED,
+        "CANCELLED": ExecutionStatus.CANCELLED,
+        "FAILED": ExecutionStatus.FAILED,
+    }
+    for state, status in sacct_status_map.items():
+        if state in stdout:
+            return status
```
ankona, line 50: There is no reason to loop here. Consider:

```python
return sacct_status_map.get(stdout, ExecutionStatus.UNKNOWN)
```

cstar/execution/scheduler_job.py
```diff
@@ -67,6 +95,8 @@ def create_scheduler_job(
         Whether to send email notifications about job status. Defaults to True.
     walltime : str, optional
         The maximum walltime for the job, in the format "HH:MM:SS". Defaults to the queue's maximum.
+    depends_on: list[str], optional
+        A list of job ids to pass as dependencies to the scheduler. Defaults to [].
```
ankona: "Defaults to []" appears to be incorrect since it defaults to `None`

cstar/execution/scheduler_job.py
```diff
@@ -674,8 +690,15 @@ def submit(self) -> int | None:
             k: v for k, v in os.environ.items() if k not in env_vars_to_exclude
         }
 
+        cmd = "sbatch"
+        if self.depends_on:
+            cmd += f" --dependency=afterok:{':'.join(str(d) for d in self.depends_on)}"
+
+        cmd += f" {self.script_path}"
+
+        print(cmd)
```
ankona: Consider using an f-string to make the final format string obvious:

```python
csv_deps = ':'.join(str(d) for d in self.depends_on)
dep_clause = f" --dependency=afterok:{csv_deps}" if csv_deps else ""

cmd = f"sbatch {dep_clause} {self.script_path}"
```

cstar/orchestration/dag_runner.py
```diff
@@ -0,0 +1,141 @@
...
+        run_path=bp.runtime_params.output_dir,
+        job_name=None,  # to fill with some convention
+        output_file=None,  # to fill with some convention
+        queue_name=os.getenv("CSTAR_QUEUE_NAME"),
+        walltime="00:10:00",  # TODO how to determine this one?
+        depends_on=job_dep_ids,
+    )
+
+    job.submit()
+    print(f"Submitted {step.name} with id {job.id}")
```
ankona, line 70: Consider using a logger here or add in `@task(log_prints=True, ...` to make sure prefect tracks this output

cstar/orchestration/dag_runner.py
```diff
@@ -0,0 +1,141 @@
...
+    return str(job.id)
```
ankona, line 71: Consider returning the full job id of the form f"{ALLOCATION_ID}.{JOB_ID}". This gives us the ability to:
- avoid casting ints everywhere
- identify the allocation of a job from a single identifier
- switch painlessly to cstar-supplied identifiers for jobs (e.g. `srun --job-name=TaskXYZ ...`) that are more descriptive and can be derived from a `Step` in a workplan.

cstar/roms/simulation.py
```diff
@@ -1068,11 +1069,21 @@ def setup(
...
+        for codebase in filter(lambda x: x is not None, self.codebases):  # type: ExternalCodeBase
             self.log.info(f"🔧 Setting up {codebase.__class__.__name__}...")
+
+            # if we're running a workplan, for now, set up a code directory for each
+            # step, otherwise they may try to clobber each other or get tripped up on
+            # detecting existing directories.
+            if os.getenv("CSTAR_ORCHESTRATED", "0") == "1":
+                codebase_dir = codebases_dir / codebase.expected_env_var.split("_")[0]
+                codebase_dir.mkdir(parents=True, exist_ok=True)
+                codebase.get(codebase_dir)
```
ankona, line 1085: this is a non-trivial startup delay with even a small amount of steps. Consider splitting this into its' own prefect task:

```python
@task 
def pull_codebase(codebase_dir: Path, codebase: Codebase) -> None:
  codebase_dir = codebase_dir / codebase.expected_env_var.split("_")[0]
  codebase_dir.mkdir(parents=True, exist_ok=True)
  codebase.get(codebase_dir)
```

which you can then run in parallel...

```python
tasks = [pull_codebase.submit(codebase_dir, cb) for cb in self.codebases]
wait(tasks)
```

ankona: alternatively, we can inspect the codebases in the steps to identify the unique ones & use the os function to `cp -r ...` the directory.

ankona: alternatively, consider using the run id you previously defined to manipulate a prefix:

```python
class Codebase:
  ...

class RunAwarePath:
  def __enter__(self, path: Path) -> path:
    if run_id := os.getenv("CSTAR_RUNID"):
      return path / run_id
    return path
...

with RunAwarePath(codebase_dir) as cbd:
  cbd.mkdir(parents=True, exist_ok=False)
  codebase.get(cbd)

File: cstar/execution/scheduler_job.py (hunk 4 above)
Comment (ScottEilerman) [reply cstar/execution/scheduler_job.py]: gah I need to update pycharm to use the docstring syntax we actually use instead of the one I'm used to. thx.

File: cstar/execution/scheduler_job.py (hunk 5 above)
Comment (ScottEilerman) [reply cstar/execution/scheduler_job.py:50]: Not my code, and I was about to change it as you suggest, then realized we're actually searching the full stdout response from sacct for the given status, not looking up the status in the enum map. There are certainly better ways to do this but I'm going to leave it for the moment, since we may need to expand / modify our status checking anyway (per noted deficiencies in PR description).

File: cstar/execution/scheduler_job.py (hunk 6 above)
Comment (ScottEilerman) [reply cstar/execution/scheduler_job.py]: good catch. swear I did this somewhere but I guess not. changed typehint to `Iterable[str]` and default to `()` (because `[]` would be a mutable default yada yada yada)

File: cstar/orchestration/dag_runner.py (hunk 8 above)
Comment (ScottEilerman) [reply cstar/orchestration/dag_runner.py:70]: will do log_prints for now since the integration point (and real logging tie-ins) of this code is TBD.

File: cstar/orchestration/dag_runner.py (hunk 9 above)
Comment (ScottEilerman) [reply cstar/orchestration/dag_runner.py:71]: so right now the sacct for one of these jobs looks like:
```
JobID           JobName  Partition    Account  AllocCPUS      State ExitCode 
------------ ---------- ---------- ---------- ---------- ---------- -------- 
14205318     cstar_job+  wholenode  ees250129        256 CANCELLED+      0:0 
14205318.ba+      batch             ees250129        128  CANCELLED     0:15 
14205318.ex+     extern             ees250129        256  COMPLETED      0:0 
14205318.0         roms             ees250129        256  COMPLETED      0:0 
```
(note this job was cancelled / timed out because I haven't gotten around to finishing the parallelized join step and I set my walltime too low to allow for the slow serial version)

The jobid here is `14205318` I am not entirely sure I understand the `ba+` and `ex+` extensions, though I could speculate. I assume the `.0` task gets created once we actually `srun` roms which is some amount of time into the overall job.

If each job/allocation is a one-off step It seems like the overall jobid is what we want to track? My "bad" timed-out case is actually a good example, roms itself completed but the overall process didn't finish so the job as a whole is bad (unless/until we split out the "subjobs" and have the ability to rerun just the join portion, e.g..

File: cstar/roms/simulation.py (hunk 10 above)
Comment (ScottEilerman) [reply cstar/roms/simulation.py:1085]: I agree there's lots of room for improvement and speedup here, but the order zero task was to make it not crash. Since this is on the worker side of things, it's currently not running prefect at all, it's just python-ing along. We could still parallelize for sure, but I'm iffy on adding a prefect layer over there at the moment.

File: cstar/execution/scheduler_job.py (hunk 5 above)
Comment (ankona) [reply cstar/execution/scheduler_job.py:50]: the format string appears to return only the state:


Each line of the `sacct` output should only contain the status from the enumeration. Consider:

image

It looks for the first appearance of any status and assumes that's the status for the entire job. This is a problem for any job running more than one subtask we can easily solve.

If the Job ID is included, we can avoid all looping and solve this problem.

using `sacct -j 44304934 --format=JobID,State --noheader`

image

```python
>>> from collections import defaultdict
>>> from cstar.execution.handler import ExecutionStatus
>>>
>>> sacct_out = "12345 RUNNING\n12345.1 RUNNING\n12345.2 COMPLETED"
>>>
>>> lines = [x.split(maxsplit=1) for x in sacct_out.split("\n")]
>>> 
>>> task_stats = defaultdict(lambda: str(ExecutionStatus.UNKNOWN),
...                          {line[0]: line[1] for line in lines})
>>> task_stats["12345"]
'RUNNING'
>>>
>>> raw_status = task_stats["12345"]
>>>
>>> status = ExecutionStatus[raw_status]
>>> status
<ExecutionStatus.RUNNING: 3>
```

We can now:

- avoid soft dependency on state order (least to most progress)
- index directly
- be guaranteed to have the right status
- solve unknowns "inline"
- skip creating the map in this function of "states to iterate through"
- update multiple statuses from a single `sacct` call
- retrieve subtask statuses instead of only the main SLURM job

cstar/orchestration/dag_runner.py
ankona: - `ba+` and `ex+` are what happens when the column size isn't large enough to display the full job id (so, the `+` is the way I'd normally use ellipses: `pi=3.141...`

- each srun in a batch gets an "auto-numbered" step ID: `xxx.0`, `xxx.1`, ..., `xxx.N`

If we want any visibility into the behavior of the job, we want the full IDs. But... running everything as a new batch can negate this.

cstar/roms/simulation.py
ankona: noted as a potential task for the future.

cstar/execution/scheduler_job.py
ankona: an alternative here (if we don't care about subtasks) is pretty simple:

```python
>>> sacct_out = "12345 RUNNING\n12345.1 RUNNING\n12345.2 COMPLETED"
>>>
>>> status = next(x for x in sacct_out.split("\n") if x in ExecutionStatus.__dict__)
>>> status
'RUNNING'
```

This also takes the first status that occurs and gets rid of the local status dict.

cstar/execution/scheduler_job.py
ankona: I think I did this originally. It's more my fault than yours... Thanks for fixing my mistake

ScottEilerman:
per slack, going to go ahead and merge and handle the rest of these as we go.

PR #359, opened 2025-10-21
Parallelize join operation

ScottEilerman:
This PR speeds up the post_run join operation by running ncjoin from a thread pool.

Note that the following options were additionally tried and for the time being, this method was the fastest:
* using a process pool instead of a thread pool was a slight decrease in performance since we are calling a subprocess anyway, so the extra process overhead is for nothing
* using a process pool (or thread pool) with roms-tools' join_netcdf function was about twice as slow compared to ncjoin
* I tried a variety of hacky implementations to use a dask client with our roms-tools joining functions and they were all laughably slow (like 10x the simple solution here). I'm sure I was still not using dask quite right, and that with a lot of tinkering we should be able to get good performance out of it, but I'll take this quick and easy speed-up for now.

The default number of threads is set to `cpu_count // 3` which was the sweet spot for my mac (partly due to M4 performance vs efficiency processors), but is almost certainly not the optimized number for an HPC. I'd worry a little bit about going too wide on the HPC and blowing out memory. For now, `cpu_count // 3` _seems_ conservative, and I added an env-var so that users can tweak to their use case if desired. We can do a better optimization pass someday...

I guess I also made a sort of arbitrary decision to change the output structure. Instead of moving the partitioned files into a subdirectory of output_dir after joining, I move the joined files into an adjacent JOINED_OUTPUT directory. This seemed a little safer and predictable in terms of leaving the original outputs where they are, and having them all there if the join process gets interrupted. It's also way fewer move/rename operations, which should be fast anyway, but 🤷 . Open to arguments against this change.

- [x] Tests passing
- [x] Full type hint coverage


dafyddstephenson:
Cool - haven't tried it but looks pretty unintrusive? Have you used it on a login node on PM/Anvil?

I think longer-term the general issue with `ncjoin` in C-Star is that the operation is still running on a resource-limited node and will fail for large runs. I'm guessing that this will be handled with the workflow stuff. I would still maintain that a jobarray is potentially the most portable/scalable way to parallelize this problem for HPC, even if it isn't the most elegant. These is also the MPI version of `ncjoin` which may be worth poking at further if we're serious about parallelizing this, but again, PIO is coming...

I never implemented any sort of scheduler version of `ncjoin` in C-Star because I think it's not obvious _when_ it's safe to run `ncjoin` on a login node vs. when it's necessary to submit it as a job. Factoring in queueing time there's a sweet spot beyond which it's faster to submit a job, and it varies between simulations. It's also possible to just roll it into `run()` and hope there's enough walltime for both. If we're doing smart walltime calculations to split jobs up, this might be worth revisiting? I realize I've now been rambling for 2 paragraphs bye

ScottEilerman:
Tried this on Anvil (on a compute node, as part of a full C-Star simulation). Worked great! Going to go ahead and merge it into orchestration for now (trying to reduce spaghetti merges between main/dev/orch, just get everything into orch which will make its way to dev and main soon enough).

PR #360, opened 2025-10-21
Fix CI (codecov and docs build failures)

ScottEilerman:
Thanks to @NoraLoose for already fixing the codecov failures in roms-tools and paving the way.

Additionally, I'm pinning the docs build to use python 3.12 for now. It actually didn't have a matrix defined and wasn't pinning a python version at all, which is why if was getting 3.14 (which fails).

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -51,9 +51,12 @@ jobs:
         shell: micromamba-shell {0}                
         run: |
           coverage report -m ; coverage xml
-
+        with:
+          fail_ci_if_error: false
+          
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v4.0.1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.xml
+          fail_ci_if_error: false
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -50,10 +50,9 @@ jobs:
       - name: Get coverage report
         shell: micromamba-shell {0}                
         run: |
-          coverage report -m ; coverage xml
-        with:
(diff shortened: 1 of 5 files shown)


dafyddstephenson (approved):

.github/workflows/integration_tests.yaml
```diff
@@ -50,10 +50,12 @@ jobs:
       - name: Get coverage report
         shell: micromamba-shell {0}                
         run: |
-          coverage report -m ; coverage xml
+          coverage report -m || true ; coverage xml || true
```
dafyddstephenson, line 53: approved but what's the `|| true` here doing?

.github/workflows/integration_tests.yaml
```diff
@@ -50,10 +50,12 @@ jobs:
...
+
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v4.0.1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.xml
+          fail_ci_if_error: false
```
dafyddstephenson, line 61: Is this necessary for the fix?

.github/workflows/integration_tests.yaml
ScottEilerman: Both this and the one below, I copied from Nora and ROMS-tools, and they deal with the CI sometimes failing if the coverage report can't upload to codecov (either because codecov is having an issue, or if you're on a fork and don't have the creds on your fork). The `|| true` is just a shell hack to get a "success" exit code regardless of whether `coverage` fails.

.github/workflows/integration_tests.yaml
ScottEilerman: Not for the docs build stuff, separate issue (see other comment) but decided to do it while I was in CI land.

PR #365, opened 2025-10-22
CSD-310: add CLI

ankona:
Implement initial CLI


cstar/cli/blueprint/actions/check.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Perform content validation on the blueprint supplied by the user.
+
+    Parameters
+    ----------
+    ns : argparse.Namespace
+        User inputs parsed by the CLI
+    """
+    try:
+        model = deserialize(ns.path, RomsMarblBlueprint)
```
ScottEilerman, line 20: Future TODO: we are going to need an application-aware deserializer of sorts to handle different applications having different blueprint schemas/classes. We can probably use a base blueprint model that has the req'd fields (including `application`) and from there, map to the correct application/class.

cstar/cli/blueprint/actions/check.py
```diff
@@ -0,0 +1,54 @@
...
+        """Add a parser for the command: `cstar blueprint check path/to/blueprint.yaml`"""
+        parser = sp.add_parser(
+            action,
+            help="Validate the contents of a blueprint",
+            description="Validate the contents of a blueprint.",
+        )
+        parser.add_argument(
+            dest="path",
+            help="Path to the blueprint (YAML)",
+            action=PathConverterAction,
```
ScottEilerman, line 49: is `type=Path` sufficient here instead of a custom action?

cstar/cli/blueprint/actions/run.py
```diff
@@ -0,0 +1,122 @@
...
+    path : Path
+        The path to the blueprint to execute
+
+    Returns
+    -------
+    SimulationRunner
+        A simulation runner configured to execute the blueprint
+    """
+    account_id = "m4632"
+    walltime = "48:00:00"
```
ScottEilerman, line 26: We should probably have account_id pull from the same `CSTAR_ACCOUNT_KEY` env-var. And maybe we should make a walltime env-var for now too. That's still hard-coded to 10 minutes in the slurm launcher, I believe. Hopefully we'll eventually have a smart way to predict the right value, but user-specified is the quickest and most flexible path for now.

cstar/cli/workplan/actions/run.py
```diff
@@ -0,0 +1,61 @@
...
+    Parameters
+    ----------
+    ns : argparse.Namespace
+        User inputs parsed by the CLI
+    """
+    # TODO: load from ~/.cstar/config (e.g. cstar config init)
+    os.environ["CSTAR_INTERACTIVE"] = "0"
+    os.environ["CSTAR_ACCOUNT_KEY"] = "ees250129"
+    os.environ["CSTAR_QUEUE_NAME"] = "shared"
+    os.environ["CSTAR_ORCHESTRATED"] = "1"
```
ScottEilerman, line 22: Per slack discussion, we should either remove these (and document somewhere all the `CSTAR_` options we have made and/or at a minimum, os.getenv the values we set here. At the moment, I can't change the queue name without changing the hard code here.

cstar/cli/workplan/actions/run.py
```diff
@@ -0,0 +1,61 @@
...
+
+    os.environ["CSTAR_RUNID"] = ns.name
```
ScottEilerman, line 24: Since `name` is optional, you get an error here if it's not specified (python won't cast None to str)

ScottEilerman (commented):
Still reviewing but want to push the current comments before I get distracted by meetings tomorrow. I've reviewed everything under `cli`, haven't gotten into `orchestration` yet.

cstar/orchestration/launch/slurm.py
```diff
@@ -0,0 +1,303 @@
...
+            account_key=os.getenv("CSTAR_ACCOUNT_KEY", ""),
+            cpus=bp.cpus_needed,
+            nodes=None,  # let existing logic handle this
+            cpus_per_node=None,  # let existing logic handle this
+            script_path=None,  # puts it in current dir
+            run_path=bp.runtime_params.output_dir,
+            job_name=job_name,
+            output_file=None,  # to fill with some convention
+            queue_name=os.getenv("CSTAR_QUEUE_NAME"),
+            walltime="00:10:00",  # TODO how to determine this one?
```
ScottEilerman, line 171: I can't remember if I left this note somewhere already or only in my head.
1. We should introduce an env-var for walltime control, until we come up with a broader scheme for launcher config management.
2. We should collect and document somewhere all of the env-vars we're creating.
3. Maybe now is the time, or else after this PR but before we really put this into internal testing, we should do a pass on env-var/config naming conventions. For example, maybe anything used in this function gets changed to `CSTAR_SLURM_[ACCOUNT|QUEUE|WALLTIME]` for clarity.

ScottEilerman: (acknowledging that I created this mess to begin with)

cstar/orchestration/dag_runner.py
```diff
@@ -1,141 +1,187 @@
...
+        print(f"[on-exit::{mode}] Open nodes: {open_set}, Closed: {closed_set}")
+
+        sleep_duration = next(delay_iter)
+        print(f"Sleeping for {sleep_duration} seconds before next {mode}.")
+        await asyncio.sleep(sleep_duration)
+
+    print(f"Workplan {mode} is complete.")
+
+
+# @flow(log_prints=True)
```
ScottEilerman, line 130: It's a little curious that we're not invoking a Prefect flow anywhere, as far as I can tell. In general I'm not really sure if Prefect is tracking the task graph at all. I guess it would only be doing so if outputs from one flow to inputs of another task. 

Based on a live test, both the overall graph layout and the task caching for the submit still works fine, so maybe we don't need anything more for Prefect for now, and can reconsider further integration when/if we get to the centralized, persistent orchestrator.

cstar/orchestration/launch/slurm.py
```diff
@@ -0,0 +1,303 @@
...
+    Returns
+    -------
+    str
+        The complete CLI command.
+    """
+    return f'{sys.executable} -c "import time; time.sleep(10)"'
+
+
+app_to_cmd_map: dict[str, StepToCommandConversionFn] = {
+    Application.ROMS.value: convert_roms_step_to_command,
```
ScottEilerman: I was going to go make this comment where we define `Application` in models.py, but apparently this came in from a previous PR, and this is where I noticed it anyway. At the moment, I don't think we have any plans to support non-MARBL ROMS runs. It totally makes sense as a thing a person could do, and we might want some form of it in the next year or so, but there would be more work needed on other parts of the codebase to really enable it as a robust option.

cstar/orchestration/models.py
```diff
@@ -440,6 +440,24 @@ class Step(BaseModel):
...
+    #         The value received from the input.
+
+    #     Returns
+    #     -------
+    #     Path
+    #         The input value converted to a pathlib.Path
+
+    #     """
+    #     return Path(value)
+
```
ScottEilerman, line 459: Note commented code, up to you to keep/remove.

cstar/orchestration/orchestration.py
```diff
@@ -0,0 +1,652 @@
...
+class Status(IntEnum):
+    """The state of a running task."""
+
+    Unsubmitted = auto()
+    """A task that has not been submitted by a launcher."""
+    Submitted = auto()
+    """A task that has been submitted by a launcher and awaits a status update."""
+    Running = auto()
+    """A task that was submitted and has not terminated."""
+    Ending = auto()
```
ScottEilerman, line 52: Do you have a planned use for Ending? I could imagine something like getting a spot termination notice and entering a checkpoint dump state. Or when/if we have better two-way communication happening between the worker and the orchestrator, we could obviously signal this.

cstar/orchestration/orchestration.py
```diff
@@ -0,0 +1,652 @@
...
+        Paramters
+        ---------
+        status : "Status"
+            The status to evaluate.
+
+        Returns
+        -------
+        bool
+        """
+        return status in {Status.Submitted, Status.Running, Status.Ending}
```
ScottEilerman, line 104: If Submitted ends up mapping to Slurm PENDING (not sure yet if it does), maybe we shouldn't consider that as Running.

cstar/orchestration/launch/slurm.py
ScottEilerman: Maybe we just do item 1 above in this PR, and push the rest to a follow-up. Default walltime could be like, 48 hours for now -- could argue for different numbers but the intention is people should be setting it themselves in most cases.

ScottEilerman (commented):
pushing another batch of comments in between meetings. I'm up to orchestration.py

cstar/cli/blueprint/actions/run.py
ScottEilerman: create task

cstar/cli/blueprint/actions/check.py
ScottEilerman: create task

cstar/cli/blueprint/actions/check.py
ScottEilerman: ignore, typer will handle

cstar/cli/workplan/actions/run.py
ScottEilerman: create task

cstar/cli/workplan/actions/run.py
ScottEilerman: I think you fixed this one, or else we can refactor it with typer

cstar/orchestration/launch/slurm.py
ScottEilerman: same env task as previous comment

cstar/orchestration/dag_runner.py
ScottEilerman: we've talked about this one, follow up task for orchestrator

cstar/orchestration/launch/slurm.py
ScottEilerman: remove

cstar/orchestration/orchestration.py
ScottEilerman: ignore for now

ScottEilerman (approved):
Per discussion, we're merging this into`orchestration` and building follow-up tasks to knock out quickly from there.

cstar/cli/blueprint/actions/check.py
ankona: Added CSD-332

cstar/cli/workplan/actions/run.py
ankona: yeah, this particular bug was previously fixed.

cstar/cli/blueprint/actions/run.py
ankona: Added `CSD-333`

cstar/cli/workplan/actions/run.py
ankona: Added `CSD-333`

cstar/orchestration/launch/slurm.py
ankona: added `CSD-333`

cstar/orchestration/launch/slurm.py
ankona: Removed!

cstar/orchestration/dag_runner.py
ankona: Added `CSD-334`

cstar/orchestration/models.py
ankona: removed

cstar/orchestration/orchestration.py
ankona: covered in CSD-334

cstar/orchestration/orchestration.py
ankona: This class reproduces the breadth of statuses covered by SLURM with the intent of using the integer values to enable more intuitive comparisons (which got lost somewhere in a refactor) and avoiding frequent changes by using "ranges" instead of specific set comparisons:

e.g.

```
if current_status > Status.Unsubmitted:
  ... # task has been sent to a launcher

if current_status > Status.Submitted:
  ... # task status can be queried

if current_status > Status.Ending:
  ... # tasks is terminal

if current_status > Status.Done:
  ... # task is not successful
```

A local launcher via prefect might soft-encode something like ending (e.g. duration exceeds 75% of average time consumed in prior task runs), or choose not to implement returning all statuses. I do not have a specific scenario for using `Ending`, yet.

PR #368, opened 2025-11-17
Set partitioning attribute if source is partitioned

ScottEilerman:
I was getting an error where C-Star was trying to re-partition input files that were already partitioned. I think this was introduced with the input dataset refactor. This fix sets the older `partitioning` attribute if the newer`source_partitioning` is defined, and therefore ensures that the `partition()` method exits immediately. I'm not sure if there is an intended difference between the two partitioning-related properties or if they're duplicative; we can refactor later if one or the other should be pruned out.

### DIFF
--- a/cstar/roms/input_dataset.py
+++ b/cstar/roms/input_dataset.py
@@ -97,8 +97,8 @@ class ROMSInputDataset(InputDataset, ABC):
                 locations.append(location.replace(old_suffix, new_suffix))
             self.partitioned_source = SourceDataCollection.from_locations(locations)
             self.partitioning = ROMSPartitioning(
-                np_xi=source_np_xi,
-                np_eta=source_np_eta,
+                np_xi=self.source_np_xi,  # type: ignore[arg-type]
+                np_eta=self.source_np_eta,  # type: ignore[arg-type]
                 files=[Path(p) for p in self.partitioned_source.locations],
             )
         self._working_copy = None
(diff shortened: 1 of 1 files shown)


cstar/roms/input_dataset.py
```diff
@@ -96,6 +96,11 @@ def __init__(
                 new_suffix = f".{i:0{ndigits}d}.nc"
                 locations.append(location.replace(old_suffix, new_suffix))
             self.partitioned_source = SourceDataCollection.from_locations(locations)
+            self.partitioning = ROMSPartitioning(
+                np_xi=self.source_np_xi,  # type: ignore[arg-type]
```
ankona, line 101: Is there a good reason why we want these linter mitigations? 

I think the vars are `int | None` in the `ROMSInputDataset`, while `ROMSPartitioning` specifies `int`. This will probably result in a runtime error that we'll need to track down (e.g. after some math, receiving `No operation for None-type and int`)

Should we use `np_xi=self.source_np_xi if self.source_np_xi is not None else 0` to ensure an unexpected null doesn't slip in?

- Ditto on `np_eta`

ankona (approved):
Conditionally approved w/optional avoidance of linter mitigations per your judgement.

cstar/roms/input_dataset.py
ScottEilerman: The `if self.source_partitioning` block it's under will only be entered if both of the values are non-None, but mypy isn't quite smart enough to realize it. I was going to add some additional asserts right above or rework things a bit and realized I already convinced myself it was fine and just trying to convince the linter, so I said "trust me bro" instead.

ScottEilerman: ya know what i'll add a comment saying just that.

PR #369, opened 2025-11-25
Slurm improvements

ScottEilerman:
Small PR with a few slurm/sbatch related changes:
* Add `--kill-on-invalid-dep=yes` so that if a job's dependency fails, slurm kills the dependent job instead of it staying in pending forever.
* Add `set -e` to the sbatch script. The FD team found this was needed on Anvil to tell batch to exit properly if ROMS had an error. I _think_ this isn't strictly needed since we already wrap the ROMS execution in python subprocesses and our python script should exit cleanly when appropriate, but I also had a few cases where things timed out and I don't know for sure if this would have helped or if it was a different issue. I think it also doesn't hurt to add it?

### DIFF
--- a/cstar/execution/scheduler_job.py
+++ b/cstar/execution/scheduler_job.py
@@ -666,8 +666,9 @@ class SlurmJob(SchedulerJob):
         ) in self.scheduler.other_scheduler_directives.items():
             scheduler_script += f"\n#SBATCH {key} {value}"
 
+        scheduler_script += "\n\nset -e"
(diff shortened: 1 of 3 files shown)


ankona (approved):
Since we run every simulation / task in a separate batch, I don't have any reservations regarding `set -e`. Until we identify some optional step, such as starting a jupyter server at the end of a run, we don't have a need to to allow individual items to "fail quietly..." 

LGTM.

PR #370, opened 2025-11-25
Remove user env file

ScottEilerman:
Per discussion, these sorts of variables should really be set in blueprints or via other mechanisms in the future. For now, we are having lots of trouble with simultaneous runs writing/reading from the same user env location.

### DIFF
--- a/cstar/roms/external_codebase.py
+++ b/cstar/roms/external_codebase.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 
 from cstar.base.external_codebase import ExternalCodeBase
@@ -37,7 +38,7 @@ class ROMSExternalCodeBase(ExternalCodeBase):
         roms_root = self.working_copy.path
         cstar_sysmgr.environment.set_env_var(self.root_env_var, str(roms_root))
         cstar_sysmgr.environment.set_env_var(
-            "PATH", f"${{PATH}}:{roms_root / 'Tools-Roms'}"
+            "PATH", f"{os.environ.get('PATH')}:{roms_root / 'Tools-Roms'}"
         )
 
         # Compile NHMG library
--- a/cstar/roms/external_codebase.py
+++ b/cstar/roms/external_codebase.py
@@ -38,7 +38,7 @@ class ROMSExternalCodeBase(ExternalCodeBase):
         roms_root = self.working_copy.path
         cstar_sysmgr.environment.set_env_var(self.root_env_var, str(roms_root))
         cstar_sysmgr.environment.set_env_var(
-            "PATH", f"{os.environ.get('PATH')}:{roms_root / 'Tools-Roms'}"
+            "PATH", f"{roms_root / 'Tools-Roms'}:{os.environ.get('PATH')}"
         )
 
         # Compile NHMG library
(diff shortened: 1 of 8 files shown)


cstar/system/environment.py
```diff
@@ -318,9 +303,13 @@ def load_lmod_modules(self, lmod_file) -> None:
         for mod in lmod_list:
             self._call_lmod(f"load {mod}")
 
-    def set_env_var(self, key: str, value: str) -> None:
-        """Set value of an environment variable and store it in the user environment
-        file.
+    @staticmethod
```
ankona, line 306: Per your comment, this method was created to wrap the need to update both the dotenv file and local environment. Can we just add a `# TODO: remove method, user-env updates are deprecated and this is no longer necessary` instead of the story? `# TODO` will be easy to find later, versus the blurb.

cstar/roms/external_codebase.py
```diff
@@ -37,7 +38,7 @@ def _configure(self) -> None:
         roms_root = self.working_copy.path
         cstar_sysmgr.environment.set_env_var(self.root_env_var, str(roms_root))
         cstar_sysmgr.environment.set_env_var(
-            "PATH", f"${{PATH}}:{roms_root / 'Tools-Roms'}"
+            "PATH", f"{os.environ.get('PATH')}:{roms_root / 'Tools-Roms'}"
```
ankona: The prior value should be automatically expanded without the need to copy it "verbatim", just like:

`export PATH="$PATH:/my/new/thing"`

The double squiggly-brackets tell python "this is not an f-string replacement". 

Are you sure you need to expand manually? I know `dotenv` will expand values with `load_dotenv` but copies as-is when using `dotenv_values`

The current version makes it much easier to quickly debug/eyeball issues (e.g. i don't have to look through 1000 characters of wrapping path variable content to get to the end and see tools-roms)

ankona (changes requested):
Let's clean up the deprecation comment so it can be discovered later (or create a ticket and remove the comment altogether).

Otherwise, LGTM.

cstar/tests/unit_tests/system/test_environment.py
```diff
@@ -440,9 +400,8 @@ def test_env_file_load_count(
 
             env.set_env_var(new_var, new_value)
 
-            # update should trigger load from disk
-            assert loader.call_count == initial_num_calls + 1
-            assert new_var in env.environment_variables
+            # update should not trigger new load from disk
+            assert loader.call_count == initial_num_calls
```
ankona: This assertion will pass, but it makes no sense and could just be omitted (dotenv is not reloading the environment by writing the .cstar.env file). This probably means the entire test can go

ScottEilerman: You're right, the test can go now. At some point I had it still reloading the system env file but there was no reason for that: either it's clobbering changes you were trying to make, or it's reloading the same thing it already loaded.

cstar/system/environment.py
ScottEilerman: added todo for now

cstar/roms/external_codebase.py
ScottEilerman: Per discussion:
* since we don't dump back out to a .env file before reading back in, bash doesn't get a pass at expanding $PATH
* sys.path.insert doesn't end up working since it only controls python import paths and we need this for a subprocess
* I did at least swap the order so our new paths go in the front

PR #371, opened 2025-11-25
Add function to clear the working directory before launching a job

ScottEilerman:
* Added function that clears problematic directories (e.g. ones that will make RomsSimulation crash if they are already populated) ahead of launching a job in slurm. As I write that, maybe we want this functionality somewhere more general, but that location is working for me for now
* Put this functionality behind an opt-in env-var. I'm going to set this in my bash profile, but probably best to not accidentally delete things for some new user. If/when we get to the point of fully owning the paths for every run, we might reconsider making it default behavior.
* Added unit test to ensure behavior respects env-var

### DIFF
--- a/cstar/orchestration/utils.py
+++ b/cstar/orchestration/utils.py
@@ -1,4 +1,10 @@
+import os
 import re
+import shutil
+from pathlib import Path
+
+from cstar.orchestration.models import RomsMarblBlueprint, Step
+from cstar.orchestration.serialization import deserialize
 
 
 def slugify(source: str) -> str:
@@ -18,3 +24,23 @@ def slugify(source: str) -> str:
         raise ValueError
 
     return re.sub(r"\s+", "-", source.casefold())
+
+
+def clear_working_dir(step: Step) -> None:
+    """Clear the working directory for a step if CSTAR_CLOBBER_WORKING_DIR is set.
+
+    Parameters
+    ----------
+    step: the step who's working directory should be cleared
+
+    Returns
+    -------
+    None
+    """
+    if os.getenv("CSTAR_CLOBBER_WORKING_DIR") == "1":
+        _bp = deserialize(Path(step.blueprint), RomsMarblBlueprint)
+        out_path = _bp.runtime_params.output_dir
+        print(f"clearing {out_path}")
+        shutil.rmtree(out_path / "ROMS", ignore_errors=True)
+        shutil.rmtree(out_path / "output", ignore_errors=True)
+        shutil.rmtree(out_path / "JOINED_OUTPUT", ignore_errors=True)
--- a/cstar/orchestration/utils.py
+++ b/cstar/orchestration/utils.py
@@ -3,9 +3,6 @@ import re
 import shutil
 from pathlib import Path
 
-from cstar.orchestration.models import RomsMarblBlueprint, Step
-from cstar.orchestration.serialization import deserialize
-
 
 def slugify(source: str) -> str:
     """Convert a source string into a URL-safe slug.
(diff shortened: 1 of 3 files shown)


cstar/orchestration/utils.py
```diff
@@ -18,3 +24,23 @@ def slugify(source: str) -> str:
...
+
+    Returns
+    -------
+    None
+    """
+    if os.getenv("CSTAR_CLOBBER_WORKING_DIR") == "1":
+        _bp = deserialize(Path(step.blueprint), RomsMarblBlueprint)
+        out_path = _bp.runtime_params.output_dir
+        print(f"clearing {out_path}")
+        shutil.rmtree(out_path / "ROMS", ignore_errors=True)
```
ankona: Can we just `shutil.rmtree(out_path)`?

- bad suggestion. you'd have to recreate it and gain nothing. ignore

ankona (approved):
I'm not sure what's left in the output directory, but we might be able to remove 2 lines by deleting a parent. If not, the rest looks good to me.

cstar/orchestration/utils.py
```diff
@@ -18,3 +24,23 @@ def slugify(source: str) -> str:
...
+    Parameters
+    ----------
+    step: the step who's working directory should be cleared
+
+    Returns
+    -------
+    None
+    """
+    if os.getenv("CSTAR_CLOBBER_WORKING_DIR") == "1":
+        _bp = deserialize(Path(step.blueprint), RomsMarblBlueprint)
```
ankona: i added automatic str-to-path conversion in deserialize in my last commit. feel free to omit `Path()` wrapping this.. 

...actually, i take it back. i was lazy allowing strings and should have gone this route. ignore this.

cstar/tests/unit_tests/orchestration/test_utils.py
```diff
@@ -0,0 +1,75 @@
...
+import pytest
+
+from cstar.orchestration.utils import clear_working_dir
+
+
+@pytest.fixture
+def populated_output_dir(tmp_path: Path) -> tuple[Path, list[Path]]:
+    output_dir = tmp_path / "my_output_dir"
+    files = [
+        (output_dir / "ROMS" / "some_file"),
```
ankona, line 12: I wonder if we should create a ticket just to come back and look at these output directory names.

It's out of scope here, but it would be nice to avoid the inconsistency of some being lower-case and some being all-caps, if possible.

cstar/orchestration/utils.py
ankona: It seems really heavy-weight to force clearing the working dir to have to deserialize a blueprint (again). 

In line 152 of `slurm.py` you've already deserialized. Can't this be tweaked to pass the `Path` instead of a `Step`?:

```py
# slurm.py
        ...
        bp = deserialize(bp_path, RomsMarblBlueprint)
        job_dep_ids = [d.pid for d in dependencies]

        clear_working_dir(Path(bp.runtime_params.output_dir))
```
and
```
def clear_working_dir(working_dir_path: Path) -> None:
   ...
```

The bonus would be that we can re-use `clear_working_dir` with the upcoming Blueprint classes, e.g. 

`list(map(shutil.rmtree, [output_dir / subdir for subdir in os.listdir(output_dir)]))`

cstar/tests/unit_tests/orchestration/test_utils.py
ScottEilerman: We can include this in the upcoming path overhaul! I'll make a note.

cstar/orchestration/utils.py
ScottEilerman: agree! will change.

PR #372, opened 2025-11-26
Env var cleanup

ScottEilerman:
* Removed CSTAR_INTERACTIVE and related functionality (it was only being used to confirm indefinite updates, and in a CLI prompt to create a directory)
* Rename SLURM env-vars for clarity
* Add walltime env-var
* Rename ORCHESTRATED env-var to FRESH_CODEBASES  for it's only current usage
* Document env-vars in readme

Note: this is a bandaid for a better eventual config solution. I'm not concerned with things like validating/casting walltime to the appropriate HH:MM:SS format, or centralized loading or default management at the moment. This at least surfaces things and removes some problematic hardcodes. I intend to stuff my own defaults into my .bashrc for now and forget about it.


cstar/roms/simulation.py
```diff
@@ -1186,7 +1186,7 @@ def setup(
             # if we're running a workplan, for now, set up a code directory for each
             # step, otherwise they may try to clobber each other or get tripped up on
             # detecting existing directories.
-            if os.getenv("CSTAR_ORCHESTRATED", "0") == "1":
+            if os.getenv("CSTAR_FRESH_CODEBASES", "0") == "1":
```
ankona, line 1189: this fits with our prior discussion on renaming `CSTAR_CLOBBER_DIR` to `CSTAR_FORCE_INITIALIZE`
- the renamed env var could be used for both wiping existing directories and cloning code

cstar/orchestration/launch/slurm.py
```diff
@@ -161,16 +161,18 @@ async def _submit(step: Step, dependencies: list[SlurmHandle]) -> SlurmHandle:
...
             cpus_per_node=None,  # let existing logic handle this
             script_path=None,  # puts it in current dir
             run_path=bp.runtime_params.output_dir,
             job_name=job_name,
             output_file=None,  # to fill with some convention
-            queue_name=os.getenv("CSTAR_QUEUE_NAME"),
-            walltime="00:10:00",  # TODO how to determine this one?
+            queue_name=os.getenv("CSTAR_SLURM_QUEUE"),
+            walltime=os.getenv(
+                "CSTAR_SLURM_MAX_WALLTIME", "48:00:00"
```
ankona, line 174: at this location, i think we should not have a default. it should be the "environment loading component"'s job to ensure defaults are loaded instead. Why not put these defaults in `CStarEnvironmentManager`?

- consider using the `_SystemContext` classes to set system-specific defaults that can override the global defaults?

cstar/cli/template/actions/create.py
```diff
@@ -10,31 +10,17 @@
...
-        if "y" in do_create.lower():
-            path.mkdir(parents=True, exist_ok=True)
-        else:
-            msg = "Unable to create template without valid directory"
-            raise ValueError(msg)
-
     if path and not path.exists():
-        msg = f"The specified directory does not exist: {path}"
-        raise ValueError(msg)
+        path.mkdir(parents=True, exist_ok=True)
```
ankona: - this method can go away completely since it really existed to hold the logic for performing the check and allowing interactive use. 
- if path and not path.exists() doesn't make sense 
  - you're passing `exist_ok=True`. 
  - consider `if path: path.mkdir(...)`

cstar/tests/unit_tests/execution/test_handler.py
```diff
@@ -196,26 +194,13 @@ def test_updates_indefinite_with_seconds_param_0(
...
-                mock_input.assert_called_once_with(
-                    "This will provide indefinite updates to your job. You can stop it anytime using Ctrl+C. "
-                    "Do you want to continue? (y/n): "
-                )
-            # Patch input to simulate the confirmation prompt
-            with patch("builtins.input", side_effect=["n"]) as mock_input:
-                with patch("builtins.open", create=True) as mock_open:
-                    handler.updates(seconds=0)
-                    mock_open.assert_not_called()
+            # Replace `time.sleep` with a side effect that raises KeyboardInterrupt
```
ankona: Do we really need this sentence explaining what the patch method does?

README.md
```diff
@@ -55,6 +55,22 @@ pytest
...
+| CSTAR_SLURM_QUEUE            | None (must be set for SLURM usage) | The SLURM queue or partition to use for jobs.                                                                                                                                                                     |
+| CSTAR_SLURM_MAX_WALLTIME     | "48:00:00"                         | Maximum walltime to set for jobs submitted to SLURM.                                                                                                                                                              |
+| CSTAR_RUNID                  | None, set by CLI                   | The run ID should be unique to a given "run" of an orchestration as it controls caching/restoring previous workflow state.                                                                                        |
+| CSTAR_CMD_CONVERTER_OVERRIDE | None                               | Testing only. If set, submit a custom command as the execution command to SLURM jobs, instead of the default application command.                                                                                 |
```
ankona: One thing we might consider (now or later...) is that we can separate and use a different convention for internal-only vars.

- For example, `CSTAR_CMD_CONVERTER_OVERRIDE` isn't meant to be used by clients, so we might name it `CSTARINT_xxx` or more simply `cstar_cmd_...`

This doc is great, but the main idea would be that we can document those internal env vars separately to stop clients from needing to read about things we don't really want them to use.

README.md
```diff
@@ -55,6 +55,22 @@ pytest
 
 To learn how to use `C-Star`, check out the [documentation](https://c-star.readthedocs.io/en/latest/index.html).
 
+# Environment Variables
+The following environment variables control C-Star's behavior:
+
+| Variable                     | Default                            | Effect                                                                                                                                                                                                            |
+|------------------------------|------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
ankona: unless you're using some formatting (centering, etc), you might use this to accomplish the same thing and avoid text wrap:

`|-|-|-|`

README.md
```diff
@@ -55,6 +55,22 @@ pytest
...
+| CSTAR_NPROCS_POST            | os.cpu_count() / 3                 | How many parallel processes to use for join operations post-run.                                                                                                                                                  |
```
ankona:
```suggestion
| CSTAR_NPROCS_POST            | os.cpu_count() / 3                 | The number of parallel processes to use for post-run join operations.                                                                                                                                                  |
```

README.md
```diff
@@ -55,6 +55,22 @@ pytest
...
+| CSTAR_SLURM_ACCOUNT          | None (must be set for SLURM usage) | The account name to be passed to SLURM for compute accounting.                                                                                                                                                    |
+| CSTAR_SLURM_QUEUE            | None (must be set for SLURM usage) | The SLURM queue or partition to use for jobs.                                                                                                                                                                     |
+| CSTAR_SLURM_MAX_WALLTIME     | "48:00:00"                         | Maximum walltime to set for jobs submitted to SLURM.                                                                                                                                                              |
+| CSTAR_RUNID                  | None, set by CLI                   | The run ID should be unique to a given "run" of an orchestration as it controls caching/restoring previous workflow state.                                                                                        |
```
ankona: Let's not re-define runs in this table. Add a link to the workplan docs?

```suggestion
| CSTAR_RUNID                  | None, set by CLI                   | The unique identifier for an execution of a [Workplan](https://docs.com/workplans/#specifying-a-run-id).                                                                                        |
```

cstar/cli/blueprint/actions/run.py
```diff
@@ -22,8 +23,8 @@ def configure_simulation_runner(path: Path) -> SimulationRunner:
     SimulationRunner
         A simulation runner configured to execute the blueprint
     """
-    account_id = "m4632"
-    walltime = "48:00:00"
+    account_id = os.getenv("CSTAR_SLURM_ACCOUNT", "")
+    walltime = os.getenv("CSTAR_SLURM_WALLTIME", "48:00:00")
```
ankona, line 27: consider performing all the env configuration in a single place, the entrypoint `cli.py`

cstar/tests/unit_tests/execution/test_handler.py
```diff
@@ -268,8 +253,7 @@ def append_updates_and_change_status():
         updater_thread.start()
 
         # Run the `updates` method
```
ankona: i am so embarrassed that i left this comment. delete it and save me from myself.

cstar/cli/template/actions/create.py
ScottEilerman: removed function and put the mkdir in the right place to replace it.

README.md
ScottEilerman: Pycharm did all the table formatting and adjusted it for me. Pretty nice if you're using Pycharm, but I agree it's excessive if not. I trimmed the headers and the last column. FYI apparently Github's markdown requires at least three `---` for each column. Boy markdown is fun when everyone has different standards.

README.md
ScottEilerman: There is no documentation of workplans on our open github yet. 🙃  We'll clean it up before release.

cstar/tests/unit_tests/execution/test_handler.py
ScottEilerman: removed

cstar/roms/simulation.py
ScottEilerman: I'm not opposed to it, but am going to leave it for now, with the reasons being: (a) I can imagine a not-too-farfetched use case where someone would want to control these behaviors separately, (b) probably one or both of these go away or change defaults once we get path management in place, and (c) I want to see the full array of options we generate by the end of this before deciding where to consolidate.

README.md
ScottEilerman: agree in principle. user-facing stuff should probably end up in RTD whenever we do that documentation push, and internal elsewhere. really just collecting everything _somewhere_ for now.

cstar/orchestration/launch/slurm.py
ScottEilerman: per PR description, I'm considering this out of scope for now. this PR is a small iteration to get things a little cleaner and bypass hardcodes so I can actually run it on anvil.

cstar/cli/blueprint/actions/run.py
ScottEilerman: per PR description, I'm considering this out of scope for now. this PR is a small iteration to get things a little cleaner and bypass hardcodes so I can actually run it on anvil.

PR #376, opened 2025-12-02
suggested fix for post-run failure

ScottEilerman:
Haven't tested this against your use case, but it seems the problem is the worker relies only on the status of the ROMS execution handler in determining its final disposition, so something going wrong in post-run doesn't change the disposition. This ought to work, though there are probably cleaner ways of doing it.


ankona:
nifty. good starting point.

PR #377, opened 2025-12-02
Vertical mixing fix

ScottEilerman:
When we read in the ROMS runtime settings, if the tracer vertical mixing list is all zeros but too short, pad it with more zeros. If any of the values are non-zero, don't change anything, and leave it up to the user to fix the file if they don't have enough values (which will cause ROMS to barf).

This is a convenience feature for people who are using older roms.in files, or want want to use the background vertical mixing parameter but not the tracer ones (and maybe don't put enough zeros in the list). There are a lot of smarter ways we could probably do this (checking .opt files for the real number of tracers) but this at least makes our current codebase and test files play nicely with the latest commits on main for ucla-roms.

### DIFF
--- a/cstar/roms/runtime_settings.py
+++ b/cstar/roms/runtime_settings.py
@@ -422,7 +422,9 @@ class VerticalMixing(ROMSRuntimeSettingsSection):
     @classmethod
     def fill_tracers(cls, value):
         list_len = len(value)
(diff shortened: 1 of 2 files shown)


dafyddstephenson (approved):
Looks good. I have a feeling this isn't needed after all, as that block is basically never going to be read in any of our simulations if I understand the situation correctly, but it's worth having for completeness.

PR #381, opened 2025-12-19
CSD 44

ankona:
- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


cstar/base/utils.py
```diff
@@ -277,3 +279,53 @@ def _run_cmd(
...
+        The slugified version of the source string.
+    """
+    if not source:
+        raise ValueError
+
+    return re.sub(r"\W+", "-", source.casefold())
+
+
+def deep_merge(d1: dict[str, t.Any], d2: dict[str, t.Any]) -> dict[str, t.Any]:
+    """Deep merge two dictionaries.
```
ScottEilerman, line 340: It might be good to add in the docstring the data structures that are handled (dicts and lists, but not sets, for example), and also note that d2 overwrites d1 for values present in both that aren't explicitly handled. We _shouldn't_ go down the path of trying to handle everything.

cstar/cli/workplan/run.py
```diff
@@ -6,13 +6,17 @@
 import typer
 
 from cstar.orchestration.dag_runner import build_and_run_dag
+from cstar.orchestration.orchestration import ENV_CSTAR_RUNID
```
ScottEilerman: I think this import should be `from cstar.orchestration.utils` (unless there was some reason to have it be second-hand).

cstar/cli/workplan/run.py
```diff
@@ -6,13 +6,17 @@
...
 
 app = typer.Typer()
 
 
 @app.command()
 def run(
     path: t.Annotated[Path, typer.Argument(help="Path to a workplan file.")],
+    output_dir: t.Annotated[
+        Path, typer.Argument(help="Path to a directory where outputs will be written.")
+    ],
```
ScottEilerman: Leaving a comment here for now. I started thinking about if there's a sensible default for this (current directory, or something like `~/cstar/`), then I started pulling on the thread of what this actually controls (just where workplan files get written, or is it the whole ROOT_DIR for the file system class, or is it something else) and didn't quite get to my answer. I'm going to keep reviewing for now and I'm guessing I'll see it come together.

cstar/entrypoint/worker/worker.py
```diff
@@ -150,35 +149,7 @@ def _prepare_file_system(self) -> None:
...
-        #     self.log.debug(msg)
-        #     shutil.rmtree(externals_path)
-        # externals_path.mkdir(parents=True, exist_ok=False)
-
-        # create a clean location to write outputs.
-        if not self._output_dir.exists():
-            msg = f"Creating clean output dir: {self._output_dir}"
-            self.log.debug(msg)
-            self._output_dir.mkdir(parents=True, exist_ok=True)
+        ...
```
ScottEilerman: Time to remove this method? Or are you thinking we'll want to override it sometime?

cstar/entrypoint/worker/worker.py
```diff
@@ -496,21 +456,29 @@ async def main(raw_args: list[str]) -> int:
         The exit code of the worker script. Returns 0 on success, 1 on failure.
     """
     try:
+        raw_args = sys.argv[1:]
         parser = create_parser()
         args = parser.parse_args(raw_args)
```
ScottEilerman: nit, any reason I don't know about for `raw_args`, or should we just reduce to
```python
        parser = create_parser()
        args = parser.parse_args()
```
?

cstar/entrypoint/worker/worker.py
```diff
@@ -496,21 +456,29 @@ async def main(raw_args: list[str]) -> int:
...
         return 1
     else:
         service_cfg = get_service_config(args)
         blueprint_req = get_request(args)
-        job_cfg = JobConfig()  # use default HPC config
+        job_cfg = JobConfig(
+            account_id=os.environ.get("CSTAR_SLURM_ACCOUNT", ""),
+            walltime=os.environ.get("CSTAR_SLURM_MAX_WALLTIME", "01:00:00"),
+            priority=os.environ.get("CSTAR_SLURM_QUEUE", ""),
+        )
```
ScottEilerman: We can pull these magic strings from your consolidated literals now!

I also listed the default walltime as `48:00:00`. We haven't really had a discussion about what it should be, but I'd argue that the maximum (which is actually 96 for anvil, I think, but also depends on queue...) makes sense and setting it lower is really to optimize your queue time at the risk of setting it _too_ low and killing your job before it's done.

cstar/orchestration/utils.py
```diff
@@ -1,26 +1,20 @@
 import os
-import re
 import shutil
+import typing as t
 from pathlib import Path
 
+ENV_CSTAR_ORC_OUTDIR: t.Literal["CSTAR_ORC_OUTDIR"] = "CSTAR_ORC_OUTDIR"
```
ScottEilerman: This might be personal preference: `ORCH` reads to me as `orchestration` but my brain didn't immediately know what to do with `ORC`. I got there, but it took a minute. Maybe I just consume too much fantasy media. I'd advocate for adding the `H` to all of these for clarity.

cstar/execution/file_system.py
```diff
@@ -0,0 +1,127 @@
...
+    def prepare(self) -> None:
+        """Construct the directory tree for the job."""
+        for task_dir in self._dir_set():
+            task_dir.mkdir(parents=True, exist_ok=True)
+            self.log.debug(f"Created {task_dir.name} directory `{task_dir}`")
+
+    def clear(self) -> None:
+        """Ensure the job's working directory is empty."""
+        shutil.rmtree(self.root)
+        self.root.mkdir(parents=True)
```
ScottEilerman, line 80: nit, only since you already did it in `prepare`, a log message here might be very helpful someday.

cstar/execution/file_system.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class RomsJobFileSystem(JobFileSystem):
+    COMPILE_TIME_NAME: Literal["roms/compile_time_code"] = "roms/compile_time_code"
+    RUNTIME_NAME: Literal["roms/runtime_code"] = "roms/runtime_code"
+    INPUT_DATASETS_NAME: Literal["roms/input_datasets"] = "roms/input_datasets"
+    CODEBASES_NAME: Literal["roms/codebases"] = "roms/codebases"
```
ScottEilerman: Discussion topic: Is the `roms` directory doing any critical work here or should we drop it? I know it's inherited from the existing structure, not something you added, but I'm open to culling it now if it's extraneous. We're already inside a step, which has a defined application... I guess you could have multiple structures like this under an application step, but seems somewhat unlikely. We do already have MARBL and it ends up _under_ `roms/codebases`. I think I'm in favor of ditching it unless you see a good reason to keep it.

cstar/execution/handler.py
```diff
@@ -153,24 +154,31 @@ def updates(self, seconds: float = 10) -> None:
...
+                self.log.info(msg)
+                return
+
         try:
             with open(self.output_file) as f:
                 f.seek(0, 2)  # Move to the end of the file
                 start_time = time.time()
                 while seconds == 0 or (time.time() - start_time < seconds):
-                    line = f.readline()
+                    if self.output_file.exists():
```
ScottEilerman, line 174: Is this `if` unnecessary given the addition above (line 164)? Or a safeguard against the file getting deleted mid-read (which would be problematic on a number of levels).

cstar/orchestration/dag_runner.py
```diff
@@ -113,36 +128,87 @@ async def process_plan(orchestrator: Orchestrator, mode: RunMode) -> None:
...
         sleep_duration = next(delay_iter)
-        print(f"Sleeping for {sleep_duration} seconds before next {mode}.")
+        print(f"Next `{mode}` check in {sleep_duration:4.1f} seconds.")
         await asyncio.sleep(sleep_duration)
 
     print(f"Workplan {mode} is complete.")
 
 
+async def prepare_workplan(
+    wp_path: Path, output_dir: Path, run_id: str
```
ScottEilerman: run_id isn't currently used, not sure if it should get used, or should be dropped.

cstar/orchestration/dag_runner.py
```diff
@@ -113,36 +128,87 @@ async def process_plan(orchestrator: Orchestrator, mode: RunMode) -> None:
...
+        transformer = WorkplanTransformer(wp_og, RomsMarblTimeSplitter())
+        wp = transformer.apply()
+
+        if transformer.is_modified:
+            print("A time-split workplan will be executed.")
+    else:
+        wp = wp_og
+
+    # make a copy of the original and modified blueprint in the output directory
+    persist_og = WorkplanTransformer.derived_path(wp_path, output_dir, "_og")
```
ScottEilerman: I was thinking about this section for a bit and whether we should move some of this into the `if` block so that we don't write out duplicate files if the feature is turned off. I think I decided I don't care. What I did realize though, which made me laugh because my brain is so used to seeing/using `og` like this, is that we probably should professionalize `_og` -> `_orig` so we don't have to explain what OG means someday when these are part of an audit trail. 😂

cstar/orchestration/dag_runner.py
```diff
@@ -152,31 +218,120 @@ async def build_and_run_dag(path: Path) -> None:
...
+        type=Path,
+        help="The path where outputs must be written",
+    )
+    # parser.add_argument(
+    #     "-r",
+    #     "--reset-base",
+    #     required=True,
+    #     type=str,
+    #     help="The basename of the reset files output by the simulation (e.g. output_rst).",
+    # )
```
ScottEilerman: remove

cstar/orchestration/dag_runner.py
```diff
@@ -113,36 +128,87 @@ async def process_plan(orchestrator: Orchestrator, mode: RunMode) -> None:
...
     """Execute the steps in the workplan.
 
     Parameters
     ----------
     path : Path
         The path to the blueprint to execute
+    output_dir : Path
+        The path to the output directory.
+    reset_base : str
+        The basename of the reset files output by the simulation.
```
ScottEilerman: remove reset_base from docstring

ScottEilerman (commented):
I got halfway through this before break but forgot to post what I had. No working during PTO!

cstar/cli/workplan/run.py
ankona: Great eye! Fixed.

cstar/entrypoint/worker/worker.py
ankona: I was trying to avoid changing the tests until the last minute and forgot it was leftover. Great catch. Removed.

cstar/entrypoint/worker/worker.py
ankona: It was really only used for debugging (e.g. conditional breakpoint looking for an argument to be provided) but can be removed, now. Done.

cstar/orchestration/utils.py
ankona: Good eye. My lack of MMORPG cred becomes glaringly obvious! Added `H`, as needed.

cstar/execution/handler.py
ankona: The total duration can be exceeded in the first loop if the task sits in `pending` status for a decent duration, which is caught by the file DNE check.

On subsequent iterations, the `if _status == PENDING` block isn't entered and the second loop can vomit. 

Shifting the output file check after the first loop would alleviate the need to check twice, but it also changes the behavior of the function in an unintended way - it no longer waits for the user-supplied number of seconds. This broke some tests so I err'd on the side of keeping the `ExecutionHandler` behavior consistent.

cstar/orchestration/dag_runner.py
ankona: Good eye. Removed unused parameter from `prepare_workplan`

cstar/orchestration/dag_runner.py
ankona: Good eye. Removed.

cstar/orchestration/dag_runner.py
ankona: Good eye. Removed.

cstar/entrypoint/worker/worker.py
ankona: w00t! replaced w/values from consolidated literals.

cstar/execution/file_system.py
ankona: Agreed. Added.

cstar/orchestration/dag_runner.py
ankona: Agreed. Replaced informal `_og` with clearer `_orig`, as needed.

cstar/base/utils.py
ankona: Improved docstring clarity. Great idea.

cstar/orchestration/launch/slurm.py
```diff
@@ -127,6 +142,45 @@ def convert_step_to_placeholder(step: Step) -> str:
...
+        """Get the account to use for SLURM jobs.
+
+        Read from the environment variable `CSTAR_SLURM_ACCOUNT`.
+
+        Returns
+        -------
+        str
+            The account to use for SLURM jobs.
+        """
+        return os.getenv("CSTAR_SLURM_ACCOUNT") or ""
```
ScottEilerman: I like these methods, good placeholders that'll be easy to swap when we evolve our config management.

cstar/orchestration/models.py
```diff
@@ -439,23 +459,104 @@ class Step(BaseModel):
...
+            The blueprint instance loaded by the step
+
+        Returns
+        -------
+        RomsJobFileSystem
+        """
+        return RomsJobFileSystem(self.working_dir(bp))
+
+
+class ChildStep(Step):
```
ScottEilerman, line 521: I think this is fine for now, but want to plant a seed in the back of your head that we might have a domain language conflict in a few months. When we start building workflows for nesting, the terminology "parent grid" and "child grid" is going to come into play a lot and we are probably going to have transforms relating to corresponding parent/child parts of the workflow. We might consider `SubStep` here? I don't like it as much, so maybe I'm inclined to leave it as `ChildStep` for now and only change it if it actually becomes a problem instead of guessing.

ScottEilerman: I guess I referred to these as "tasks" in the design doc but that's pretty overloaded with Prefect meaning.

cstar/roms/runtime_settings.py
```diff
@@ -38,6 +40,9 @@
 MIN_NUM_TRACERS = 37
 """Vertical mixing requires enough values for all tracers or it raises an error."""
 
+DEFAULT_OUTPUT_ROOT_NAME: Literal["output"] = "output"
```
ScottEilerman: The lovely thing about having this defined here is that I can easily suggest something like "can we call this `roms_output`" so that it doesn't get confused with the `output` layer of the FileSystem structures, and because it is specifically where ROMS itself is putting outputs (as opposed to our post-processed joined outputs). I guess that was less of an issue when it was also under a `ROMS` directory layer.

cstar/roms/simulation.py
```diff
@@ -1173,25 +1180,28 @@ def setup(
...
-                codebase_dir.mkdir(parents=True, exist_ok=True)
-                codebase.get(codebase_dir)
-                os.environ[codebase.root_env_var] = str(codebase_dir)
-            codebase.setup()
+            codebase_dir = codebases_dir / codebase.root_env_var.split("_")[0]
+            if os.getenv("CSTAR_FRESH_CODEBASES", "0") == "1" and codebase_dir.exists():
+                shutil.rmtree(codebase_dir)
+
+            codebase_dir.mkdir(parents=True, exist_ok=True)
+            codebase.setup(codebase_dir)
```
ScottEilerman, line 1210: nice subtle change here, seems like we were calling `get` twice before.

cstar/tests/unit_tests/roms/test_roms_runtime_settings_fixed_output.py
```diff
@@ -0,0 +1,29 @@
...
+
+@pytest.mark.parametrize(
+    "output_root_name_input",
+    [
+        "output",
+        "another_name",
+        "a_third_name",
+    ],
+)
+def test_fixed_output_root_name(tmp_path, output_root_name_input):
```
ScottEilerman: should this get moved into `test_roms_runtime_settings.py`? I'm guessing it was a quick temporary thing to make it in its own file.

cstar/roms/runtime_settings.py
ScottEilerman: I noticed this ends up in a bunch of test outputs so you'd need to update it there too (or have those import this variable and use it)

ScottEilerman (commented):
Another small batch of comments to keep things moving while I go do meetings. I think I've reviewed all of the non-test modules except for transforms.py now.

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+        """
+        ...
+
+
+TRANSFORMS: dict[str, list[Transform]] = defaultdict(list)
+"""Storage for transform registrations."""
+
+
+def register_transform(application: str, transform: Transform) -> None:
+    """Register a splitter for an application.
```
ScottEilerman: docstring 
```suggestion
    """Register a transform for an application.
```

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+
+    _transformed: Workplan | None = None
+    """The post-transformation workplan."""
+
+    DERIVED_PATH_SUFFIX: t.Literal["_trx"] = "_trx"
+    """Suffix appended to the original workplan path when generating a derived path."""
+
+    def __init__(self, wp: Workplan, transform: Transform) -> None:
+        """Initialize the instance."""
+        self.original = Workplan(**wp.model_dump(by_alias=True))
```
ScottEilerman, line 189: I feel like we talked about this sometime last year, but remind me... why not `self.original = wp`? Is this effectively a deepcopy, or is there something about the alias behavior that we need to invoke here?

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+        return directory / filename
+
+    def apply(self) -> Workplan:
+        """Create a new workplan with appropriate transforms applied.
+
+        Returns
+        -------
+        Workplan
+        """
+        if not is_feature_enabled("ORC_TRANSFORM_AUTO"):
```
ScottEilerman: missed an ORC-->ORCH (and then I did a search and there's another one in dag_runner.py)

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+            return self.original
+
+        if self._transformed:
+            return self._transformed
+
+        steps = []
+        for step in self.original.steps:
+            if not self.transform_fn:
```
ScottEilerman: I was wondering if this should be an `or` with the first check (the feature flag), but after looping back to look where we set this in init, I'm not sure we'd ever hit it (unless someone passes None or some other false-y transform), so maybe this is an old use case? Or maybe I'm not getting something.

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+            transformed_steps = list(self.transform_fn(step))
+
+            # replace dependencies on the original step with the last transformed step
+            tweaks = (s for s in self.original.steps if step.name in s.depends_on)
+            for tweak in tweaks:
+                tweak.depends_on.remove(step.name)
+                tweak.depends_on.append(transformed_steps[-1].name)
+
+            overridden_steps: list[Step] = []
+            override_transform = OverrideTransform()
```
ScottEilerman, line 284: I'm thinking the override portion should maybe be its own method or get called elsewhere, as I view that as a core component of workplans rather than an application-specific optional feature. We could make this a follow-up ticket though if it's gonna muck up the flow of this method and how the updates all tie in to each other.

Or maybe this ties in to the previous comment about what code gets executed if transforms are disabled or not provided?

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+        self._transformed = Workplan(**wp_attrs)
+        return t.cast(Workplan, self._transformed)
+
+
+class RomsMarblTimeSplitter(Transform):
+    """A step tranformation that splits a ROMS-MARBL simulation into
+    multiple sub-steps based on the timespan covered by the simulation.
+    """
+
+    OUTPUT_ROOT_DEFAULT: t.Literal["output"] = "output"  # see also: RomsRuntimeSettings
```
ScottEilerman: I just _had_ to make a comment about how this would be easy to change since it's only defined in one place, huh?

I guess we don't want to import because it feels like a could lead to some circular import nonsense?

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+        job_fs = step.file_system(blueprint)
+        job_fs.prepare()
+
+        serialize(job_fs.root / Path(step.blueprint_path).name, blueprint)
+
+        time_slices = list(get_time_slices(start_date, end_date, self.frequency))
+        n_slices = len(time_slices)
+
+        # if (start_date - end_date).total_seconds() < timedelta(days=30).total_seconds():
+        #     # TODO: leave this ask discussion driver. must determine
```
ScottEilerman: Is there still discussion to be had here? I thought this condition was taken care of already.

cstar/orchestration/launch/slurm.py
ankona: Noticed my failure to use the constants and replaced hardcoded strings here w/constants

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+                    exclude_unset=True,
+                    exclude_defaults=True,
+                    exclude_computed_fields=True,
+                ),
+            )
+
+            compact_sd = sd.strftime("%Y%m%d%H%M%S")
+            compact_ed = ed.strftime("%Y%m%d%H%M%S")
+
+            dynamic_name = f"{i + 1:02d}_{step.safe_name}_{compact_sd}_{compact_ed}"
```
ScottEilerman: nit, I'd probably up the padding to three digits (to accommodate e.g. daily for a year, or monthly for 10 years)

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+            child_step_name = slugify(dynamic_name)
+
+            subtask_root = (
+                job_fs.tasks_dir
+            )  # subtask_fs = RomsJobFileSystem(subtask_root)
```
ScottEilerman: nit, comment cleanup

cstar/orchestration/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+                    "end_date": ed,
+                    "output_dir": subtask_out_dir.as_posix(),
+                },
+            }
+
+            if last_restart_file:
+                rst_path = last_restart_file.as_posix()
+                overrides["initial_conditions"] = {"data": [{"location": rst_path}]}
+
+            child_bp_path = subtask_out_dir / f"{child_step_name}bp.yaml"
```
ScottEilerman:
```suggestion
            child_bp_path = subtask_out_dir / f"{child_step_name}_bp.yaml"
```
should there be a `_`?

cstar/tests/unit_tests/orchestration/test_transforms.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+
+    bp_content = bp_tpl_path.read_text()
+    bp_content = bp_content.replace(
+        "output_dir: .", f"output_dir: {test_output_dir.as_posix()}"
+    )
+    test_bp_path.write_text(bp_content)
+
+    wp_tpl_path = (
+        Path(__file__).parent.parent.parent.parent
```
ScottEilerman: ...what if we set the `tests` root as a global in the highest-level conftest.py? Or the `additional_files` dir. So you don't have to know how many parents to add on these things. I know you mentioned something like this in the past. Can be a spin-off issue if you don't want to muck with it now.

ScottEilerman (commented):
One or two questions in transforms.py, we can chat about those whenever. Maybe we avoid any structural changes for this PR though? Otherwise a few nits. I didn't review the tests closely, but one or two things jumped out at me from a really quick skim.

cstar/roms/runtime_settings.py
ankona: I updated tests using both the roms default root name and the output directory to avoid hardcoded strings in both cases

cstar/tests/unit_tests/roms/test_roms_runtime_settings_fixed_output.py
ankona: Yes! Moved to correct location

cstar/orchestration/transforms.py
ankona: Yes, I'm just ensuring i have a deep copy so no changes in the original cannot effect the transform

cstar/orchestration/transforms.py
ankona: nit is picked!

cstar/orchestration/transforms.py
ankona: W00t. good eye. fixed.

cstar/orchestration/transforms.py
ankona: mos'def. done!

cstar/orchestration/transforms.py
ankona: good eye! done.

cstar/orchestration/transforms.py
ankona: I saw the same "failure to re-use existing constant" issue - probably as you were writing this comment.  You are correct that there _was_ an issue with circular references; I've resolved it and we are down to a single change for `output_root_name` tweaks.

cstar/tests/unit_tests/orchestration/test_transforms.py
ankona: there's a saying about this... forests and trees are involved. TYVM and done!

cstar/orchestration/transforms.py
ankona: Derp. Fixed!

cstar/roms/simulation.py
ankona: 💯

cstar/orchestration/transforms.py
ankona: done.

cstar/cli/workplan/run.py
ScottEilerman: Revisiting this... my understanding is that this `output_dir` gets put into an environment variable and then picked up by `Step.working_dir()`, where (unless the workplan has overrides specified), it gets the `run_id` appended and eventually becomes the root directory for `RomsFileSystemManager`. That becomes the location for all inputs/outputs related to the entire workplan. Let me know if any of that is incorrect.

I think having the CLI override option is good, but we can still implement a default that ends up similar to the [design doc](https://github.com/CWorthy-ocean/cworthy-design-docs/blob/main/design_docs/cwep0001-blueprints-workplans.md#storage-and-predictable-work-plan-directory-structures) (which is sort of outdated, admittedly). I'd propose we add another user-settable env-var `CSTAR_ROOT_DIR` that defaults to `~/cstar`. In practice, I'd recommend everyone on Anvil actually set this in their `.bashrc` to a location of choice on the scratch drive, for performance. Then we can additionally inject the slugified run plan name in the path, so that without specifying something on the CLI, if you run `cstar workplan run my_workplan.yaml --run_id this_unique_id`, results would go into: `/anvil/scratch/x-seilerman/cstar/my_workplan/this_unique_id/....`.

Open to other thoughts, or spinning this off into a follow-up. The hope is to set an output path once and then never worry about it again because the system will organize things for you.

cstar/orchestration/transforms.py
ankona: I think this is a leftover from the code evolving. As-is, you're 100% correct that the transform function must be passed and this conditional is useless. Removed.

ScottEilerman (approved):
I believe everything has been addressed. Great work on this! Feel free to merge when you're ready.

cstar/execution/file_system.py
ankona: Removed per offline discussion.

cstar/orchestration/transforms.py
ankona: I've created a follow-up ticket to fix this bug (the overrides are not applied when the transforms feature-flag is disabled - CSD-404)

cstar/orchestration/transforms.py
ankona: I've created a follow-up ticket (CSD-405) for discussion during grooming.

PR #382, opened 2025-12-24
Orchestration

matt-long:
The `CodeRepository` was internally inconsistent, require either `branch` or `commit`, not both — but each also having a minimum length of 1. This PR fixes that issue.

### DIFF
--- a/cstar/orchestration/models.py
+++ b/cstar/orchestration/models.py
@@ -113,7 +113,7 @@ class ForcingConfiguration(ConfiguredBaseModel):
 
     corrections: Dataset | None = Field(default=None, validate_default=False)
     """Wind or other forcing corrections."""
-
+print("ForcingConfiguration") 
 
 class CodeRepository(DocLocMixin, ConfiguredBaseModel):
     """Reference to a remote code repository with optional path filtering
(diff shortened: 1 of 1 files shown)


ScottEilerman (approved):
This should work as-is if you omit the unused field entirely, but I can see where it could potentially cause issues if you are initializing or manipulating from code rather than yaml (or if you want to keep both fields in a template but only fill one, for example). I don't think we need to be this strict with validating min_length at the moment, so if this change helps your use case, it's fine by me!

matt-long:
I found this necessary to avoid validation errors like this
```
ValidationError: 4 validation errors for RomsMarblBlueprint
code.roms.branch
  String should have at least 1 character [type=string_too_short, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/string_too_short
code.run_time.commit
  String should have at least 1 character [type=string_too_short, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/string_too_short
code.compile_time.commit
  String should have at least 1 character [type=string_too_short, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/string_too_short
code.marbl.branch
  String should have at least 1 character [type=string_too_short, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/string_too_short
```

PR #384, opened 2026-01-21
pin roms-tools to 3.3.0

matt-long:


pyproject.toml
```diff
@@ -33,15 +33,15 @@ dependencies = [
...
+    "roms_tools[dask]==3.3.0",
     "typer>=0.17.5",
 ]
 keywords = ["MCDR", "CDR", "ocean carbon", "climate"]
 
 [project.optional-dependencies]
 test = [
      "pytest>=7.0",
-     "roms_tools[dask]==3.1.2"
+     "roms_tools[dask]==3.3.0"
```
ankona: We seem to have duplicated the dependency unnecessarily. Can we remove this copy in the `test` optional deps instead of updating it?

ankona (changes requested):
Let's get rid of the duplicated dep and this will be good to go.

cstar/system/scheduler.py
```diff
@@ -450,7 +450,8 @@ class SlurmScheduler(Scheduler):
     @property
     def in_active_allocation(self) -> bool:
         """Determine whether the process is in an active allocation by looking for the SLURM_JOBID environment variable."""
-        return os.getenv("SLURM_JOBID") is not None
+        # TODO: Implement this properly
+        return False #os.getenv("SLURM_JOBID") is not None
```
ankona: I think you may have added an unintended change to this PR for pinning the version of the ROMS dependency.

matt-long:
Sorry for the stray commits — I have backed them out and will address separately. I think this is good to go.

Issue #385, opened 2026-01-22
`in_active_allocation()` mis-detects allocations on JupyterHub because `SLURM_JOBID` is set for the hub session


matt-long:
The function `in_active_allocation()` currently attempts to determine whether the process is running inside an active Slurm allocation by checking for the presence of the `SLURM_JOBID` environment variable:
https://github.com/CWorthy-ocean/C-Star/blob/5964251e977cc61f8c1276367c067b5d19e12167/cstar/system/scheduler.py#L453C21-L453C23

This heuristic breaks in managed JupyterHub environments on HPC systems, where the Jupyter server itself is often launched via Slurm and therefore inherits `SLURM_JOBID`.

As a result, C-Star incorrectly treat the environment as “in an active allocation” and take the wrong execution path.

dafyddstephenson:
related: https://github.com/CWorthy-ocean/C-Star/pull/195
C-Star previously dealt with this by clearing any `SLURM_` prefixed environment variables except those that are always needed (see above PR)
I'm not sure if this practice has been maintained in the current code @ScottEilerman ? Either way it is an easy fix

ScottEilerman:
The current intended execution path is to instantiate C-star via the CLI on a login node -- this does indeed break if someone is running it from a compute node (via Forge or otherwise). We'll need to think a little bit about the right way to handle this.

PR #387, opened 2026-01-24
Temporary fix for `in_active_allocation` & correct lmod/load_env for Perlmutter build

matt-long:
This provides at least a temporary workaround for #385:
- running within JupyterHub is detected by checking a series of related environment variables
- C-Star can set an environment variable `CSTAR_IN_ACTIVE_ALLOCATION=1` to override
- If none of the above, the old behavior is preserved

Also, I discovered that I could not build the model on Perlmutter. The `load_lmod_modules` call was happening after the `_load_env`, which is not the correct order. This changes the order, loads the modules first — and now the build on Perlmutter works.


cstar/system/environment.py
```diff
@@ -136,9 +141,18 @@ def _load_env(self) -> dict[str, str]:
...
+        # This allows env file to reference variables set by modules (which are loaded first)
+        expanded_vars = {}
+        for key, value in env_vars.items():
+            # Use os.path.expandvars to expand ${VAR} and $VAR syntax
+            expanded_value = os.path.expandvars(value)
+            expanded_vars[key] = expanded_value
+
+        os.environ.update(expanded_vars)
+
+        return expanded_vars
```
ScottEilerman, line 155: I'm curious about these env changes. It's odd that it worked ok in the past, and on Anvil (where the env file also has lmod-reliant paths). My assumption is that it all worked ok as long as you don't use any of these paths from python (because the shell would resolve them at time of execution even if they were put in out-of-order). Maybe Forge is relying on these differently? Anyway, what you have here seems more correct, so if it's working, we can go ahead and merge it.

cstar/system/scheduler.py
```diff
@@ -449,7 +449,20 @@ class SlurmScheduler(Scheduler):
...
+
+        jupyter_env_vars = (
+            "JUPYTERHUB_API_TOKEN",
+            "JUPYTERHUB_USER",
+            "JUPYTERHUB_SERVICE_PREFIX",
+            "JUPYTER_SERVER_ROOT",
+            "JUPYTER_SERVER_URL",
+        )
+        if any(k in os.environ for k in jupyter_env_vars):
+            return False
```
ScottEilerman, line 464: Have you run a test where you launch from a Jupyter environment and have it submit a the ROMS run to a slurm job? I'll admit I'm not 100% sure how machines/slurm handle the forwarding of env-vars from where the job gets submitted to the runtime environment. My recollection is that they carry over, and if so, I'd assume these Jupyter variables end up on the allocated compute nodes as well, which would be a problem. I could be wrong though, or it could be machine dependent? If it is a problem, a potential solution would be to add commands to `unset` them in the batch script.

ScottEilerman (commented):
Two questions, but tentative approval if everything is working as expected.

cstar/system/scheduler.py
matt-long: I have submitted and run C-Star jobs from JupyterHub (e.g., see [here](https://github.com/matt-long/cson-forge/tree/main/workflows/generate-models/outputs/forge-user/RCAC_anvil)).

I don't think the batch environment inherits the submitting environment — I see no linkage that would make that possible unless implemented explicitly: once the batch script enters the scheduler, it's self-contained. That said, it's definitely worth paying attention and verifying environments.

PR #388, opened 2026-01-26
Docs overhaul

ankona:
Updating documentation generation to include orchestration / workplan / blueprint content and shift focus on internal API. 

- [X] Closes #CSD-322
- [X] Tests passing
- [X] New functions/methods are listed in `api.rst`


.gitignore
```diff
@@ -139,3 +139,4 @@ dask-worker-space/
 
 # C-Star specific
 cstar/externals/
+/docs/generated/
```
ankona: Shouldn't this be relative `docs/generated`

docs/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -j auto
```
ankona, line 6: @ScottEilerman Did you notice after adding this that your local documentation builds consumed all CPU and made your machine unresponsive?

ScottEilerman: it spun up 10 workers but it was never unresponsive. it would do the build in maybe 10-30 seconds?

PR #401, opened 2026-01-30
Fix incorrect path to joined outputs

ankona:
# Fix incorrect path to joined outputs

This PR fixes a copypasta error resulting in the outputs from `ncjoin` being placed in the incorrect output directory.

# Changes

- Update `RomsFileSystemManager` to use `self.output_path` when calculating the joined output path

### DIFF
--- a/cstar/execution/file_system.py
+++ b/cstar/execution/file_system.py
@@ -145,7 +145,7 @@ class RomsFileSystemManager(JobFileSystemManager):
     @property
     def joined_output_dir(self) -> Path:
         """The directory for de-partitioned outputs."""
-        return self.root / self._JOINED_OUTPUT_NAME
+        return self.output_dir / self._JOINED_OUTPUT_NAME
 
     def codebase_subdir(self, key: str) -> Path:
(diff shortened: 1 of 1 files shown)


ScottEilerman:
It was semi-intentional to not put the joined outputs under `output`, because (a) I got scared about possible recursive wildcard operations in some variations of the post-processing code, and (b) the possibility that we may want to nuke the unjoined files after we join them, either as a part of the process, or some day many months from now to clear up some space. That said, I was never thrilled about my implementation or directory naming, and am definitely open to other ideas.

ankona:
> It was semi-intentional to not put the joined outputs under `output`, because (a) I got scared about possible recursive wildcard operations in some variations of the post-processing code, and (b) the possibility that we may want to nuke the unjoined files after we join them, either as a part of the process, or some day many months from now to clear up some space. That said, I was never thrilled about my implementation or directory naming, and am definitely open to other ideas.

Superb! I don't think this is necessary then and I made a mistake identifying non-existent bug.

Closing.

PR #403, opened 2026-02-03
Remove double slashes from anvil env vars

ankona:
# Double-slash fix

This PR fixes two values in `anvil.env` where repeated concatenation causes doubled forward slashes. 

For example, given `NETCDFHOME=${NETCDF}/`, `PATH=${PATH}:${NETCDFHOME}/bin` results in `PATH=...:<netcdfhome>//bin`

## Changes

- Remove two slashes from paths.

## Review Checklist

- [X] Tests passing


ScottEilerman (approved):
Approved because it's correct. It's weird to me that this wasn't breaking anything though. Or was it?

ankona:
> Approved because it's correct. It's weird to me that this wasn't breaking anything though. Or was it?

A posix compliant file system treats multiple successive slashes as a single slash (and it won't break). 

It is just "wrong" and is best removed to avoid confusion and keep the files clean.

PR #404, opened 2026-02-03
CSD-402: Mitigate git clone errors

ankona:
# Mitigate git clone errors

Repeated `git clone` operations when executing blueprints as part of an orchestrated workplan sometimes result in failures due to rate limiting.

## Changes

- Add utility methods for retrieving the C-Star `home` and `cache` directories
- Add a `Stager` sub-type `CachedRemoteRepositoryStager`
  - the stager writes unmodified repositories to the cache
  - after caching a copy, the stager copies the cloned repository to the target
- Modify repository `Retriever` to retry on clone failures.
  - add `LoggingMixin` to enable logging operations
- Update docstrings

## Review Checklist

- [X] Closes CSD-402
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/base/utils.py
```diff
@@ -24,34 +24,107 @@
...
 
+ENV_CSTAR_CACHEDIR: t.Literal["CSTAR_CACHEDIR"] = "CSTAR_CACHEDIR"
+"""Environment variable containing a path to the asset cache directory (user override)."""
 
-def get_output_dir(value_override: Path | None = None) -> Path:
-    """Return the path to the C-star output directory.
+
+def get_home_dir(value_override: Path | None = None) -> Path:
+    """Return the path to the C-star home directory.
+
```
ScottEilerman, line 42: nit: docstring for value_override param

cstar/io/stager.py
```diff
@@ -120,8 +121,38 @@ def stage(self, target_dir: "Path") -> "StagedRepository":
...
+        target_dir : Path
+            The local directory in which to stage the repository
+
+        Returns
+        -------
+        StagedRepository
+        """
+        cache_dir = get_cache_dir()
+        source_key = slugify(self.source.location.casefold().strip())
+        cache_path = cache_dir / source_key
```
ScottEilerman: I think we need a more complex cache key that incorporates the commit hash, right? Unless that gets wrapped into `source.location` somewhere. Otherwise you're going to cache roms once and the specific checkout target would get ignored. 

This actually gets tricky for branches I guess. If you only specify a branch in the blueprint, your cache could be old compared to the remote. I guess that's already sort of a problem for blueprints being reproducible, though I think the intention is that once we get to the point of implementing `locked`, you'd need to specify a commit and not a branch. But even if we consider unlocked, it's going to be a pain if we are frequently relying on something like `main`.

Options in order of complexity:
1. too bad so sad, clear your cache
2. disable caches for branches, only do it if you specify commit hashes
3. modify code such that if a branch is specified, we ping the remote to translate to a commit hash and actually pull/cache based on that
4. any other ideas?

I like 3 the best if it's not too much work. I assume we wouldn't get rate-limited on that query like we do on the clone.

_Someday_ we will additionally care about validating that the cached copy of the repo hasn't been tampered with for audit trail purposes, but we don't need to worry about that right now as we haven't implemented all the `locked` stuff yet.

ScottEilerman (changes requested):

cstar/io/stager.py
ankona: The more complex hash actually ruins the idea, here. By avoiding incorporation of the hash, the stager can return the same cached item to anybody requesting the repository.

The cached copy should _never_ be modified. After the stager clones the repo to the cache, unique changes/operations are only performed on a copy.

If we tried to do "smart clone", we hurt ourselves (e.g. `--depth 1` or `--single-branch`). That renders the cached copy only useful to users desiring the same branch.  We have to re-clone to achieve re-use. As-is, I clone once per repo, and can serve all subsequent requests.

I do think the flow could be improved by adding a `git pull` before copying to the target directory to ensure the cached copy is up-to-date if the cache is long-lived.

ankona: I've added a refresh for the cached clone. This ensures behavior pre- and post-cache are the same: both will use the latest copy of a branch.

3 is the right idea if we want to enable something like `cstar publish --run-id 12345`. We really should just store the hash in a local database as a column/attribute (use `git rev-parse HEAD` after clone or before a pull) instead of a convoluted retrieval process. This should be in a follow-up ticket.

cstar/base/utils.py
ankona: Fixed! Good eye.

ScottEilerman (approved):

PR #406, opened 2026-02-04
two bug fixes

smaticka:
Two small bug fixes are in this PR.

1) the `make` command when building NHMG in ROMS is fixed (previous error is below)
2) when cstar looks for partitioned files, it looked for the wrong number of digits in the files if the number of partitions equaled to 10^x (for x>=1). (previous error also below)

**Previous error for `make`:**
image

**Previous error for handling partitioned files:**
image

- [ ] Closes #xxxx
- [x] Tests passing

### DIFF
--- a/cstar/roms/external_codebase.py
+++ b/cstar/roms/external_codebase.py
@@ -43,7 +43,7 @@ class ROMSExternalCodeBase(ExternalCodeBase):
 
         # Compile NHMG library
         _run_cmd(
-            f"make nhmg COMPILER={cstar_sysmgr.environment.compiler}",
+            f"make all COMPILER={cstar_sysmgr.environment.compiler}",
             cwd=roms_root / "Work",
             msg_pre="Compiling NHMG library...",
             msg_err="Error when compiling ROMS' NHMG library.",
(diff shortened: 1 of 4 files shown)


cstar/roms/external_codebase.py
```diff
@@ -43,7 +43,7 @@ def _configure(self) -> None:
 
         # Compile NHMG library
         _run_cmd(
-            f"make nhmg COMPILER={cstar_sysmgr.environment.compiler}",
+            f"make all COMPILER={cstar_sysmgr.environment.compiler}",
```
ankona: This fix is the likely cause of the integration test failures because they reference an older version of roms. 

Try updating the  template blueprint loaded from `cstar/tests/integration_tests/blueprints/fixtures.py::modify_template_blueprint` and configured in `cstar/tests/integration_tests/config.py` or adjusting the blueprint instance in the failing tests.

You might want to confirm internally that the change to the ROMS make files is going to be merged upstream. If not, this will cause c-star to no longer work with "vanilla ucla-roms."

ankona (changes requested):

cstar/tests/unit_tests/roms/test_roms_external_codebase.py
smaticka: @ScottEilerman  is this the block of code you removed? 
The tests are still failing, but I think the py file you sent me did not have all of your changes.

ScottEilerman: I sent an updated one in Slack, see if that works!

ScottEilerman (approved):

PR #407, opened 2026-02-07
Consolidate dev dependencies in `pyproject.toml`

ankona:
# Consolidate dev dependencies in `pyproject.toml`

This PR moves additional dependencies used by `pre-commit` into the 'dev' optional dependency group in `pyproject.toml`, with the following goals:

- ensure that `pyproject.toml` is the single source of truth for project wide dependency management.
- ensure that developers do not need to manually install type stubs for full IDE support
- enables manual execution of `mypy` without invoking pre-commit

## Change list

1. Move additional dependencies from `.pre-commit-config.yaml` into `pyproject.toml`
2. Alphabetize depdendencies in `pyproject.toml`
3. Update github action to use local dependencies
4. Mitigate previously existing minor typing issues

## Check list:

- [x] Tests passing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ keywords = ["MCDR", "CDR", "ocean carbon", "climate"]
 [project.optional-dependencies]
 dev = [
     "coverage>=7.13",
-    "isort>=7.0.0",
     "mypy>=1.16.1",
     "pre-commit==3.8.0",
     "pytest>=7.0",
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -26,13 +26,13 @@ classifiers = [
 requires-python = ">=3.10"
 dynamic = ["version"]
 dependencies = [
+    "pooch>=1.8.1",
```
ScottEilerman: I think we can actually move pooch to dev dependencies. We used to rely on it for fetching netcdf data but switched to `requests` with Dafydd's IO refactor. Looks like we still use it in the integration tests though.

ScottEilerman (approved):
Approved -- see one optional comment.

pyproject.toml
ankona: great catch. moved.

PR #408, opened 2026-02-07
Reduce magic string usage

ankona:
# Reduce Magic String Usage

This PR reduces usage of magic strings by using or creating additional constants. I was inspired by the need to update the `configuration.rst` docs and am moving toward being able to generate this automatically.

## Change list

1. Create constants for feature flags
2. Use FF constants where hardcoded strings were previously used

## Review Checklist

- [X] Tests passing
- [X] Full type hint coverage


ScottEilerman (commented):
Approved in principle -- looks like some missing imports are causing test failures. If it's an easy fix feel free to merge after.

ScottEilerman (approved):

PR #409, opened 2026-02-07
Add env-show command to CLI

ankona:
This PR adds an experimental `cstar env show` command to the C-Star CLI. This command is useful for identifying the currently active set of output directories and enabled-status of feature flags.

## Change list

1. Add `env` command
2. Add `show` subcommand
3. Refactor ENV var declarations w/`typing.Annotated` for metadata
4. Use (3) to generate `configurtion.rst` content automatically from code

## Review checklist

- [X] Tests passing
- [X] Full type hint coverage
- [X] Docs updated


cstar/base/utils.py
```diff
@@ -1,89 +1,304 @@
...
 
-DEFAULT_OUTPUT_ROOT_NAME: t.Literal["output"] = "output"
-"""A fixed `output_root_name` to be used when generating outputs with ROMS."""
+    name: str = ""
+    """The standard environment variable name used for the setting."""
 
-DEFAULT_CACHE_HOME: t.Final[str] = "~/.cache"
-"""The default, XDG-compliant directory where c-star cache is written"""
+    @property
+    def value(self) -> str:
```
ScottEilerman, line 45: Is this property meant to get the actual value from the environment? If so, it doesn't. If it's just meant to get the default, let's rename it to avoid confusion (`default_value`?).

ScottEilerman: Based on trying `cstar env show` out locally, I think we do need to add:

```python
        if (value := os.getenv(self.name)) is not None:
            return value
```
at the top of this function.

ankona: Great catch. That definitely was a bug. I've fixed it and welcome your re-test.

PR #410, opened 2026-02-07
Implement XDG-compliant DirectoryManager

ankona:
This PR adds a `DirectoryManager` to manage the configuration used to specify directories where C-Star will write outputs. The manager implements the [XDG Base Directory Standard](https://specifications.freedesktop.org/basedir/latest/).

## Change List

1. Implement the `DirectoryManager`
2. Update constants as needed to establish the XDG defaults and associated env vars
3. Consolidates the use of `CSTAR_HOME` and `CSTAR_OUTDIR` into `CSTAR_STATE_HOME`

> [!NOTE]
> Documentation updates incoming...

- [X] Closes #CSD-497
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/execution/file_system.py
```diff
@@ -125,27 +249,27 @@ def _dir_set(self) -> set[Path]:
...
     def _codebases_dir(self) -> Path:
         """The directory for codebases."""
-        return self.root / self._CODEBASES_NAME
+        return self.input_dir / self._CODEBASES_NAME
 
     @property
     def joined_output_dir(self) -> Path:
         """The directory for de-partitioned outputs."""
-        return self.root / self._JOINED_OUTPUT_NAME
+        return self.output_dir / self._JOINED_OUTPUT_NAME
```
ScottEilerman: I think this one goes up a level, per a previous discussion. Still not totally set on it but would want user feedback to inform a change.

ScottEilerman (approved):
I think I reviewed pieces of this in other PRs, and from a look at this branch in an earlier state. Gave it a quick test run on Anvil and local and it seems to be behaving as expected. One small change I noted below, otherwise I think we can merge it!

cstar/execution/file_system.py
ankona: My mistake trying to copy/paste my way through the task. Updated to not be a subdirectory of `output_dir`... again.

PR #411, opened 2026-02-07
Use XDG base directories as write targets

ankona:
This PR integrates the `DirectoryManager` throughout the orchestration codebase.

## Change List

1. Use `DirectoryManager.xxx_home()` for root of write directories
2. Fix bug in environment variable discovery (variable name vs. environment variable key)
3. Relocate feature flag constants to features module
4. Fix bug when evaluating precedence order of indirect env var value
5. Rename FF_ON to FLAG_ON for clarity when used in non-FF context
6. Adjust `Step.working_dir` to return a relative path suitable for use with `xxx_home`

## Review Checklist

- [X] Closes #CSD-521
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage
- [X] New functions/methods are listed in `api.rst`


cstar/execution/file_system.py
```diff
@@ -242,27 +249,27 @@ def _dir_set(self) -> set[Path]:
...
     def _codebases_dir(self) -> Path:
         """The directory for codebases."""
-        return self.root / self._CODEBASES_NAME
+        return self.input_dir / self._CODEBASES_NAME
 
     @property
     def joined_output_dir(self) -> Path:
         """The directory for de-partitioned outputs."""
-        return self.root / self._JOINED_OUTPUT_NAME
+        return self.output_dir / self._JOINED_OUTPUT_NAME
```
ScottEilerman:
```suggestion
        return self.root / self._JOINED_OUTPUT_NAME
```
I think this one goes up a level, per a previous discussion. Still not totally set on it but would want user feedback to inform a change.

PR #412, opened 2026-02-09
Remove outdated examples

ScottEilerman:
Remove outdated example script and notebook. The notebook in particular was using very old code (`Case` etc.). These are effectively replaced by the RTD and the upcoming user guide.

- [x] Tests passing


ankona (approved):
LGTM!

PR #420, opened 2026-02-13
Logging and usability improvements

ScottEilerman:
This PR contains a number of unrelated (sorry) adjustments, mainly around logging, presentation, and usability.

- Implement CSTAR_LOG_LEVEL env-var
- Reimplement CSTAR_CLOBBER_WORKING_DIR for single-blueprint cases
- Implement CSTAR_LOG_LEVEL env-var
- Use remote mypy precommit hook instead of relying on local installation
- Move env-var handling methods to separate module to avoid circular imports
- Remove excess new-lines while logging the ROMS handler
- Remove __init__-based import shortcuts
- Adjust error handling and presentation in the worker/service layer

- 

- [x] Tests passing


cstar/entrypoint/service.py
```diff
@@ -423,16 +427,20 @@ async def execute(self) -> None:
...
                     self.log.exception(
                         "Terminating service due to failure in _on_delay."
                     )
+                    exc = e
 
             self._send_update_to_hc({self.CMD_PREFIX: "heartbeat"})
 
         self._shutdown()
 
+        if exc:
```
ankona, line 441: I'm not sure I understand why we take the time to collect and raise this exception. We're already exiting the application and the logging above should have captured the underlying issue.

cstar/execution/handler.py
```diff
@@ -189,7 +189,7 @@ async def updates(self, seconds: float = 10) -> None:
                     if self.status != ExecutionStatus.RUNNING:
                         return
                     elif line:
-                        self.log.info(line)
+                        self.log.info(line.rstrip())
```
ankona, line 192: good catch

cstar/base/env.py
```diff
@@ -0,0 +1,278 @@
...
+_GROUP_SIM: t.Final[str] = "Simulation Configuration"
+_GROUP_UNK: t.Final[str] = "Uncategorized Configuration"
+
+FLAG_ON: t.Final[str] = "1"
+"""Value indicating a feature flag is enabled."""
+
+FLAG_OFF: t.Final[str] = "0"
+"""Value indicating a feature flag is disabled."""
+
+DEFAULT_OUTPUT_ROOT_NAME: t.Literal["output"] = "output"
```
ankona: `DEFAULT_OUTPUT_ROOT_NAME` is related to the ROMS `.in` files. It probably shouldn't have moved to this new module.

cstar/base/env.py
```diff
@@ -0,0 +1,278 @@
...
+"""A fixed `output_root_name` to be used when generating outputs with ROMS."""
+
+SCRATCH_DIRS: t.Final[list[str]] = ["SCRATCH", "SCRATCH_DIR", "LOCAL_SCRATCH"]
```
ankona: We should be able to get rid of `SCRATCH_DIRS`, since we really only need the single value in the default below:

```
ENV_CSTAR_SCRATCH_DIRS: t.Annotated[
    t.Literal["CSTAR_SCRATCH_DIRS"],
    EnvVar(
        "A comma-separated list of environment variable names used to identify scratch paths on HPC systems, in search order.",
        _GROUP_FS,
        "SCRATCH,SCRATCH_DIR,LOCAL_SCRATCH",
    ),
] = "CSTAR_SCRATCH_DIRS"
```

cstar/roms/input_dataset.py
```diff
@@ -219,18 +219,37 @@ def partition_files(files: list[Path]) -> list[Path]:
...
-                            idfile, np_xi=np_xi, np_eta=np_eta,
-                            include_coarse_dims=False)
-                    except Exception:
+                            idfile,
+                            np_xi=np_xi,
+                            np_eta=np_eta,
+                            include_coarse_dims=include_coarse_dims,
+                        )
+                    except Exception as e:
+                        self.log.error(
```
ankona:
```suggestion
                        self.log.exception(
```
Using `logger.exception` will automatically include the exception information and you can get rid of the dynamic f-string

cstar/roms/simulation.py
```diff
@@ -622,6 +627,14 @@ def fs_manager(self) -> RomsFileSystemManager:
         """Return the file system manager for the simulation."""
         return cast(RomsFileSystemManager, self._fs_manager)
 
+    def _conditionally_clear_root(self):
+        env_item = get_env_item(ENV_CSTAR_CLOBBER_WORKING_DIR)
+        if env_item.value != FLAG_OFF and self.fs_manager.root.exists():
+            self.log.warning(
```
ankona: Consider moving the logging into `FileSystemManager.clear` in case you want the logs no matter where it's called from.

I also think it's a little awkward to include information about the environment variable in that logging message. By virtue of the logging occurring, we know the value was not `FLAG_OFF`.

If you're worried the flag has a "non-flaggy" value, we might implement some validation to the env-var discovery, instead of mixing w/random other logic.

cstar/roms/simulation.py
```diff
@@ -622,6 +627,14 @@ def fs_manager(self) -> RomsFileSystemManager:
         """Return the file system manager for the simulation."""
         return cast(RomsFileSystemManager, self._fs_manager)
 
+    def _conditionally_clear_root(self):
+        env_item = get_env_item(ENV_CSTAR_CLOBBER_WORKING_DIR)
+        if env_item.value != FLAG_OFF and self.fs_manager.root.exists():
```
ankona: Consider putting `if self.root.exists` into `FileSystemManager.clear` instead. It will clean up this calling code.

cstar/roms/simulation.py
```diff
@@ -1193,9 +1206,18 @@ def setup(
         runtime_code_dir = self.fs_manager.runtime_code_dir
         input_datasets_dir = self.fs_manager.input_datasets_dir
 
-        compile_time_code_dir.mkdir(parents=True, exist_ok=True)
-        runtime_code_dir.mkdir(parents=True, exist_ok=True)
-        input_datasets_dir.mkdir(parents=True, exist_ok=True)
+        self._conditionally_clear_root()
+
+        try:
+            compile_time_code_dir.mkdir(parents=True)
```
ankona, line 1210: good catch to remove the `exist_ok=True`'s

cstar/roms/simulation.py
```diff
@@ -1193,9 +1206,18 @@ def setup(
...
+            runtime_code_dir.mkdir(parents=True)
+            input_datasets_dir.mkdir(parents=True)
+        except OSError as e:
+            raise CstarExpectationFailed(
+                f"The inputs directory ({self.fs_manager.input_dir}) is not empty and has an unknown state. "
+                f"Set {ENV_CSTAR_CLOBBER_WORKING_DIR} and rerun if you want to delete this directory and start from "
+                f"scratch."
```
ankona: - unnecessary f-string on `f"scratch"`. 
- Consider editing this long message to avoid the third line....

```
msg = f"The input directory is not empty. Re-run with {ENV_CSTAR_CLOBBER_WORKING_DIR}={FLAG_ON} to delete the directory and continue."
raise CStarExpectationFailed(msg)
```

docs/releases/v0.2.x.rst
```diff
@@ -39,13 +40,20 @@ Improvements
...
+- Implement CSTAR_LOG_LEVEL env-var
+- Automatically retry partitioning without coarse dims if it fails when they are included
 
 Miscellaneous
 ~~~~~~~~~~~~~
 
 - Additional details added in docstrings
 - Silence OMP warning originating from numba
 - Modify a blueprint in the tests directory for a refactored layout in the remote data repository
+- Use remote mypy precommit hook instead of relying on local installation
```
ankona: The change to precommit from a couple commits back isn't in the notes. This doesn't add any useful information (I removed something that you didn't know was changed). Consider removing.

docs/releases/v0.2.x.rst
```diff
@@ -39,13 +40,20 @@ Improvements
...
+- Move env-var handling methods to separate module to avoid circular imports
+- Remove excess new-lines while logging the ROMS handler
+- Remove __init__-based import shortcuts
+- Adjust error handling and presentation in the worker/service layer
```
ankona, line 54: Couldn't "adjust error handling and presentation" encompass "remove excess newlines". I don't think the newlines item warrants it's own mention in release notes.

ankona (commented):
Looks pretty good. I'm happy to simplify those `__init__.py` files and avoid circular import headaches.

cstar/entrypoint/service.py
ScottEilerman: Yeah, I made these changes after we talked through it, so I should have given some more context. In general I wanted to hide some of the service/worker errors so the "real" error was more apparent to the user (this is seen in the code changes where I changed an error raise to a `logger.debug()`. I was also trying to clean up a few instances where the service/worker reported "unknown" status even when a clear failure occurred -- I don't remember if I actually achieved that. The particular change you commented on here was because even though I was hitting an error during setup (I was intentionally repeatedly hitting the "you already have files in your working dir error), I was still getting a `"Blueprint execution completed"` at the very end instead of a `"Blueprint execution failed"` (these get printed from the blueprint cli `run.py`). The `execute_runner` wrapper wasn't actually entering the `except` clauses that result in a return code of 1. I'm not sure if the route I chose to get there is a good one, but it seems to be working.

cstar/base/env.py
ScottEilerman: Good catch, I'll move it back!

cstar/base/env.py
ScottEilerman: Yep, IDE says it's unused now. Deleted.

cstar/roms/simulation.py
ScottEilerman: I think my IDE does that automatically. I'll clean it up.

cstar/entrypoint/service.py
ankona: aaaahhh. Good catch. nothing to see here!

ankona (approved):
LGTM!

PR #421, opened 2026-02-13
Pass `include_coarse_dims=False` if `roms_tools.partion_netcdf` fails

matt-long:
This puts the call to `roms_tools.partition_netcdf` in a try/except block:
1. try without using `include_coarse_dims`
2. if that fails, use `include_coarse_dims=False`
3. if that fails, abort


ScottEilerman (approved):
I've got pretty much the same thing that I tacked on to another branch waiting for review, but we can go ahead and merge this one, since I don't think mine will be reviewed before the weekend!

PR #422, opened 2026-02-14
Fix unintended checkout on cached remote repository

ankona:
## Summary

This PR fixes an error that occurs when using the `CachedRemoteRepositoryRetriever`.

There are two underlying issues:

1. A checkout was performed on the cache
2. Copied symlinks pointed to the cache directory

## Change List

1. Split clone && checkout behavior in remote repository retriever
2. Execute checkout on copied repository
3. Ensure cache content (not folder) is copied to target dir 

## Review Checklist

- [X] Closes #CSD-538
- [X] Tests passing
- [X] Full type hint coverage

PR #423, opened 2026-02-14
Update/edit Csd 538

ScottEilerman:
I took a look at #422 (even though it's in draft). Here are a few edits:
* also call checkout from RemoteRepositoryStager
* edit tests to pass

Some things to note with this fix (I don't think any of them are problematic, but good to make sure it's intended):
* Previously, the `Retriever` was responsible for getting _and checking out_ a repository. Now the checkout responsibility is in the `Stager`.
* Thus, adjusted the retriever test to not expect a checkout
* Adjusted the cached stager test to not expect the files in the cache are the same as the target (since only the target gets checked out)

### DIFF
--- a/cstar/tests/unit_tests/io/test_cached_stager.py
+++ b/cstar/tests/unit_tests/io/test_cached_stager.py
@@ -66,15 +66,17 @@ def test_cached_stager_reuse(tmp_path: Path) -> None:
     stage_path_1 = tmp_path / "my-roms-1"
     staged_data_1 = source_data_1.stage(stage_path_1)
 
-    source_data_2 = SourceData(repo_uri)
+    source_data_2 = SourceData(repo_uri, identifier="main")
     stage_path_2 = tmp_path / "my-roms-2"
 
     fake_pull = mock.MagicMock()
     fake_clone = mock.MagicMock()
+    fake_checkout = mock.MagicMock()
 
     with (
         mock.patch("cstar.io.retriever._pull", new=fake_pull) as mock_pull,
         mock.patch("cstar.io.retriever._clone", new=fake_clone) as mock_clone,
+        mock.patch("cstar.io.retriever._checkout", new=fake_checkout) as mock_checkout,
     ):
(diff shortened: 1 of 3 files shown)


cstar/tests/unit_tests/io/test_cached_stager.py
```diff
@@ -66,15 +66,17 @@ def test_cached_stager_reuse(tmp_path: Path) -> None:
     stage_path_1 = tmp_path / "my-roms-1"
     staged_data_1 = source_data_1.stage(stage_path_1)
 
-    source_data_2 = SourceData(repo_uri)
+    source_data_2 = SourceData(repo_uri, identifier="main")
```
ankona, line 69: I left identifier empty so it would pull the default for whatever that target repo is. For example, I think MARBL uses "stable". `main` isn't guaranteed.

cstar/tests/unit_tests/io/test_retriever.py
```diff
@@ -157,23 +157,18 @@ def test_read_raises(self, mocksourcedata_remote_repo):
...
 
         mock_clone.assert_called_once_with(
             source_repo=source.location, local_path=tmp_path
         )
-        mock_checkout.assert_called_once_with(
-            source_repo=source.location,
-            local_path=tmp_path,
-            checkout_target="test_target",
-        )
+
```
ankona, line 171: I was too tired to deal with the borken test on friday. TYVM!

cstar/tests/unit_tests/io/test_cached_stager.py
ScottEilerman: This is just in the test, to make sure it calls `_checkout` (unless I'm misunderstanding your comment).

ankona: I realized it was the test after I left my comment... I figured it'd be sneaky trying to remove it and pretend I didn't make that mistake. lol!

ScottEilerman: 🙈

PR #429, opened 2026-02-20
Fix environment.yml laptop

smaticka:
`environment.yml` was missing `jupyterlab` and the command in the tutorial won't work without it. 

Also a single line describing where the output of C-star will end up is included. Please confirm this is correct: `~/cstar/<NAME_OF_BLUEPRINT>`.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -3,14 +3,15 @@ channels:
   - conda-forge
 dependencies:
   - python>=3.10,<3.14
-  - numpy>2.0,<2.4  # currently numba doesn't support 2.4
-  - roms-tools
+  - numpy
+  - roms-tools>=3.5,<4
   - compilers
   - netcdf-fortran
   - mpich
   - xarray
   - netCDF4
   - pip
+  - jupyterlab
   - ipykernel
   - rsync
   # only needed for building docs
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -2,17 +2,17 @@ name: cstar_env
 channels:
   - conda-forge
 dependencies:
-  - python>=3.10,<3.14
-  - numpy>2.0,<2.4
-  - roms-tools>=3.5,<4
   - compilers
-  - netcdf-fortran
-  - mpich
-  - xarray
-  - netCDF4
-  - pip
-  - jupyterlab
   - ipykernel
+  - jupyterlab
+  - mpich
+  - netCDF4
+  - netcdf-fortran
+  - numpy>2.0,<2.4
+  - pip
+  - python>=3.10,<3.14
+  - roms-tools>=3.5,<4
   - rsync
+  - xarray
   # only needed for building docs
   - pandoc
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python>=3.10,<3.14
-  - numpy
+  - numpy>2.0,<2.4
   - roms-tools>=3.5,<4
   - compilers
   - netcdf-fortran
--- a/docs/tutorials/tutorial_bp.ipynb
+++ b/docs/tutorials/tutorial_bp.ipynb
@@ -315,7 +315,8 @@
    "id": "7fcfe37f48a14f7",
    "metadata": {},
    "source": [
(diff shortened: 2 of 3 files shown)


ankona:
> `environment.yml` was missing `jupyterlab` and the command in the tutorial won't work without it.
> 
> Also a single line describing where the output of C-star will end up is included. Please confirm this is correct: `~/cstar/<NAME_OF_BLUEPRINT>`.

We're using the XDG-base directory specification, as of last week. That change makes `~/cstar` incorrect. Right now, everything is written to the data home directory.

- highest precedence is a path in the env var `CSTAR_DATA_HOME`
- it falls back to `Path(os.environ["XDG_DATA_HOME"]) / "cstar"` if no `CSTAR_` env var is present
- if both `CSTAR_DATA_HOME` and `XDG_DATA_HOME` env vars are not found, it writes to the default XDG data dir with a cstar subdirectory appended: `Path("~/.local/share") / "cstar"`

ci/environment.yml
```diff
@@ -3,14 +3,15 @@ channels:
   - conda-forge
 dependencies:
   - python>=3.10,<3.14
-  - numpy>2.0,<2.4  # currently numba doesn't support 2.4
-  - roms-tools
+  - numpy
```
ankona: Are you sure you want to remove the pin for numpy? This looks like it could potentially try to use 2.4. Is the `numba` issue from the comment resolved? Maybe pinning `roms-tools` already handles it?

ankona: After reading the offline discussion, I know this is good to go.

docs/tutorials/tutorial_bp.ipynb
```diff
@@ -315,7 +315,8 @@
    "id": "7fcfe37f48a14f7",
    "metadata": {},
    "source": [
-    "If the output directory remains from a previous run, it may need to be remove before running the simulation."
+    "If the output directory remains from a previous run, it may need to be remove before running the simulation.\\\n",
+    "Note that all `C-Star` runs of a roms-marbl blueprint will output to the root directory `~/cstar/<NAME_OF_BLUEPRINT>`."
```
ScottEilerman:
```suggestion
    "Note that for the blueprints in this tutorial, `C-Star` will output results to the directory `~/cstar/<NAME_OF_BLUEPRINT>`."
```

ScottEilerman:
> > `environment.yml` was missing `jupyterlab` and the command in the tutorial won't work without it.
> > Also a single line describing where the output of C-star will end up is included. Please confirm this is correct: `~/cstar/<NAME_OF_BLUEPRINT>`.
> 
> We're using the XDG-base directory specification, as of last week. That change makes `~/cstar` incorrect. Right now, everything is written to the data home directory.
> 
>     * highest precedence is a path in the env var `CSTAR_DATA_HOME`
> 
>     * it falls back to `Path(os.environ["XDG_DATA_HOME"]) / "cstar"` if no `CSTAR_` env var is present
> 
>     * if both `CSTAR_DATA_HOME` and `XDG_DATA_HOME` env vars are not found, it writes to the default XDG data dir with a cstar subdirectory appended: `Path("~/.local/share") / "cstar"`

I ended up setting the two blueprints we use within the tutorial to output with that convention (`~/cstar/blueprint_name`) as an improvement from dumping everything in a subdirectory of the current directory (in this case, within the repo itself). I suggested a modification to Sam's text just now to limit the scope of that statement to "blueprints in this tutorial." I guess if we haven't yet, we should make clear in the docs somewhere that `cstar blueprint` cli respects the blueprint output dir, and `cstar workplan` will place things in the relevant XDG locations.

ci/environment.yml
ScottEilerman:
```suggestion
  - numpy>2.0,<2.4
```
I think my slack message came across unclear. I think we _should_ pin it here for safety for now, because I don't trust our dependency resolution through multiple rounds of conda/pip installs. I think if it's unpinned, it's possible/probable to get incompatible numpy and numba versions.

ScottEilerman: see https://github.com/CWorthy-ocean/C-Star/pull/429/changes#r2842704220

ScottEilerman (approved):

PR #460, opened 2026-03-13
Add trace log level

ankona:
# Summary
This change adds a trace logging level and adjusts the logging calls from an overly verbose method.

## New Features

- Add trace log level to reduce log frequency during `Workplan` execution

## Code Review Checklist

- [X] Tests passing
- [X] Full type hint coverage

### DIFF
--- a/cstar/base/log.py
+++ b/cstar/base/log.py
@@ -14,9 +14,14 @@ TRACE_LOG_NAME: t.Final[str] = "TRACE"
 
 
 class TraceLogger(logging.Logger):
+    """A customized logger type that offers a `trace` method used to emit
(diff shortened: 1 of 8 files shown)


cstar/base/log.py
ScottEilerman, line 180: Did you verify that `addLevelName` up at the top also gets it added to this lookup? Or do we need to add additional handling here?

ScottEilerman (approved):
One question but if that's handled it looks good to me!

cstar/base/log.py
ankona: Yep, `addLevelName` is the mechanism for making this seamless!

PR #465, opened 2026-03-23
Implement automatic symlinking for CDR and extract_data files

ScottEilerman:
# Summary
The primary addition in this PR is to add/improve functionality with ROMS inputs that have hard-coded paths buried in .opt files. This applies to the existing `cdr_forcing` file, and a new `nesting_info` file added in the PR. A corresponding PR in cson-forge will create blueprints that point to a predetermined filename in the run directory for each type of file. C-Star will create a symlink from those file locations that are expected by ROMS to the real locations of the files as specified in the blueprint.

As part of this PR, I have changed the `roms` executable is now executed in the `work` subdirectory instead of `output`. I did this as part of this PR so that the new symlinks don't need to be in the `output` directory, which could be confusing. It seems like the right place for us to be running anyway. I think we only ever ran in the `output` directory so that the outputs went to the right place by default.

Finally, just because I was in the neighborhood, I slipped in an unrelated change to default to the correct ROMS Makefile if one isn't provided.

## Breaking Changes

- The `roms` executable is now executed in the `work` subdirectory instead of `output`
- The output location in the runtime settings (`roms.in`) file is now _always_ modified to place outputs in the output directory (since we no longer run ROMS in the output directory)

## New Features

- Add `nesting_info` block to RomsMarblBlueprint
- Create work-dir symlinks to predetermined paths for CDR and nesting info files
- If no `Makefile` is provided in the `compile_time` additional code block, ROMS will copy and use the appropriate `Makefile` from the ROMS repo

## Improvements

- The original runtime settings file and the one that is written and used by C-Star are now appended with `_ORIGINAL` and `_PATCHED` to reduce confusion.

## Code Review Checklist

- [ ] Closes #xxxx
- [x] Tests added
- [x] Tests passing
- [x] Full type hint coverage


cstar/orchestration/adapter.py
```diff
@@ -270,6 +271,22 @@ def adapt(self) -> ROMSCdrForcing | None:
...
 
 
+class NestingInfoAdapter(ModelAdapter[models.RomsMarblBlueprint, ROMSNestingInfo]):
+    @t.override
+    def adapt(self) -> ROMSNestingInfo | None:
+        if self.model.nesting_info is None:
+            return None
+
+        return ROMSNestingInfo(
+            location=str(self.model.nesting_info.data[0].location),
```
ankona, line 297: I think a unit test here might discover that this can blow up if `self.model.nesting_info.data == []`?

cstar/roms/input_dataset.py
```diff
@@ -430,7 +430,29 @@ def validate(self):
             )
 
 
-class ROMSCdrForcing(ROMSInputDataset):
+class ROMSInputDatasetWithOptSymlink(ROMSInputDataset):
+    """Input dataset that requires a compile-time path redirect"""
+
+    required_opt_file: str
```
ankona: Don't forget your docstrings on these attributes (they will populate the generated documentation)

cstar/roms/input_dataset.py
```diff
@@ -430,7 +430,29 @@ def validate(self):
...
+
+    symlink_name: str
+
+
+class ROMSCdrForcing(ROMSInputDatasetWithOptSymlink):
     """An implementation of the ROMSInputDataset class for CDR forcing files."""
 
-    pass
+    required_opt_file: str = "cdr_frc.opt"
```
ankona: If this attribute is fixed/predetermined, we may want to  avoid mutable attributes. Consider declaring these as `Final[str]` instead of `str`

cstar/roms/input_dataset.py
```diff
@@ -430,7 +430,29 @@ def validate(self):
...
+
+
+class ROMSNestingInfo(ROMSInputDatasetWithOptSymlink):
+    """An implementation of the ROMSInputDataset class for nesting info files."""
+
+    required_opt_file: str = "extract_data.opt"
+    """Compile-time opt file that must be present and reference ``symlink_name``."""
+
+    symlink_name: str = "nesting.nc"
+    """Filename ROMS expects in the run directory for this dataset."""
```
ankona: I would expect something to happen in this class that makes use of the symlink attributes. Some sort of hook-type behavior, like:

```
class RomsInputDataset:
  def _hook(self):
    ..

  def behavior(self):
    self._hook()

class SpecialDataset
  @t.override
  def _hook():
    # create symlink here
```

without this, we force downstream code to implement the symlinking behavior and they appear tightly coupled

cstar/roms/simulation.py
```diff
@@ -814,36 +822,63 @@ def roms_runtime_settings(self) -> ROMSRuntimeSettings:
...
+        but not partitioned or included in ROMS runtime settings.
 
         Returns
         -------
         list of ROMSInputDataset
-            A list containing all input datasets used in the simulation.
+            A list containing input datasets that need partitioning.
         """
-        input_datasets: list[ROMSInputDataset] = []
+        datasets: list[ROMSInputDataset] = []
```
ankona:
```suggestion
        return [
            x
            for x in [
                self.model_grid,
                self.initial_conditions,
                self.tidal_forcing,
                self.river_forcing,
                *self.boundary_forcing,
                *self.surface_forcing,
                *self.forcing_corrections,
            ]
            if x is not None
        ]
```
Consider a comprehension to make it clear that you're not really doing anything unique to each item

cstar/roms/simulation.py
```diff
@@ -1256,6 +1299,87 @@ def setup(
...
+            [f.basename for f in self.compile_time_code.source]
+            if self.compile_time_code is not None
+            else []
+        )
+
+        work_dir = self.fs_manager.work_dir
+        work_dir.mkdir(parents=True, exist_ok=True)
+
+        for dataset in active:
+            opt_filename = dataset.required_opt_file
```
ankona: Consider extracting content of this `for` loop into it's own little method to improve readability of this method.

It should also get rid of the need to cast something in the middle of the loop.

```
for x in [x for x in self.mydatasets if isinstance(x, ROMSInputDatasetWithOptSymlink)]:
  validate_supplementary_dataset(x)
```

cstar/roms/simulation.py
```diff
@@ -1256,6 +1299,87 @@ def setup(
...
+            link_name = dataset.symlink_name
+
+            if self.compile_time_code is None:
```
ankona: This will be true or false for all items in the loop, and should be executed prior to the loop to keep the repetitive behavior clear

cstar/roms/simulation.py
```diff
@@ -1256,6 +1299,87 @@ def setup(
...
+            working_copy = cast("StagedFile", dataset.working_copy)
+
+            if not working_copy:
+                raise CstarExpectationFailed(
+                    f"No local copy of {dataset.__class__.__name__} found after "
+                    "fetching input datasets."
+                )
+
+            symlink_path = work_dir / link_name
+            if symlink_path.is_symlink() or symlink_path.exists():
```
ankona: consider 1-liner `symlink_path.unlink(missing_ok=True)`

cstar/roms/simulation.py
```diff
@@ -1256,6 +1299,87 @@ def setup(
...
+                )
+
+            if opt_filename not in compile_time_filenames:
+                raise CstarExpectationFailed(
+                    f"{opt_filename!r} must be listed in compile_time_code when "
+                    f"{dataset.__class__.__name__} is specified. "
+                    f"Found: {compile_time_filenames}"
+                )
+
+            opt_content = (compile_time_code_dir / opt_filename).read_text()
```
ankona: Consider a context manager to ease/speed removal of the entire file content from memory
```suggestion
            with (compile_time_code_dir / opt_filename).open() as opt_content:
```

cstar/roms/simulation.py
```diff
@@ -1371,7 +1495,16 @@ def build(self, rebuild: bool = False) -> None:
                 + "\nROMSSimulation.compile_time_code is not staged locally."
                 + "\n Call ROMSSimulation.compile_time_code.get() and try again"
             )
+
         build_dir = self.compile_time_code.working_copy.common_parent
+
+        # in most cases, the Makefile in the ROMS repo is the right one to use.
```
ankona: consider extracting this block into it's own little method. The docstring can explain what you have here as a comment more cleanly

```python
def _ensure_makefile_target(self) -> Path:
  """Use a user-specified Makefile when one is supplied and fallback
  to copying one from the source repo when it is not.

  Returns
  -------
  Path
  """
  makefile_target = build_dir / "Makefile"
  if makefile_target.exists():
    return makefile_target

  makefile_repo = self.codebase.working_copy.path / "Work" / "Makefile"
  shutil.copyfile(makefile_repo, makefile_target)
  return makefile_target
```

cstar/roms/simulation.py
```diff
@@ -1371,7 +1495,16 @@ def build(self, rebuild: bool = False) -> None:
...
+        # if they don't provide their own, copy the one from the repo's working copy to the build dir
+        makefile_target = build_dir / "Makefile"
+        if not makefile_target.exists():
+            makefile_repo = self.codebase.working_copy.path / "Work" / "Makefile"  # type: ignore[union-attr]
```
ankona: Consider avoiding hiding this and raise in case the unexpected occurs:

```
if self.codebase.working_copy is None:
    raise ValueError("Unable to copy makefile with null working copy")
```

cstar/roms/simulation.py
```diff
@@ -1542,26 +1678,35 @@ def run(
...
+        old_runtime_settings_path = Path(self.runtime_code.working_copy[0].path)
         final_runtime_settings_file = (
-            Path(self.runtime_code.working_copy[0].path).parent / f"{self.name}.in"
+            old_runtime_settings_path.parent / f"PATCHED_{self.name}.in"
         )
         self.roms_runtime_settings.to_file(final_runtime_settings_file)
 
+        old_runtime_settings_path.rename(
+            old_runtime_settings_path.parent
+            / f"ORIGINAL_{old_runtime_settings_path.name}"
```
ankona: consider a suffix instead of a prefix so original and derived files appear next to each other in an alphabetically sorted list of paths.

It may also help ensure that a generic glob search like `*.in` doesn't erroneously find the original file

cstar/roms/simulation.py
ankona: update: we already have `_validate_input` that is run on every input dataset. you can probably piggyback off that and avoid adding this method performing looping completely (and maybe then also avoid having extra properties to isolate symlink items)

cstar/roms/simulation.py
```diff
@@ -1440,11 +1573,14 @@ def pre_run(self, overwrite_existing_files=False) -> None:
         run : Executes the compiled ROMS model.
         post_run : Performs post-processing steps after execution.
         """
-        # Partition input datasets and add their paths to namelist
-        if self.input_datasets is not None and all(
-            [isinstance(a, ROMSInputDataset) for a in self.input_datasets]
+        # Partition input datasets that require partitioning
+        # (excludes cdr_forcing, nesting_info, and similar datasets)
```
ankona: - this "excludes x, y, and z" comment should be moved into the docstring of `partitionable_datasets` so its purpose is clear without coming to this random spot. Consider a more general purpose description of the exceptions when documenting in the docstring.
- "partition things that should be partitioned" is probably not necessary to document.
- The property returns a non-nullable type (`list[ROMSInputDataset']`) so I'm not sure why we need to do a null check?
- The property return a `list[ROMSInputDataset]` so I'm not sure why we need this isinstance check?

cstar/roms/simulation.py
```diff
@@ -1256,6 +1299,87 @@ def setup(
...
+            Directory where compile-time code files have been staged by ``setup()``.
+
+        Raises
+        ------
+        CstarExpectationFailed
+            If ``compile_time_code`` is None when a supplementary dataset is defined,
+            if the required opt file is not listed in ``compile_time_code``, or if the
+            opt file does not contain the expected filename string.
+        """
+        supplementary: list[ROMSInputDatasetWithOptSymlink | None] = [
```
ankona: consider property to avoid changing multiple spots:

```python
@property
def supplemental_datasets(self):
  return filter((self.cdr_forcing, self.nesting_info), lambda x: x is not None)
```

which makes us avoid repeating the definition of the non-partitionables in:

```python
def input_datasets():
  return [*self.partitionable_datasets, *self.supplemental_datasets]
```
and in this method...

it will also avoid having the null filter (`active = [... if d is not None]`) only found here in the validation method so it must also be repeated...

ankona (changes requested):
I'm not a huge fan of the subclasses that don't really do anything except set values for what should be constants. It looks like something that could be done with composition instead of inheritance. 

Here's a non-working alternative idea:

```python
@dataclass
class DatasetLinkage:
  """Identify files that will be symlinked from an external source."""

  name: t.Final[str]
  """The desired name for the symlinked file."""
  source: t.Final[str]
  """The name of the file that will be symlinked to the desired name."""

  def create_linkage(self, target_dir: Path) -> Path:
      target_path = target_dir / self.name
      target_path.unlink(missing_ok=True)
      target_path.symlink_to(self.source)

class ROMSInputDataset:
    linked_files: list[DatasetLinkage]

class ROMSSimulation:
    def with_linked_cdr_forcing(self, ds: ROMSInputDataset) -> ROMSInputDataset:
        CDR_OPT = "cdr_frc.opt"
        CDR_NC = "cdr.nc"

        ds.linked_files.append(
            DatasetLinkage(CDR_OPT, CDR_NC),
        )
        return ds

    def with_linked_nesting(self, ds: ROMSInputDataset) -> ROMSInputDataset:
        CDR_OPT = "cdr_frc.opt"
        CDR_NC = "cdr.nc"

        ds.linked_files.append(
            DatasetLinkage(CDR_OPT, CDR_NC),
        )
        return ds
```

I think you have a few spots with easy tweaks that should occur (clean up inline comments that should be docstrings) so I'll mark as "request changes". If you see a need for the subclasses that I'm not seeing, feel free to ignore that part.

cstar/roms/input_dataset.py
ScottEilerman: docstrings added for the refactored version of these params

cstar/roms/input_dataset.py
ScottEilerman: These get defined differently in the refactor and are no longer hard-coded in the class definition (but are effectively hardcoded in the adapted calls to DatasetLinker). I unfortunately couldn't make DatasetLinker frozen because I also used the LoggingMixin with it, and it seems we can't do that (it tried to assign _log but can't because it's frozen).

cstar/roms/input_dataset.py
ScottEilerman: Refactored per discussion!

cstar/roms/simulation.py
ScottEilerman: done, much cleaner.

cstar/roms/simulation.py
ScottEilerman: This code is gone in the refactor.

cstar/roms/simulation.py
ScottEilerman: This whole function is gone in the refactor, each dataset now handles its own linking via DatasetLinker calls in .get()

cstar/roms/simulation.py
ScottEilerman: validation has bee cleaned up in refactored version

cstar/roms/simulation.py
ScottEilerman: this remains in the refactored version but I'll leave it for now. these opt files are known to be small.

cstar/roms/simulation.py
ScottEilerman: I swear I tried this but apparently not. Done. Longer-term, I'd love to understand if there are use cases for this type ambiguity or if we can refactor to guarantee it won't ever be None.

cstar/roms/simulation.py
ScottEilerman: bad AI code is bad. cleaned up comment, and the logic is improved in the refactor.

cstar/roms/simulation.py
ScottEilerman: good suggestion! done.

cstar/orchestration/adapter.py
ScottEilerman: This pattern is used throughout the adapters, and I guess in the current internal usage, these are always coming from yaml and I'm not sure if you can even define an empty list. Anyway, I added a `min_length=1` annotation of the relevant pydantic field, so I _think_ that should protect against this.

ScottEilerman (commented):
I'm going to submit the current round of comment responses and push the refactored code. I still need to fix some failing unit tests, and do a self-review for missing docstrings.

ankona (approved):
Per offline convo, LGTM!

PR #470, opened 2026-03-27
Enable specifying user-defined variables when launching a workplan

ankona:
# Summary

This change adds the ability to specify user-defined variables when executing a workplan with the `cstar` CLI. 

A user may now pass variables one-by-one with:

- `cstar workplan run --run-id ... --var xxx=XXX --var yyy=YYY`

And, they may pass a file containing all variables with:

- `cstar workplan run --run-id ... --varfile </path/to/varfile>`

The user-defined variables will be validated against the variables defined in the target workplan. Appropriate error messaging for missing and malformed variables has been added.

## Breaking Changes

- N/A

## New Features

- Enable user supplied variables for templating from `cstar workplan run ...`

## Bug Fixes

- Fix workplan template with inconsistent blueprint path

## Improvements

- Refactor `cli.workplan.run` to make use of `typer` callbacks for input validation
- Add `--dry-run` option to test workplan generation in `cstar workplan run`

## Miscellaneous

- N/A

## Security Fixes

- N/A

## Code Review Checklist

- [X] Closes #CSD-567
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/additional_files/templates/wp/workplan.yaml
```diff
@@ -21,7 +21,7 @@ steps:
     - name: Ensemble X
       application: sleep
       depends_on: ["Prepare"]
-      blueprint: ./blueprint.yaml
+      blueprint: ~/code/cstar/cstar/additional_files/templates/blueprint.yaml
```
ScottEilerman, line 24: Is this change intended?

cstar/cli/workplan/run.py
```diff
@@ -17,63 +21,262 @@
 log = get_logger(__name__)
 
 
+def preprocess_vars(
+    ctx: typer.Context,
+    var: list[str],
```
ScottEilerman: nit: `var` -> `vars` in this function since it's a list?

cstar/cli/workplan/run.py
```diff
@@ -17,63 +21,262 @@
...
+    run_id : str
+        The run-id to reload
+    """
+    repo = TrackingRepository()
+    wp_run = asyncio.run(repo.get_workplan_run(run_id))
+    if wp_run is None:
+        msg = f"No runs with the id `{run_id}` could be found."
+        raise typer.BadParameter(msg)
+
+    # TODO @ankona: identify appropriate reloading behavior for user vars
```
ScottEilerman: flagging this -- not quite sure what the context is. follow up task?

cstar/cli/workplan/run.py
```diff
@@ -17,63 +21,262 @@
...
+    # when re-using a run-id.
+
+    # ensure the environment matches the prior run
+    os.environ.update(wp_run.environment)
+
+    path = wp_run.workplan_path
+    msg = f"Re-starting run-id `{run_id}` with workplan originating in `{path}`"
+    log.info(msg)
+
+    return wp_run.trx_workplan_path.as_posix()
```
ScottEilerman, line 186: I'm guessing this is what caused the change in the template file I flagged earlier.

cstar/cli/workplan/run.py
ScottEilerman: oh right, `vars` is a built-in. could do `var_list` but `var` is also fine to leave.

cstar/cli/workplan/shared.py
```diff
@@ -80,3 +83,75 @@ def display_summary(
...
+
+    if not v:
+        msg = f"Found variable with empty value for key: {entry}"
+        raise typer.BadParameter(msg)
+
+    return k, v
+
+
+def check_and_capture_kvps(var: list[str]) -> t.Mapping[str, str] | None:
+    """Capture all unique keyj-value pairs from user-supplied configuration
```
ScottEilerman, line 135:
```suggestion
    """Capture all unique key-value pairs from user-supplied configuration
```

ScottEilerman (approved):
A few nits/typos but no blockers. I didn't look at the tests but can if there are any concerns!

cstar/cli/workplan/run.py
ankona: Since typer defaults to taking the name directly from the variable and it's provided individually, it worked out nicely, e.g. `--var xxx=XxXx --var yyy=YyYy`

I've renamed the variable to be more clear but override the default so `--var` and `-v` are the exposed options.

cstar/additional_files/templates/wp/workplan.yaml
ankona: i use the _default blueprint path_ in tests as a key to replace on (in lieu of _real templating_). 

all the other blueprint paths in that file were the default and when I executed a new test, it ended up vomiting when trying to validate existence of a file at `./blueprint.yaml`

definitely intended

cstar/cli/workplan/run.py
ankona: see prior comment for the root cause. 

i think you're probably referring to `as_posix` since it was a recent issue... at the point where a run is re-loaded, the path is always a local path so this is not a problem.

cstar/cli/workplan/shared.py
ankona: w00t. fixed.

cstar/cli/workplan/run.py
ankona: Re-running is kind of annoying in that i need to restore the prior environment state to ensure the run is the same. For example, if you originally ran with `CSTAR_XXX=1 cstar workplan run --run-id foo file.yaml` then re-run with `cstar workplan run --run-id foo`, it would not have `CSTAR_XXX` set. That is solved by capturing all `CSTAR_XXX` variables when a run is performed and reloading them.

In this case, you might do `cstar workplan run --var foo=F --run-id abc ./wp.yaml`

- what happens when this is re-run?

After I made the initial _TODO_, I solved by persisting the `varfile` in the _run folder_ so it can be reused. The old todo is removed (no longer needed)

cstar/additional_files/templates/wp/workplan.yaml
ScottEilerman: Oh I see. Makes sense. I didn't realized we had a default or used it that way, I assumed this was your local repository path or something. Carry on!

PR #473, opened 2026-04-02
Add helpful error for misread HTML yamls

ScottEilerman:
# Summary
I was getting a really obtuse error when trying to run a remote blueprint from github. It ended up being because the content it was getting was HTML content, not raw YAML content. This seems like a potentially common problem, if you just copy a github link without going to the "raw" version. I added some additional error text, wrapping the original exception. We could do additional validation of the content or try to modify to the right link, but I'm not sure it's worth all that effort right this moment.

## Improvements

- Improve error message when failing to load a remote YAML file.

### DIFF
--- a/cstar/orchestration/serialization.py
+++ b/cstar/orchestration/serialization.py
@@ -5,6 +5,7 @@ from pathlib import Path, PosixPath
 
 import yaml
 from yaml import safe_load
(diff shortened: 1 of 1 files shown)


ankona (approved):
Looks perfect to me.

PR #474, opened 2026-04-02
loosen nested dict typealias

ScottEilerman:
# Summary
This solves the warning spam that Sam was getting. I'm not 100% sure if we want to loosen this much, but I also don't want to try to guess every possible combination of nested override types for a wide expanse of possible blueprints in the future.

## Improvements

- Loosen Pydantic type validation for override dictionaries to mitigate pydantic warnings


ankona (approved):
This thing was awful. Either way works as a janky solution for now, and one without warnings is better!

PR #475, opened 2026-04-02
pin versions of example files

ScottEilerman:
# Summary
I update the example file repo to _not_ have a Makefile and to include a CDR file. This immediately broke the tutorial examples (which I serendipitously was using to test something else, so I noticed). Those blueprints now pin the commit of the test data repo and can be updated more intentionally when desired.

## Bug Fixes

- Pin version of tutorial data repo


ankona (commented):
LGTM

PR #476, opened 2026-04-02
bump typer version

ScottEilerman:
# Summary
The version of typer I had installed (0.17.something) was causing an error with the some of the new preprocess hooks. I think they fixed the bug in 0.19.something, but I went ahead and bumped our req to the latest.

## Bug Fixes

- Bump typer version to avoid list-default bug


ankona (approved):
LGTM!

PR #477, opened 2026-04-02
avoid mutable defaults

ScottEilerman:
# Summary
My IDE highlighted these while I was looking at other stuff. They weren't the cause of the problem I was looking at, but I figure we should fix it anyway. Though... mypy/ruff insist that you should add `| None` in these cases and I really disagree. None is not an ok value to pass and is immediately overridden. Not sure if there's a better way to handle this other than the `ignore`?


ankona (approved):
I _thought_ I'd gotten these a couple weeks back and still missed them. Good eye!

PR #478, opened 2026-04-02
Fix local status mapping

ScottEilerman:
# Summary
While testing something else, I was running a workplan locally, and my first blueprint failed, but the next one (that depends on the first) ran anyway. After some tinkering, I think both of these changes are needed to properly register a failure in the worker.

## Bug Fixes

- Fix incorrect status mapping of local runs, which caused dependencies to not be cancelled if upstream tasks failed.


ankona (approved):
Great catch.

ankona (approved):
LGTM!

PR #491, opened 2026-04-10
Bugfix ROMS runtime handling

ScottEilerman:
# Summary

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- Fix possibly incorrectly renaming a marbl input file to the `_ORIGINAL` roms.in file
- Mitigate possible ROMS output-name-length errors by using a relative output path in the `roms.in`

## Improvements

- N/A

## Miscellaneous

- N/A

## Security Fixes

- N/A

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [x] Tests passing
- [x] Full type hint coverage

### DIFF
--- a/cstar/roms/simulation.py
+++ b/cstar/roms/simulation.py
@@ -823,7 +823,7 @@ class ROMSSimulation(Simulation):
         # explicitly set output dir and basename
         simulation_runtime_settings.output_root_name = (
             cstar.roms.runtime_settings.OutputRootName(
-                output_root_name=str(self.fs_manager.output_dir) + "/output"
+                output_root_name="../output/output"
             )
         )
 
--- a/cstar/roms/simulation.py
+++ b/cstar/roms/simulation.py
@@ -1596,7 +1596,7 @@ class ROMSSimulation(Simulation):
         run_path = self.fs_manager.work_dir
 
         # save modified roms.in and rename original for clarity
-        old_runtime_settings_path = Path(self.runtime_code.working_copy[0].path)
+        old_runtime_settings_path = self.fs_manager.runtime_code_dir / self._in_file
         final_runtime_settings_file = (
             old_runtime_settings_path.parent / f"{self.name}_PATCHED.in"
         )
--- a/cstar/tests/unit_tests/roms/test_roms_simulation.py
+++ b/cstar/tests/unit_tests/roms/test_roms_simulation.py
@@ -1577,10 +1577,10 @@ class TestProcessingAndExecution:
             ),
             mock.patch("pathlib.Path.rename", new_callable=mock.Mock),
         ):
-            sim.exe_path = sim.directory / "ROMS/compile_time_code/roms"
+            sim.exe_path = sim.fs_manager.compile_time_code_dir / "roms"
             mock_process_instance = mock.MagicMock()
             mock_local_process.return_value = mock_process_instance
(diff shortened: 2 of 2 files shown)


cstar/tests/unit_tests/roms/test_roms_simulation.py
```diff
@@ -1580,7 +1580,7 @@ def test_run_local_execution(
             sim.exe_path = sim.directory / "ROMS/compile_time_code/roms"
             mock_process_instance = mock.MagicMock()
             mock_local_process.return_value = mock_process_instance
-            runtime_code_dir = sim.directory / "ROMS/runtime_code"
+            runtime_code_dir = sim.directory / "input/runtime_code"
```
ankona: You can create a `RomsFileSystemManager` here if you want to avoid hardcoding roms-specific subdirectories (includes: compiletime code, runtime code, datasets, codebases, joined outputs):

```python
roms_fs = RomsFileSystemManager(sim.directory)

# now you can use `roms_fs.runtime_code_dir` directly
sim.exe_path = roms_fs.compile_time_code_dir / "roms"

```

cstar/roms/simulation.py
```diff
@@ -823,7 +823,7 @@ def roms_runtime_settings(self) -> ROMSRuntimeSettings:
         # explicitly set output dir and basename
         simulation_runtime_settings.output_root_name = (
             cstar.roms.runtime_settings.OutputRootName(
-                output_root_name=str(self.fs_manager.output_dir) + "/output"
+                output_root_name="../output/output"
```
ankona, line 826: i think you can use `str(self.fs_manager.output_dir / "output")` to avoid having this relative path (if you're just looking to avoid string concatenation)

What was happening here?

ScottEilerman: The resulting absolute path can be very long, and ROMS has a limit on filepath lengths because Fortran, so the relative path will be shorter in most cases.

cstar/tests/unit_tests/roms/test_roms_simulation.py
ScottEilerman: I was lazy because it was in a test and I wasn't sure what I had available. Fixing...

ankona: actually, i recall that there is already a ROMS fs manager on a ROMS sim. you can use `sim.fs_manager.runtime_code_dir` directly since it is already that type

smaticka:
when I run with this as my output directory in my `roms.in` file, I get this resulting location of the `output` directory. If that's the desired architecture, it looks good to me. I can try properly testing it with your code change though, instead of my manual change. 

```
codebases
input_datasets
compile_time_code
runtime_code
output
```

ScottEilerman:
> when I run with this as my output directory in my `roms.in` file, I get this resulting location of the `output` directory. If that's the desired architecture, it looks good to me. I can try properly testing it with your code change though, instead of my manual change.
> 
> ```
> codebases
> input_datasets
> compile_time_code
> runtime_code
> output
> ```

It will be relative to the run directory, so that structure would make sense if you manually run roms from, say, `compile_time_code`. When C-Star is running it, we runs ROMS from the work directory, so I get a structure like this:

Screenshot 2026-04-10 at 1 16 44 PM

PR #495, opened 2026-04-16
Add ability to join .ext files during post_run

ScottEilerman:
# Summary

Adds ability to join .ext files using the correct tool, instead of skipping them. 

Note: these functions could be abstracted better, but this gets the job done for now, and these will all go away in a few months once the PIO project is completed and partitioning/joining is no longer needed.

## New Features

- .ext files are now passed through extract_data_join at the same time as other files are passed through ncjoin

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/cstar/roms/simulation.py
+++ b/cstar/roms/simulation.py
@@ -25,7 +25,7 @@ from cstar.base.utils import (
     _run_cmd,
     slugify,
 )
-from cstar.execution.file_system import RomsFileSystemManager
+from cstar.execution.file_system import RomsFileSystemManager, remove_files
 from cstar.execution.handler import ExecutionStatus
 from cstar.execution.local_process import LocalProcess
 from cstar.execution.scheduler_job import create_scheduler_job
@@ -72,30 +72,11 @@ if TYPE_CHECKING:
     from cstar.execution.handler import ExecutionHandler
 
 
-def _remove_wildcard_pattern(
-    input_dir: Path, wildcard_pattern: str, logger: logging.Logger
-) -> None:
-    """Removes files in the input dir matching the wildcard pattern.
-
-    Parameters
-    ----------
-    input_dir: directory in which to search for and remove matching files
-    wildcard_pattern: the glob pattern to match files against
-    logger: logger object to post log messages to
-
-    Returns
-    -------
-    None
-    """
-    logger.debug(f"Removing files matching wildcard pattern {wildcard_pattern}...")
-    for p in input_dir.glob(wildcard_pattern):
-        p.unlink()
-
-
 def _ncjoin_wildcard(
     wildcard_pattern: str, input_dir: Path, output_dir: Path, logger: logging.Logger
 ) -> None:
-    """Spatially join netcdfs matching the wildcard pattern using ncjoin, and move the joined output to output_dir.
+    """Spatially join netcdfs matching the wildcard pattern using ncjoin, move the joined output to output_dir, and
+    remove the unjoined outputs (except for rst files, which would need to be repartitioned later).
 
     Parameters
     ----------
@@ -117,17 +98,17 @@ def _ncjoin_wildcard(
     out_file = input_dir / wildcard_pattern.replace("*.", "")
     out_file_name = out_file.name
     out_file.rename(output_dir / out_file_name)
-    logger.info(f"done spatially joining {out_file}")
+    logger.info(f"Done spatially joining {out_file}")
 
     if "rst" not in wildcard_pattern:
-        _remove_wildcard_pattern(input_dir, wildcard_pattern, logger)
+        remove_files(input_dir, wildcard_pattern)
 
 
 def _extract_data_join_wildcard(
     wildcard_pattern: str, input_dir: Path, output_dir: Path, logger: logging.Logger
 ) -> None:
     """Spatially join netcdfs matching the wildcard pattern using extract_data_join,
-    and move the joined output to output_dir.
+     move the joined output to output_dir, and remove the unjoined outputs.
 
     Parameters
     ----------
@@ -158,9 +139,9 @@ def _extract_data_join_wildcard(
     ocean_file = input_dir / ocean_file_name
     ocean_file.unlink(missing_ok=True)
 
-    _remove_wildcard_pattern(input_dir, wildcard_pattern, logger)
+    remove_files(input_dir, wildcard_pattern)
 
-    logger.info(f"done spatially joining {out_file}")
+    logger.info(f"Done spatially joining {out_file}")
 
 
 class ROMSSimulation(Simulation):
@@ -1790,11 +1771,14 @@ class ROMSSimulation(Simulation):
             )
 
         nprocs = int(get_env_item(ENV_CSTAR_NPROCS_POST).value)
-        unique_wildcards = {Path(fname.stem).stem + ".*.nc" for fname in files}
+        unique_wildcards = {
+            str(Path(fname.stem).with_suffix(".*.nc")) for fname in files
+        }
 
         with ThreadPoolExecutor(max_workers=nprocs) as executor:
-            results = executor.map(_spatial_join, unique_wildcards)
-            _ = [r for r in results]  # exhaust iterator
+            # list() exhausts the returned iterator, which is needed to surface any errors
+            # that were raised in the threaded join operations
+            list(executor.map(_spatial_join, unique_wildcards))
 
         self.persist()
 
--- a/cstar/roms/simulation.py
+++ b/cstar/roms/simulation.py
@@ -91,7 +91,8 @@ def _ncjoin_wildcard(
     """
     logger.info(f"Joining netCDF files {wildcard_pattern}...")
 
-    command = "extract_data_join" if "ext" in wildcard_pattern else "ncjoin"
+    extract_data = "ext" in wildcard_pattern
+    command = "extract_data_join" if extract_data else "ncjoin"
 
     _run_cmd(
         f"{command} {wildcard_pattern}",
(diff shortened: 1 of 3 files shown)


cstar/roms/simulation.py
```diff
@@ -73,6 +72,26 @@
     from cstar.execution.handler import ExecutionHandler
 
 
+def _remove_wildcard_pattern(
```
ankona: - consider renaming to clarify, e.g. `remove_files` tells us what this does more clearly than `remove_wildcard_pattern` 
- consider moving to `cstar.base.utils` or `cstar.execution.file_system` where it can be reused (and `file_system` is good context for where you'd expect to find it)

cstar/roms/simulation.py
```diff
@@ -73,6 +72,26 @@
...
+    input_dir: directory in which to search for and remove matching files
+    wildcard_pattern: the glob pattern to match files against
+    logger: logger object to post log messages to
+
+    Returns
+    -------
+    None
+    """
+    logger.debug(f"Removing files matching wildcard pattern {wildcard_pattern}...")
+    for p in input_dir.glob(wildcard_pattern):
```
ankona: consider validation of the wildcard.

```python
if "*" not in wildcard_pattern:
  raise ValueError("invalid wildcard")
```

also...
consider verifying that some expected changes occurred:

```python
removed = False
for p in input_dir.glob(wildcard_pattern):
  p.unlink()
  removed = True

if not removed:
  logger.warning(f"Wildcard {wildcard_pattern!r} did not result in any modifications")
```

cstar/roms/simulation.py
```diff
@@ -90,15 +109,56 @@ def _ncjoin_wildcard(
...
         f"ncjoin {wildcard_pattern}",
         cwd=input_dir,
         raise_on_error=True,
     )
-
     out_file = input_dir / wildcard_pattern.replace("*.", "")
-    out_file.rename(output_dir / out_file.name)
+    out_file_name = out_file.name
+    out_file.rename(output_dir / out_file_name)
+    logger.info(f"done spatially joining {out_file}")
```
ankona: nit: try to be consistent on your log message capitalization

cstar/roms/simulation.py
```diff
@@ -90,15 +109,56 @@ def _ncjoin_wildcard(
...
+
+    if "rst" not in wildcard_pattern:
```
ankona, line 103: I think this says "delete the files after joining them unless they're reset files"...

It's ok, but having this hidden behavior that isn't documented (or parameterized) might be problematic/confusing, later.

If we added to the signature, we would be able to read the docstring and know it happens via the Parameter information.

```python
def _ncjoin_wildcard(..., exclusions=".rst"]):

  Parameters
  ----------
  ...
  exclusions: str
    comma-separated set of wildcards specifying files that should not be cleaned up after the join occurs. By default, reset files are not removed.

  ...
  all_files = (... existing code)
  all_files = {all_files}.difference(itertools.chain.from_iterable(work_dir.glob(exclusion) for exclusion in exclusions.split(",")))
```

cstar/roms/simulation.py
```diff
@@ -90,15 +109,56 @@ def _ncjoin_wildcard(
...
+
+    out_file_name = f"child_bry.{out_file_timestamp_dot_nc}"
+    out_file = input_dir / out_file_name
+    out_file.rename(output_dir / out_file_name)
+
+    ocean_file_name = f"ocean.{out_file_timestamp_dot_nc}"
+    ocean_file = input_dir / ocean_file_name
+    ocean_file.unlink(missing_ok=True)
+
+    _remove_wildcard_pattern(input_dir, wildcard_pattern, logger)
```
ankona: the hidden side effects of the method are inconsistent with `_ncjoin_wildcard` and not documented.

- Consider updating docstring to say "Move and rename the joined output..."
- Consider allowing the join to only do the join and let the caller worry about renaming.
- Consider passing the cleanup wildcards per note above

you could have the caller do `_extract_data_join_wildcard(join_pattern, join_dir, output_dir, exclusions=[wildcard_pattern.replace("*.","").split(".")[0]]` - now this method doesn't need to know the format of the exclusion at all and the caller is in control.

... i think in the end, that would basically leave you with exactly the same method except the target command... which could also be parameterized into:


```python
def _join_netcdfs(
    wildcard_pattern: str, input_dir: Path, output_dir: Path, exclusions: str, tool: str = "ncjoin",
) -> None:
    _run_cmd(
        f"{tool} {wildcard_pattern}",
        cwd=input_dir,
        raise_on_error=True,
    )
    out_file = input_dir / wildcard_pattern.replace("*.", "")
    out_file_name = out_file.name
    out_file.rename(output_dir / out_file_name)
    logger.info(f"done spatially joining {out_file}")

    _cleanup_files(wildcard_pattern, exclusions)
```

so, your clients could do

`_join_netcdfs_(..., tool="ncjoin")` or `_join_netcdfs_(..., tool="extract_data_join")`


and finally, it looks like we could parameterize the renaming of the file stem:

```python
def _join_netcdfs(
    wildcard_pattern: str, input_dir: Path, output_dir: Path, exclusions: str, stem: str = "", tool: str = "ncjoin",
) -> None:
    _run_cmd(
        f"{tool} {wildcard_pattern}",
        cwd=input_dir,
        raise_on_error=True,
    )
    out_file = input_dir / wildcard_pattern.replace("*.", "")
    out_file_name = out_file.name
    if stem:
        out_file_name = out_file.with_stem(stem).name
    out_file.rename(output_dir / out_file_name)
    logger.info(f"done spatially joining {out_file}")

    _cleanup_files(wildcard_pattern, exclusions)
```

cstar/roms/simulation.py
```diff
@@ -1708,27 +1768,33 @@ def post_run(self) -> None:
...
-            nprocs = int(get_env_item(ENV_CSTAR_NPROCS_POST).value)
-            with ThreadPoolExecutor(max_workers=nprocs) as executor:
-                results = executor.map(spatial_joiner, unique_wildcards)
-                _ = [r for r in results]  # exhaust iterator
+        nprocs = int(get_env_item(ENV_CSTAR_NPROCS_POST).value)
+        unique_wildcards = {Path(fname.stem).stem + ".*.nc" for fname in files}
+
+        with ThreadPoolExecutor(max_workers=nprocs) as executor:
+            results = executor.map(_spatial_join, unique_wildcards)
+            _ = [r for r in results]  # exhaust iterator
```
ankona: - unnecessary list comp. consider `list(results)`

cstar/roms/simulation.py
```diff
@@ -90,15 +109,56 @@ def _ncjoin_wildcard(
...
+    out_file.rename(output_dir / out_file_name)
+    logger.info(f"done spatially joining {out_file}")
+
+    if "rst" not in wildcard_pattern:
+        _remove_wildcard_pattern(input_dir, wildcard_pattern, logger)
+
+
+def _extract_data_join_wildcard(
+    wildcard_pattern: str, input_dir: Path, output_dir: Path, logger: logging.Logger
+) -> None:
```
ankona, line 109: Consider updating the return value of this to allow callers to have more certainty (are they guessing where the outputs are at or forced to search with a wildcard, later?). 

You could :

- return the path `out_file`
- return more info `{"out_file": xxx, "removed": list(yyy)}`

cstar/roms/simulation.py
ankona: I had to double-check, but i don't think we need to manually exhaust the list with the context manager. I believe it waits for all computations to be completed before it exits the `with`, unless there is a timeout.

this should be equivalent:

```python
with ThreadPoolExecutor(max_workers=nprocs) as executor:
  executor.map(_spatial_join, unique_wildcards)
```

cstar/roms/simulation.py
```diff
@@ -1708,27 +1768,33 @@ def post_run(self) -> None:
...
                 output_dir=self.fs_manager.joined_output_dir,
+                logger=self.log,
             )
 
-            nprocs = int(get_env_item(ENV_CSTAR_NPROCS_POST).value)
-            with ThreadPoolExecutor(max_workers=nprocs) as executor:
-                results = executor.map(spatial_joiner, unique_wildcards)
-                _ = [r for r in results]  # exhaust iterator
+        nprocs = int(get_env_item(ENV_CSTAR_NPROCS_POST).value)
+        unique_wildcards = {Path(fname.stem).stem + ".*.nc" for fname in files}
```
ankona: nit: i think `Path(fname.stem).with_suffix(".*.nc")` has the same result and doesn't require manual concatenation

ankona (changes requested):
Double check that you really need a new method for `_extract_data_join_wildcard` or if you can parameterize the existing method.

cstar/roms/simulation.py
ScottEilerman: I have a vivid memory of needing to do this because it didn't work without it, and yet, now it works without it. Must have been some other subtle bug that was fixed along the way. Removed now.

cstar/roms/simulation.py
ScottEilerman: I had initially gone this route but you end up with so many `if` toggles that it seemed cleaner to split it into two distinct functions. You'd need to drill down the `tool` parameterization into `_cleanup_files` too and then you've got `ifs` in there. I'm going to leave as-is for now; as mentioned, this code just needs to get us through the current push and will go away once the in-development ROMS changes go through.

cstar/roms/simulation.py
ScottEilerman: I added a note in the docstring about the removal behavior, as it should indeed be documented. I don't expect this function to see further use outside of the current use-case.

cstar/roms/simulation.py
ScottEilerman: Added a return of the `removed` flag, user can check it if they want. In theory you could pass a single file name (non-wildcard) to this to remove it... not very efficient but 🤷 .

cstar/roms/simulation.py
ScottEilerman: AHA! I'm not crazy. A unit test fails if you don't exhaust the iterator -- the reason being that if you don't evaluate the result, it won't raise any errors that occurred in the process. So in a test where we mock a failure in ncjoin, it never surfaces.

Added it back in, but did simplify to `list()` instead of a comprehension.

ScottEilerman:
I made revisions for most of the comments, though I am going to refrain from more substantial refactoring or parameterization at the moment. I don't expect most of this functionality to ever be used outside the current context, and even that context is expected to go away in the near future, so this code will too. If we do end up needing it long-term or in other places, we can revisit a better design then (e.g. we've talked about making this a separate application and/or a side-car process that runs alongside the simulation, rather than waiting till the very end to join everything).

cstar/roms/simulation.py
ScottEilerman: I'm going to pass on this one for now. The one and only caller for this doesn't need to do anything more with the outputs, and I'd rather wait and see what a potential other use case actually needed before adding more complexity that needs testing.

ankona (approved):
LGTM!

PR #498, opened 2026-04-16
Add initial SimulationRunner preprocessing directive

ankona:
# Summary

This change adds the ability to specify a preprocessing directive in a workplan that is JIT executed by the worker.

The proof-of-concept directive, "continue-from", searches for reset files in a directory that is empty at the time the task is scheduled.

## Breaking Changes

- `deep_merge` is a pure method and no longer changes the inputs

## New Features

- Add `Directives` to enable just-in-time blueprint manipulation by the runner.

## Bug Fixes

- N/A

## Improvements

- N/A

## Miscellaneous

- N/A

## Security Fixes

- N/A

## Code Review Checklist

- [X] Closes #CSD-566
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage

### DIFF
--- a/cstar/orchestration/transforms.py
+++ b/cstar/orchestration/transforms.py
@@ -1027,24 +1027,27 @@ class DirectiveConfig(BaseModel):
         -------
         str
         """
-        with local_copy(directive_uri) as local_path:
+        with (
+            local_copy(directive_uri) as local_path,
+            local_copy(blueprint_uri) as local_bp,
+        ):
             model = deserialize(local_path, DirectiveConfig)
 
-        directives = model.directives
-        if not directives:
-            return blueprint_uri
+            directives = model.directives
+            if not directives:
+                return blueprint_uri
 
-        step = LiveStep(
-            name="directive-step",
-            application=Application.SLEEP,
-            blueprint=blueprint_uri,
-        )
-        directive_map = DirectiveConfig.directive_map
-        transforms = {
-            directive_map[key](config=t.cast("dict[str, dict[str, t.Any]]", config))
-            for key, config in directives.items()
-        }
-        for transform in transforms:
-            step = transform(step)[0]
+            step = LiveStep(
+                name="directive-step",
+                application=Application.ROMS_MARBL,
+                blueprint=local_bp,
+            )
+            directive_map = DirectiveConfig.directive_map
+            transforms = {
+                directive_map[key](config=t.cast("dict[str, dict[str, t.Any]]", config))
+                for key, config in directives.items()
+            }
+            for transform in transforms:
+                step = transform(step)[0]
 
         return str(step.blueprint_path)
--- a/cstar/orchestration/transforms.py
+++ b/cstar/orchestration/transforms.py
@@ -677,12 +677,7 @@ class OverrideTransform(Transform):
         blueprint: Blueprint = deserialize(bp_path, bp_type)
 
         updated_bp = self.apply(blueprint, step.blueprint_overrides)
-
         update: dict[str, t.Any] = {"blueprint_overrides": {}}
-        if bp_type == RomsMarblBlueprint:
-            update["work_dir"] = t.cast(
-                "RomsMarblBlueprint", updated_bp
-            ).runtime_params.output_dir
 
         live_step = LiveStep.from_step(step, update=update)
 
--- a/cstar/orchestration/transforms.py
+++ b/cstar/orchestration/transforms.py
@@ -2,10 +2,12 @@ import os
 import re
 import typing as t
 from collections import defaultdict
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable, Mapping
 from datetime import datetime, timedelta
 from enum import StrEnum
 from pathlib import Path
+
+from pydantic import BaseModel
 
 from cstar.base.exceptions import CstarExpectationFailed
 from cstar.base.feature import (
@@ -18,10 +20,12 @@ from cstar.base.utils import (
     deep_merge,
     slugify,
 )
+from cstar.execution.file_system import local_copy
 from cstar.orchestration.application import APP_CAT_BLUEPRINTS
 from cstar.orchestration.models import (
     Application,
     Blueprint,
+    KeyValueStore,
     RomsMarblBlueprint,
     Workplan,
 )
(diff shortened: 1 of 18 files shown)


cstar/orchestration/transforms.py
```diff
@@ -545,11 +551,121 @@ def override_output_directory(step: LiveStep) -> LiveStep:
...
+            application="directive",
+            blueprint=blueprint_uri,
+        )
+        directive_map = DirectiveConfig.directive_map
+        transforms = {
+            directive_map[key](config=t.cast("dict[str, dict[str, t.Any]]", config))
+            for key, config in directives.items()
+        }
+        for transform in transforms:
+            step = next(iter(transform(step)))
```
ScottEilerman: It took me a little while to figure out all the nested structures here, and that we're effectively just pulling off the one and only step, since transforms return a list of steps in the general sense, but a single step in this case. Any reason to not just do `transform(step)[0]`? Maybe in a future update we'll end up tweaking the return types based on our discussion yesterday ("directives are 1->1 while transforms can be 1->N"), and this will be moot.

cstar/orchestration/transforms.py
```diff
@@ -545,11 +551,121 @@ def override_output_directory(step: LiveStep) -> LiveStep:
...
+    ) -> dict[str, t.Any]:
+        source = Path(self._config["path"])
+        matches = sorted(source.rglob("*_rst*.nc"), reverse=True)
+
+        if not matches:
+            msg = f"No reset files located. Unable to continue from {source!r}"
+            raise CstarExpectationFailed(msg)
+
+        match = matches.pop(0).as_posix()
+        return {"initial_conditions": {"data": [{"location": match}]}}
```
ScottEilerman: We will also need to adjust the `start_time` to match the IC file. I need to ask someone more knowledgeable if it can be determined from the file's timestamp alone, or if we need to actually open the netcdf and examine the time dimension. If it's just the timestamp, let's include it in this PR. If we need to read the file, we can make it a follow-up ticket.

ScottEilerman: The answer to this ended up being the easy one, so let's go ahead and include it in this PR so the feature works out of the box.

The output file names should look like `output_rst.20000103000000.000.nc` if they're pre-partitioned, or `output_rst.20000103000000.nc` if they've been joined.

We want to:
* parse the timestamp portion and set `runtime_params["start_time"]` accordingly
* optionally: if we search the `output` instead of `joined_output` and point the IC to the `.000.nc` file, we can also set `partitioned: true` and skip re-partitioning

cstar/orchestration/transforms.py
ankona: It's a [ruff linting rule](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) to `prefer next over single element slice`. 

I've tweaked the return type to `Sequence` to continue to signal immutability (and used `tuple(xxx)` to enforce it) so we can use indices more readily.

cstar/orchestration/transforms.py
ankona: I've refactored restart file handling so the code is shared between the `ContinuanceTransform` and `RomsMarblTimeSplitter`. Additionally, I enabled passing the joined path or the output path by searching with a regex that accepts either. The new `ResetFile` entity ensures partitioned is correctly set in both cases.

cstar/orchestration/transforms.py
```diff
@@ -704,11 +714,340 @@ def override_output_directory(step: LiveStep) -> LiveStep:
...
+    """The segment identifier of the file."""
+    _ts: datetime = PrivateAttr()
+    """The timestamp parsed from the file name."""
+
+    EXT: t.ClassVar[t.Literal["nc"]] = "nc"
+    """The expected file extension for a restart file."""
+    FMT_TS: t.ClassVar[t.Literal["%Y%m%d%H%M%S"]] = "%Y%m%d%H%M%S"
+    """The expected timestamp format in the restart file name"""
+    PATTERN_RST: t.ClassVar[t.Literal[r"^(.*?)_rst\.(\d{14})(?:\.(\d{3}))?\.nc$"]] = (
+        r"^(.*?)_rst\.(\d{14})(?:\.(\d{3}))?\.nc$"
```
ScottEilerman: re `(\d{3}))`, unfortunately these are not uniformly zero-padded to three digits, it depends on how many overall partitions you have. If you have twelve, it'll always zero-pad to two digits, if you have 3462, it'll zero-pad to four digits. Easy enough to tweak the regex to parse, though it might present a wrinkle if we are trying to "guess" the file name of the zeroth partition. We'll either need to scan the directory and see how they're formatted, or we can calculate it from the blueprint (nx * ny).

cstar/orchestration/transforms.py
```diff
@@ -704,11 +714,340 @@ def override_output_directory(step: LiveStep) -> LiveStep:
...
+        ValueError
+            If the search path does not exist or contains no recognizable restart files.
+        """
+        search_path = search_path.expanduser().resolve()
+
+        if not search_path.exists():
+            msg = f"No directory found at path: {search_path!r}"
+            raise ValueError(msg)
+
+        if matches := sorted(search_path.rglob(f"*{cls.SUFFIX}.*.000.{cls.EXT}")):
```
ScottEilerman: noting hardcoded `.000.` here, see previous comment. my b for not surfacing this earlier.

cstar/orchestration/transforms.py
```diff
@@ -704,11 +714,340 @@ def override_output_directory(step: LiveStep) -> LiveStep:
...
+        -------
+        ResetFile
+
+        Raises
+        ------
+        ValueError
+            If the search path does not exist or contains no recognizable restart files.
+        """
+        ts = timestamp.strftime(cls.FMT_TS)
+        parted_clause = f".{segment:03d}" if segment is not None else ""
```
ScottEilerman: noting here as well

ScottEilerman (changes requested):
One correction needed because I didn't give you enough information to do it right the first time 😬 . Happy to approve after that!

cstar/orchestration/transforms.py
ankona: tyvm. fixed and tests added for varying pad size.

cstar/orchestration/transforms.py
ankona: tyvm. fixed and tests added for varying pad size

cstar/orchestration/transforms.py
ankona: tyvm. fixed

cstar/orchestration/transforms.py
ankona: Fixed!

ScottEilerman (approved):
LGTM -- going to approve and merge so I can start building on it!

PR #500, opened 2026-04-17
Implement template-fill transforms

ScottEilerman:
# Summary

Implements functionality for filling placeholder templates for both user-variables (from the command line) or for paths to the working directory of another step in the workplan.

## New Features

- Add template-filling transform for user variables and paths

## Bug Fixes

- Generalize OverrideTransform for non-ROMS application types

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage


cstar/entrypoint/worker/worker.py
```diff
@@ -385,7 +385,7 @@ async def execute_runner(
 
     try:
         configure_environment(log)
-
+        log.info(f"Beginning simulation run for {request.blueprint_uri}")
```
ankona: Consider moving this up to the "header block" 3 lines up.

Alternatively, I might have been too aggressive in my log level changes. Consider setting the level to `INFO` in `Service._on_start` and tweaking `Worker._on_start` to call `super()._on_start()` first... 

As-is, this is sort of misleading (it's saying it's starting the simulation before the runner is even created).

cstar/orchestration/models.py
```diff
@@ -219,6 +219,8 @@ class Application(StrEnum):
     """A call to the hostname executable to simplify testing."""
     HELLO_WORLD = "hello_world"
     """Sample custom application."""
+    PLOTTER = auto()
```
ankona: I swapped to using literals here... I got an urge to spell it out, since it's _user-facing_

you may want to be consistent

cstar/orchestration/orchestration.py
```diff
@@ -208,9 +208,9 @@ def is_in_progress(cls, status: "Status") -> bool:
 class LiveStep(Step):
     """A Step enriched with runtime metadata."""
 
-    parent: t.Self | None = None
+    parent: t.Self | None = Field(None, exclude=True)
     """The step for which this step is a sub-task."""
-    work_dir: Path | None = None
+    work_dir: Path | None = Field(None, exclude=True)
```
ankona, line 213: i'd be really surprised if this doesn't break the output in some way when we try to restart a workplan

cstar/orchestration/transforms.py
```diff
@@ -87,6 +91,145 @@ def get_transforms(application: str) -> list[Transform]:
...
+            return str(self._path_resolver(step_name))
+
+        if self._variable_resolver is None:
+            raise ValueError(
+                f"No variable resolver provided for placeholder '{{{{{content}}}}}'"
+            )
+        return self._variable_resolver(content)
+
+    def _fill(self, value: t.Any) -> t.Any:
+        """Recursively replace ``{{placeholder}}`` tokens in *value*.
```
ankona: I'm not sure "recursion" applies here and if it does, that it's desirable.

Recursion would be something like `foo: {{ bar-{{baz}} }}` and expecting bar-baz to be replaced after baz.

Are you saying recursive instead of "traverse" by mistake?

cstar/orchestration/transforms.py
```diff
@@ -87,6 +91,145 @@ def get_transforms(application: str) -> list[Transform]:
...
+        step : LiveStep
+            The step whose blueprint_overrides will be traversed.
+
+        Returns
+        -------
+        Iterable[LiveStep]
+            A single-element iterable containing the updated step.
+        """
+        filled_overrides = self._fill(dict(step.blueprint_overrides))
+        yield LiveStep.from_step(step, update={"blueprint_overrides": filled_overrides})
```
ankona: Consider treating as a document to avoid traversing the object hierarchy:

```python
def apply_resolver(x: str):
  if x.startswith("path:"):
    return PathResolver(x[5,])
  return user_variables[x]

def _fill(self, step: Step) -> Step:
  content = step.model_dump_json()
  matches = ... # regex magic, all matches
  if not matches:
    return step

  for match in set(matches):
    # don't repeat replacements 
    content = content.replace(match[1], apply_resolver(match[1]))

  if "{{" in content or "}}" in content: 
    raise CstarExpectationError("Some templated values were not filled or placeholders were malformed.")
  
  return step.model_validate_json(content)
```
   
I understand that as-is, the result is a set of overrides, so I could be persuaded that it's worth it.

cstar/orchestration/dag_runner.py
```diff
@@ -208,11 +209,29 @@ async def prepare_workplan(
...
+    file_io_extras: list = []
 
-    if transformer.is_modified:
-        log.info("A time-split workplan will be executed.")
+    if user_variables is not None:
+        named_config = UserDefinedVariables(
+            keys=set(wp_orig.runtime_vars), mapping=user_variables
+        )
+        if named_config.error:
+            raise ValueError(named_config.error)
```
ankona, line 225: looks like an omission by prior bad actor, but can you add the `Raises` section in the docstring with this `ValueError` listed

cstar/orchestration/transforms.py
ScottEilerman: "You" implies I wrote this docstring 😅 .

I'd guess that the _entity_ that wrote it means it's called on list/dicts within the structure. I'll update it.

cstar/orchestration/transforms.py
ScottEilerman: Done. I felt a little uneasy about parsing as a document from the point of view of potentially replacing things you really shouldn't. But, I like the ability to use placeholders more broadly... maybe you want your step names to be based on an user variable, or once we support compute_overrides, you might want those to be swappable. IDK. Maybe we need to lock it back down later, but it works for now.

cstar/orchestration/orchestration.py
ScottEilerman: Since this still seems to be working fine in the unit tests and manual testing, I guess we can go ahead with it and revert/fix later if we find it broke something?

cstar/orchestration/dag_runner.py
```diff
@@ -204,15 +205,38 @@ async def prepare_workplan(
...
+    ValueError
+        If the expected and provided user variables are not in agreement.
     """
     wp_orig = await asyncio.to_thread(deserialize, wp_path, Workplan)
     run_root_dir = output_dir / run_id
 
-    transformer = WorkplanTransformer(wp_orig, RomsMarblTimeSplitter())
-    wp = transformer.apply()
+    fill_transform: TemplateFillTransform | None = None
+    file_io_extras: list = []
```
ankona, line 218:
```suggestion
    file_io_extras: list[Awaitable[int]] = []
```

cstar/orchestration/dag_runner.py
```diff
@@ -223,20 +247,9 @@ async def prepare_workplan(
     file_io = [
```
ankona, line 247:
```suggestion
    file_io: list[Awaitable[int]] = [
```

cstar/orchestration/orchestration.py
ankona: ok, sounds good!

PR #511, opened 2026-04-23
Fix automatic output directory overrides

ankona:
# Summary

This change fixes a regression when applying automatic output directory overrides to a step. Previously, the transformation applied the transform but failed to update the workplan with the transformed step.

The result of the bug is that the override would not be applied until a new override transform was applied, which can occurred after time splitting. This gave the erroneous sense that there was a dependency on time splitting.

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- Fix bug where automatic output directory overrides appeared dependent on time-splitting

## Improvements

- N/A

## Miscellaneous

- N/A

## Security Fixes

- N/A

## Code Review Checklist

- [X] Closes #CSD-829
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


ankona:
recreating from appropriately named feature branch to trigger jira integration

PR #513, opened 2026-04-28
Make work dir earlier

ScottEilerman:
# Summary

## Bug Fixes

- Create work-dir earlier to avoid failure when making sym-links for CDR/nesting files.

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [x] Tests passing
- [ ] Full type hint coverage


ankona (approved):
LGTM!  we should investigate (later) if we want to change the way the `prepare` method works so nothing outside the `fs_manager` has to do all these mkdir's "manually"

PR #514, opened 2026-04-30
Move roms.in, symlink roms exe in work dir

ScottEilerman:
# Summary
This PR consolidates artifacts needed at roms runtime in the `work` directory. Instead of saving the modified roms.in back to `input/runtime_code` with a modified filename, we now save it with a consistent `cstar_generated_roms.in` in the work directory. We also symlink the built roms executable into the work directory, so a user who is debugging can rerun `mpirun ./roms cstar_generated_roms.in` from the `work` dir and it will have everything it needs (including symlinks to cdr.nc or nesting.nc for certain input files).

We could consider moving the roms executable and other compiled artifacts to the `work` dir, but we are rapidly moving toward a future where we may not need to recompile very often and could simply be referencing these executables that live pre-compiled in some other location.

## Improvements

- Save the generated roms.in to the work directory, where roms is run.
- Symlink the roms executable to the work directory, where roms is run.

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [x] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/cstar/execution/file_system.py
+++ b/cstar/execution/file_system.py
@@ -356,7 +356,6 @@ class RomsFileSystemManager(JobFileSystemManager):
             self.input_datasets_dir,
             self._codebases_dir,
             self.joined_output_dir,
-            self.work_dir,
         ]:
             if directory.exists():
                 shutil.rmtree(directory)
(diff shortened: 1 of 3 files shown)


ankona (approved):
Double-check if you can get rid of hardcoding the "roms" executable name using: `self.exe_path.with_segments(self.fs_manager.work_dir)`, but otherwise G2G!

PR #515, opened 2026-04-30
Add blueprint schema migration

ankona:
# Summary

This pull request adds the ability to migrate existing blueprint files to newer schema versions. The `BlueprintMigration` class is used to plan and execute a migration across multiple versions. It makes use of a sequence of `SchemaAdapter` subclasses implemented to handle the conversion of a blueprint from a specific source schema to a specific target schema.

The feature is currently disabled by default. Enable the feature flag using `CSTAR_FF_CLI_BP_MIGRATE=1 cstar blueprint migrate ./blueprint.yaml`

## Breaking Changes

- The `-d` short `dry-run` argument of `cstar workplan run` is removed to avoid a conflict.
- The `output-dir` argument of `cstar workplan run` now uses `--o` and `--output` for consistency (previously `-d` and `--output-dir`).

## New Features

- Enable upgrading blueprint files using the `cstar blueprint migrate` command

## Improvements

- Verbosity and dry-run CLI argument callbacks now ensure the appropriate environment variables are set. 

## Code Review Checklist

- [X] Closes #CSD-853
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/additional_files/templates/bp/roms_marbl/blueprint.1.0.0.yaml
```diff
@@ -0,0 +1,96 @@
+# yaml-language-server: $schema=/private/var/folders/pj/bk3042bs0tldhwq8htgfnm640000gn/T/pytest-of-chris/pytest-9/test_serial_planner0/schema.json
```
ScottEilerman, line 1: is there a way to generalize this? can we host the schema json in the repo and point to it remotely (or locally with relative paths)?

cstar/system/migration.py
```diff
@@ -0,0 +1,258 @@
...
+
+from cstar.base.adapter import ModelAdapter
+
+APP_ROMS_MARBL: t.Literal["roms_marbl"] = "roms_marbl"
+APP_ROMS_MARBL_SCHEMA_2025_1: t.Literal["2025.1"] = "2025.1"
+APP_ROMS_MARBL_SCHEMA_2026_1: t.Literal["2026.1"] = "2026.1"
+
+
+APP_HW: t.Literal["hello_world"] = "hello_world"
+APP_HW_SCHEMA_2026_1: t.Literal["2026.1"] = "2026.1"
```
ScottEilerman: Do we want to use semvers instead of datevers to indicate adaptable vs non-adaptable schemas?

cstar/system/migration.py
```diff
@@ -0,0 +1,258 @@
...
+"""Schema bounds for the hello_world blueprint schema."""
+rm_bounds: SchemaBounds = {
+    "min": APP_ROMS_MARBL_SCHEMA_2025_1,
+    "max": APP_ROMS_MARBL_SCHEMA_2026_1,
+}
+"""Schema bounds for the roms_marbl blueprint schema."""
+default_schema_bounds: Mapping[str, SchemaBounds] = {
+    APP_HW: hw_bounds,
+    APP_ROMS_MARBL: rm_bounds,
+}
```
ScottEilerman, line 50: In the next-up AppDefs ticket, I'd encourage attached schema bounds to the AppDefs as well. No action needed in this PR, just noting it!

cstar/system/migration.py
ScottEilerman: To elaborate (and this maybe is a bastardization of semver), I'd treat X.Y.Z as:
* X indicates a breaking change that is not adaptable (a `1.y.z` schema cannot be turned into a `2.y.z` schema)
* Y indicates a change that can be run through an adapter (if the latest schema is `1.4.z`, any `1.y.z` can be used, though you'll end up with a different but equivalent blueprint file)
* Z indicates an addition or change where no adapter is needed at all, usually an addition of a new parameter that has a default value. If you have a `1.4.0` blueprint, you can run it without modification if the current version is `1.4.1`.

cstar/system/migration.py
```diff
@@ -0,0 +1,258 @@
...
+        return APP_ROMS_MARBL_SCHEMA_2026_1
+
+    @t.override
+    def adapt(self) -> dict[str, str]:
+        runtime_params = t.cast(
+            "dict[str, str]",
+            self.model.get("runtime_params", {"output_dir": None}),
+        )
+        if output_dir := runtime_params.pop("output_dir", None):
+            self.model["output_dir"] = output_dir
```
ScottEilerman: While we're standardizing, is `output_dir` the descriptor we want to go with on the new model? `work_dir`? `root_dir`? It's where the "output" of the entire step goes, even if some of that "output" is collecting/generating inputs... it's the "working directory" except we also put another `work` dir underneath it... 🤔

cstar/orchestration/models.py
```diff
@@ -351,6 +351,9 @@ class Blueprint(ConfiguredBaseModel, ABC):
     state: BlueprintState = BlueprintState.NotSet
     """The current validation status of the blueprint."""
 
+    version: str = ""
```
ScottEilerman: Should we use a more descriptive `schema_version` or `blueprint_version`?

cstar/additional_files/templates/bp/roms_marbl/blueprint.1.0.0.yaml
ankona: yes, ideally I would publish the schema and reference a remote URL. I don't have the schema published yet. I've created a follow-up to publish the schemas to `readthedocs` and reference them in the samples (see: CSD-853)

cstar/system/migration.py
ankona: i'm late to read this, but it's on the radar.

cstar/system/migration.py
ankona: I like `working_dir` or `work_dir` - maybe we need to rename `JobFileSystemManager.work_dir` to something that creates less of a conflict (e.g. `assets_dir`). I've created `CSD-874` as a follow-up to review the naming and update.

cstar/orchestration/models.py
ankona: updated to `schema_version`

cstar/system/migration.py
ankona: swapped to semver. OG @ `1.0.0` and updated to `2.0.0` with updated output path

PR #518, opened 2026-05-06
Fix duplicate log messages

ScottEilerman:
# Summary

This PR fixes several duplicated log messages that occur at the end of a run. The root cause is due to logger parent/child relationships based on name: because SimulationRunner is under `cstar.entrypoint.worker.worker.SimulationRunner`, it ends up propagating messages to `cstar.entrypoint.worker.worker` (which gets established by `get_logger(__name__)` in the `execute_runner` function). Adding a `.runner` to the name for that logger, it breaks the propagation chain. We could also tweak how we handle propagation, but I was worried about additional consequences w.r.t. things like file handlers. This problem might also clear up if we move `SimulationRunner` to a different location soon as part of the app refactor.

## Bug Fixes

- Fix duplicate log messages from the worker/runner after a ROMS Simulation completes

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [x] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/cstar/entrypoint/worker/worker.py
+++ b/cstar/entrypoint/worker/worker.py
@@ -267,7 +267,7 @@ async def execute_runner(
     request : XRunnerRequest[RomsMarblBlueprint]
         A request specifying the blueprint to be executed
     """
-    log = get_logger(__name__, level=service_cfg.log_level)
+    log = get_logger(f"{__name__}.runner", level=service_cfg.log_level)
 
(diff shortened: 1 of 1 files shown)


ankona (changes requested):
All the app code is being relocated. Let's hold until we confirm it's still required.

ankona:
Confirmed via manual testing that this is already fixed

PR #522, opened 2026-05-20
Replace single-path resolver with scoped-resolver

ankona:
# Summary

This PR replaces the use of _path-resolver_ in the template filling transform with a renamed `scoped_resolver`. The new resolver enables the resolution of all file paths exposed by the `JobFileSystemManager` instead of exposing only the `work_dir`. 

## Breaking Changes

- Use of `{{path: <step-name>}}` in templates is no longer supported. Use `{{work_dir: <step-name>}}` instead

## New Features

- N/A

## Bug Fixes

- N/A

## Improvements

- N/A

## Miscellaneous

- N/A

## Security Fixes

- N/A

## Code Review Checklist

- [X] Closes #CSD-921
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/orchestration/transforms.py
```diff
@@ -168,15 +169,75 @@ def get_time_slices(
...
+        p
+        for p in dir(next(iter(fsm_map.values())))
+        if not p.startswith("_") and p.endswith("_dir")
+    ]
+    fsm_attrs.append("root")
+    resolution_map = {
+        f"{step_name}_{fsm_attr}": str(getattr(fsm, fsm_attr))
+        for step_name, fsm in fsm_map.items()
+        for fsm_attr in fsm_attrs
+    }
```
ScottEilerman: Is there a _purpose_ (ha ha) to building out the entire map to all the steps and attrs if we know exactly which one to look for?

cstar/orchestration/transforms.py
```diff
@@ -243,18 +311,18 @@ def _resolve(self, content: str) -> str:
...
             If the appropriate resolver has not been provided.
         """
-        if content.startswith("path:"):
-            step_name = content[len("path:") :].strip()
-            if self._path_resolver is None:
-                raise ValueError(
-                    f"No path resolver provided for placeholder '{{{{path: {step_name}}}}}'"
-                )
-            return str(self._path_resolver(step_name))
+        parts = content.split(":", maxsplit=1)
```
ScottEilerman, line 311: Thinking through whether it's a problem that `:` is now effectively a forbidden character for runtime variables... it's probably fine, I can't imagine what sort of variable name would have a `:` in it, and it would probably wreck who knows what in other parts of the system anyway. No work needed, just mentally noting it.

cstar/orchestration/transforms.py
ankona: _derp!_ not at this time. fixed by filtering both step name and purpose

ScottEilerman: I guess my question was actually more, can't we skip the dict comprehension and subsequent lookup and go straight to:

```python
if return_val := getattr(fsm_map.get(name, {}), scope, None):
    return return_val

... # handle missing attribute / missing fsm / etc.
```
(or break up that one-liner if you want to check the fsm exists in fsm_map first or whatever)

ScottEilerman (approved):
One further suggestion/clarification on my previous comment, but fine to merge as-is or after an update, your preference.

cstar/orchestration/transforms.py
ankona: I think I was stuck on the idea of combining resolvers and just had a forest and trees moment. I've removed the useless dict-comprehension.

PR #523, opened 2026-05-20
Use application registrations to locate and apply transforms

ankona:
# Summary

This change removes the temporary hack/workaround used to avoid applying undesired transforms to applications they cannot be applied to.

## Breaking Changes

- The `WorkplanTransformer` no longer receives transforms to the `__init__` method

## New Features

- The `OverrideTransform` is now applied automatically after all registered transforms and does not need to be registered for the application.

## Bug Fixes

- N/A

## Improvements

- Replaced variable length tuple typehints with simpler `Sequence`

## Miscellaneous

- N/A

## Security Fixes

- N/A

## Code Review Checklist

- [X] Closes #CSD-817
- [X] Tests added
- [X] Tests passing
- [X] Full type hint coverage


cstar/orchestration/transforms.py
```diff
@@ -557,20 +546,37 @@ def suffix() -> str:
...
+    that will be applied to orchestrated steps.
+
+    Returns
+    -------
+    LiveStep
+        The transformed step.
+    """
+    if step.application == "roms_marbl":
+        return {"runtime_params": {"output_dir": step.fsm.root}}
+    return {}
```
ScottEilerman, line 560: Assumption: in one of the other PRs, this function is going to get updated to be used for all applications, and for the top-level `working_dir` attribute, right?

ScottEilerman (approved):
Approved as long as my assumption is correct!

cstar/orchestration/transforms.py
ankona: yes, exactly. just a slave to order-of-operations, here

PR #527, opened 2026-05-21
Fix workdir clear() behavior (again)

ScottEilerman:
# Summary

As previously discussed, we probably shouldn't have been calling `super().clear()` here -- that was resulting in the logs dir being trashed, among other things, which got rid of important cstar logs.

We _do_ need to clear some things from the `work` (or `run` dir in a newer PR). However, we also store overridden blueprints there, and we _don't_ want to wipe those.

This fix gets rid of the`super().clear()` and adds a sorta-messy conditional clear of the work dir (everything except  yamls).

At some point we may want to put workflow artifacts in a different location that the step's run-artifact dir.

## Bug Fixes

- Fix bug where CSTAR_CLOBBER_WORK_DIR would also delete important workplan artifacts

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/cstar/execution/file_system.py
+++ b/cstar/execution/file_system.py
@@ -350,8 +350,6 @@ class RomsFileSystemManager(JobFileSystemManager):
 
     def clear(self) -> None:
         """Ensure the job's working directories are empty."""
-        super().clear()
-
         for directory in [
             self.compile_time_code_dir,
             self.runtime_code_dir,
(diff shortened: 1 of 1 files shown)


ankona (approved):
As a workaround, this looks _fine_.

PR #551, opened 2026-06-15
Fix dropped lines from ROMS logs

ScottEilerman:
# Summary

Somewhat following the exit-code problem, if ROMS crashes fast, we often do not capture the tail of the ROMS log in the C-Star log, meaning the user needs to go open a different log to see what happened. I was going to start mucking with the order of operations of some of the returns, but asked Claude to take a look. It came up with three related issues and proposed the fixes in this PR.

```
The bug (three compounding tail-loss paths)

  In ExecutionHandler.updates():
  1. In-loop: the loop checked self.status != RUNNING and returned before logging the line it had just read — so on the terminal-transition iteration it dropped that line and stopped reading any remaining buffered lines.
  2. Between polls: the worker calls updates(seconds=1.0) repeatedly, and each call did f.seek(0, 2) (jump to EOF), so anything written between polls was skipped entirely.
  3. Already-terminal call: if the job finished between polls, the next call hit the top-of-method branch and only logged "see the output file" — never forwarding the actual tail.

The fix (cstar/execution/handler.py)

  - Log-then-check-then-drain: the streaming loop now logs every available line first, and only when caught up to EOF does it check status; on a terminal status it waits briefly for the final flush, drains the remainder to EOF,
  then returns.
  - Tracked read position (self._log_read_position, lazily initialized): each call resumes exactly where the last left off instead of seeking to EOF — so nothing between polls is skipped and nothing is double-logged. Starts at
  the beginning of the file for full capture (safe since the worker is the sole caller).
  - Terminal-at-top now forwards the tail: before the "see the file" warning, it drains any not-yet-forwarded output to the log.
  - Added a _forward_available(file_handle) helper for the drain-to-EOF logic.
  - Cleanup: removed a duplicated PENDING block that used a blocking time.sleep inside this async method.
```

## Bug Fixes

- C-Star log stream should now more fully capture ROMS log stream

## Code Review Checklist

- [ ] Closes #xxxx
- [ ] Tests added
- [ ] Tests passing
- [ ] Full type hint coverage

### DIFF
--- a/cstar/execution/handler.py
+++ b/cstar/execution/handler.py
@@ -79,6 +79,15 @@ class ExecutionHandler(ABC, LoggingMixin):
         Stream live updates from the task's output file for a specified duration.
     """
 
+    _log_position: int = 0
+    """The current position in the log file.
+
+    Log position is updated after each successful read. It enables resuming reads
+    after relinquishing control to the calling process.
+    """
+    _enabled: bool = True
+    """Flag used to disable processing update requests after the task has terminated."""
+
     @property
     @abstractmethod
     def status(self) -> ExecutionStatus:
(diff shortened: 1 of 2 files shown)


ankona (changes requested):
Edits over time have resulted in spaghetification of this class. [Here's a tweak](https://github.com/CWorthy-ocean/C-Star/pull/559/changes) that should accomplish the same things but also improves readability.

ankona (approved):
LGTM

PR #587, opened 2026-07-11
Prepare models for strict validation

ankona:
# Summary

This change includes some prerequisites and clean-up of pydantic models before implementing strict model validation.

## Breaking Changes

- `LiveStep.working_dir` is no longer optional

## New Features

- N/A

## Bug Fixes

- Fix bug caused by computed properties resulting in incorrect serialization behavior
- Fix bug where applying a `Directive` writes outputs to the incorrect directory

## Improvements

- Mitigate type and linter warnings
- Enable discriminated deserialization of workplans containing `Step` subtypes
- Add test verifying serialization and deserialization of Workplans
- Remove unnecessary multiple inheritance
- Upgrade `pydantic` pin, replace `SerializeAsAny` with use of `polymorphic_serialization=True`
- Convert `LiveStep` attributes to `frozen`

## Miscellaneous

- N/A

## Code Review Checklist

- [X] Supports #CSD-938
- [X] New functionality has documentation, type hint coverage, and tests
- [X] Tests and `pre-commit run --all-files` are passing


cstar/orchestration/orchestration.py
```diff
@@ -307,25 +311,65 @@ def from_step(
...
+            parent = LiveStep(**parent.model_dump(by_alias=True))
+
+        if parent:
+            fsm = parent.fsm.get_subtask_manager(name)
+            wd = fsm.root_dir
+            data["parent"] = parent
+
+        if wd:
+            data["working_dir"] = wd
+            fsm = JobFileSystemManager(Path(wd))
```
ankona: useless LoC?

ankona: removed

cstar/orchestration/models.py
```diff
@@ -226,7 +230,7 @@ class Step(BaseModel):
     application: RequiredString
     """The user-friendly name of the application executed in the step."""
 
-    blueprint_path: FilePath | str = Field(alias="blueprint")
+    blueprint_path: FilePath = Field(alias="blueprint")
```
ScottEilerman: Is this going to choke on remote blueprints? I know we have that "as local" wrapper but I don't remember if that hits before or after this. I think before... at least in the CLI. I noticed you removed some remote blueprints from some of the tests, so it might be worth double-checking.

ScottEilerman (approved):
approved as long as remote blueprints are working ok

ScottEilerman (approved):
approved as long as remote blueprints are working ok

cstar/orchestration/models.py
ankona: i confused myself here. reverted. good catch

PR #605, opened 2026-07-24
Skip excessive sha256 calculations

ScottEilerman:
# Summary

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- N/A

## Improvements

- Speed up simple data fetch operations by only validating sha256/stat if a value was explicitly passed

## Miscellaneous

- N/A

## Code Review Checklist

- [ ] Closes #xxxx

### DIFF
--- a/cstar/io/staged_data.py
+++ b/cstar/io/staged_data.py
@@ -131,6 +131,14 @@ class StagedFile(StagedData):
         if not resolved_path.exists():
             return True
 
+if stat := self._stat:
+    r_stat = os.stat(resolved_path.resolve())
+
+    if stat.st_mtime != r_stat.st_mtime:
+        return True
+
+    if stat.st_size != r_stat.st_size:
+        return True 
         if (
             self._stat
             and self._stat.st_mtime != os.stat(resolved_path.resolve()).st_mtime
--- a/cstar/io/staged_data.py
+++ b/cstar/io/staged_data.py
@@ -116,15 +116,9 @@ class StagedFile(StagedData):
         """
         super().__init__(source, path)
 
-        if sha256:
-            self._sha256 = sha256
-        else:
-            self._sha256 = _get_sha256_hash(self.path)
+        self._sha256 = sha256
 
-        if stat:
-            self._stat = stat
-        else:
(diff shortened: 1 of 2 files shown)


cstar/io/staged_data.py
```diff
@@ -136,11 +130,18 @@ def changed_from_source(self) -> bool:
         resolved_path = self._path.resolve()
         if not resolved_path.exists():
             return True
-        if self._stat.st_mtime != os.stat(resolved_path.resolve()).st_mtime:
+
```
ankona, line 133:
```suggestion

if stat := self._stat:
    r_stat = os.stat(resolved_path.resolve())

    if stat.st_mtime != r_stat.st_mtime:
        return True

    if stat.st_size != r_stat.st_size:
        return True 
```

you might be able to avoid a file system operation by computing `os.stat(resolved_path.resolve())` ahead of the conditionals, and can improve readability a bit.

cstar/io/staged_data.py
```diff
@@ -136,11 +130,18 @@ def changed_from_source(self) -> bool:
...
+        ):
             return True
-        if self._stat.st_size != os.stat(resolved_path.resolve()).st_size:
+        if (
+            self._stat
+            and self._stat.st_size != os.stat(resolved_path.resolve()).st_size
+        ):
             return True
-        if self._sha256 != _get_sha256_hash(resolved_path.resolve()):
+        if self._sha256 and self._sha256 != _get_sha256_hash(resolved_path.resolve()):
```
ankona, line 143: consider tweaking `_get_sha256_hash` to accept the size and creation date, then adding `@lru_cache` - we may avoid recomputing these hashes even more often

(a method wrapping `_get_sha256_hash` is probably a better solution than adding those params to this method)

ankona (approved):
overall LGTM. i have a couple minor suggestions you can take or leave

cstar/io/staged_data.py
ScottEilerman: Good idea -- going to skip it for now just because I'm already in the middle of something else and it _shouldn't_ get triggered in our current workflows. But I'll add a backlog issue.

PR #615, opened 2026-08-05
Add support for registering external apps

ScottEilerman:
# Summary

This feature is currently needed to for some in-development work in Forge.

## New Features

- Add support for registration of external apps via the CSTAR_APP_MODULES env-var

## Code Review Checklist

- [ ] Closes #xxxx

### DIFF
--- a/cstar/applications/core.py
+++ b/cstar/applications/core.py
@@ -1,5 +1,4 @@
 import importlib
-import os
 import typing as t
 from collections.abc import Sequence
 from dataclasses import dataclass, field
@@ -8,12 +7,14 @@ from itertools import chain
 from pathlib import Path
 
 from cstar.base.adapter import SchemaAdapter
+from cstar.base.env import get_env_item
 from cstar.base.log import get_logger
 from cstar.entrypoint.config import JOBFILE_DATE_FORMAT
 from cstar.execution.file_system import local_copy
 from cstar.execution.handler import ExecutionStatus
 from cstar.orchestration.models import BlueprintCore
 from cstar.orchestration.serialization import SerializableModel, deserialize
+from cstar.orchestration.utils import ENV_CSTAR_APP_MODULES
 
 if t.TYPE_CHECKING:
     from cstar.entrypoint.config import JobConfig, ServiceConfiguration
@@ -375,7 +376,8 @@ def get_application(name: str) -> ApplicationDefinition[t.Any, t.Any]:
     before lookup so its ``@register_application`` decorators run.
     """
     if name not in _registry:
-        for module in filter(None, os.environ.get("CSTAR_APP_MODULES", "").split(",")):
+        modules = get_env_item(ENV_CSTAR_APP_MODULES).value
+        for module in filter(None, modules.split(",")):
             importlib.import_module(module.strip())
 
     if name not in _registry:
--- a/cstar/applications/core.py
+++ b/cstar/applications/core.py
@@ -376,11 +376,10 @@ def get_application(name: str) -> ApplicationDefinition[t.Any, t.Any]:
     before lookup so its ``@register_application`` decorators run.
     """
     if name not in _registry:
-        modules = get_env_item(ENV_CSTAR_APP_MODULES).value
-        for module in filter(None, modules.split(",")):
-            importlib.import_module(module.strip())
+        if modules := get_env_item(ENV_CSTAR_APP_MODULES).value:
+            for module in filter(lambda x: name in x, modules.split(",")):
+                importlib.import_module(module.strip())
 
-    if name not in _registry:
         try:
             importlib.import_module(f"cstar.applications.{name}")
         except ModuleNotFoundError:
--- a/cstar/applications/core.py
+++ b/cstar/applications/core.py
@@ -377,13 +377,27 @@ def get_application(name: str) -> ApplicationDefinition[t.Any, t.Any]:
     """
     if name not in _registry:
         if modules := get_env_item(ENV_CSTAR_APP_MODULES).value:
-            for module in filter(lambda x: name in x, modules.split(",")):
-                importlib.import_module(module.strip())
+            if matches := {m.strip() for m in modules.split(",") if name in m}:
+                if len(matches) > 1:
+                    msg = f"An application name collision may occur using {name!r} for modules {','.join(matches)}"
+                    log.warning(msg)
+
+                module = next(iter(matches))
+
+                try:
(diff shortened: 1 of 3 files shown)


cstar/applications/core.py
```diff
@@ -365,9 +367,23 @@ def get_application(name: str) -> ApplicationDefinition[t.Any, t.Any]:
...
+    Notes
+    -----
+    Applications defined outside the ``cstar.applications`` package can be made
+    discoverable by setting the ``CSTAR_APP_MODULES`` environment variable to a
+    comma-separated list of importable module paths. Each module is imported
+    before lookup so its ``@register_application`` decorators run.
     """
     if name not in _registry:
-        importlib.import_module(f"cstar.applications.{name}")
+        if modules := get_env_item(ENV_CSTAR_APP_MODULES).value:
```
ankona, line 379: A name should uniquely identify a single application. This loading behavior will likely produce unexpected failures by allowing multiple 3rd party plugins to map to the same name. We should always only load a single thing here.

Consider:
- This loading behavior bypasses the intent of the lazy loader and can cause unnecessary imports.
- `export CSTAR_APP_MODULES=foo.analyze_app,bar.analyze_app` - Using `get_application(analyze_app)` will fail "randomly" based on the order of the values in the env var.
- `export CSTAR_APP_MODULES=foo.analyze_app,foo.other_app` - Using `get_application('foo')` results in a seemingly successful priming of the registry, but the same mapping failure
- `pip uninstall foo`
- the fallthrough behavior looks like it would allow 3rd parties to use internal app names and hijack expected implementations

**Suggestion**

In this suggested alternative, I:
- log a warning for the multiple matches to proactively help solving name collisions.
- never load more than one module per key
- log a warning message in case we have key issues (app was not able to be loaded even if name matches a module name in the env var) that identifies the difference between a bad name and bad module
- log a warning message that identifies the load failure as an internal failure (and helps debug "I didn't have that in the env var" problems)

```python

def get_application(name: str) -> ApplicationDefinition[t.Any, t.Any]:
    """Get an application from the application registry.

    Returns
    -------
    Application
        The application matching the supplied name

    Raises
    ------
    ValueError
        if no registered application is associated with this classification

    Notes
    -----
    Applications defined outside the ``cstar.applications`` package can be made
    discoverable by setting the ``CSTAR_APP_MODULES`` environment variable to a
    comma-separated list of importable module paths. Each module is imported
    before lookup so its ``@register_application`` decorators run.
    """
    if name not in _registry:
        if modules := get_env_item(ENV_CSTAR_APP_MODULES).value:
            if matches := {m.strip() for m in modules.split(",") if name in m}:
                if len(matches) > 1:
                    msg = f"An application name collision may occur using {name!r} for modules {','.join(matches)}"
                    log.warning(msg)

                module = next(iter(matches))

                try:
                    importlib.import_module(module)
                except ModuleNotFoundError:
                    log.warning(
                        f"Unable to load external application {name!r} from {module!r}"
                    )
                    raise

        module = f"cstar.applications.{name}"

        try:
            importlib.import_module(module)
        except ModuleNotFoundError:
            log.warning(f"Unable to load C-Star application {name!r} from {module!r}")

    if application := _registry.get(name):
        log.trace(f"Located application context {application.__name__!r} for {name!r}")
        return application()

    msg = f"No application for {name!r}"
    raise ValueError(msg)
```

ankona (changes requested):
Please review comment re: app name collisions

cstar/applications/core.py
ankona: We may alternatively consider using a fully-qualified name instead of our original short name to improve this...

If `foo.roms_marbl`, `bar.roms_marbl`, and `cstar.applications.roms_marbl` are the actual application names, name collisions should no longer be an issue.

cstar/orchestration/utils.py
```diff
@@ -17,6 +17,18 @@
 ] = "CSTAR_ORCH_DELAYS"
 """Environment variable containing configurable delay for the orchestrator."""
 
+ENV_CSTAR_APP_MODULES: t.Annotated[
+    t.Literal["CSTAR_APP_MODULES"],
+    EnvVar(
+        "A comma-separated list of importable module paths for applications defined "
+        "outside the `cstar.applications` package. Each module is imported before "
```
ankona, line 24: Consider adding an extended description to `EnvVar` for the second sentence here (or `EnvVar.info` or `EnvVar.details` ...). The output from `cstar env show` is kind of cumbersome (ditto on docs page output).

Also... The convention here should be "the triple quote description matches the description in the EnvVar" and I may have broken it before you. Consider just making the triple-quote and first sentence match to keep that convention going.

cstar/orchestration/utils.py
```diff
@@ -17,6 +17,18 @@
...
+        "application lookup so its `@register_application` decorators run.",
+        _GROUP_ORCH,
+        "",
+    ),
+] = "CSTAR_APP_MODULES"
+"""A comma-separated list of importable module paths for externally-defined applications."""
```
ankona:
```suggestion
"""A comma-separated list of package names where external applications are defined."""
```
Just to be pedantic, I dislike saying "importable module paths" here because "path" is not "fully qualified name" and may imply _file system path_. Also, it's inconsistent with the `EnvVar` (module vs package verbiage)

cstar/tests/unit_tests/applications/test_core.py
```diff
@@ -1,11 +1,85 @@
...
+                long_name: str = "Externally Defined App"
+                runner = HelloWorldRunner
+                blueprint = HelloWorldBlueprint
+                applicable_transforms = ()
+            """
+        )
+    )
+
+    monkeypatch.syspath_prepend(str(tmp_path))
+    monkeypatch.setenv("CSTAR_APP_MODULES", f"{app_name}_module")
```
ankona: consider using `ENV_CSTAR_APP_MODULES` since it's already imported and we can tweak the name later without updating the test
```suggestion
    monkeypatch.setenv(ENV_CSTAR_APP_MODULES, f"{app_name}_module")
```

cstar/applications/core.py
ScottEilerman: adopted your version of the function, thanks!

cstar/orchestration/utils.py
ScottEilerman: punting for now, we can clean these up in the pre-launch docs pass

ankona (approved):
LGTM, thanks for the quick fix.

PR #629, opened 2026-08-17
Change continue dir for PIO; fix latent sorting bug

ScottEilerman:
# Summary

Previous behavior was to use pre-partitioned restart files from the `output` directory when continuing. With PIO, those outputs never get made, and we should instead look in the `joined_outputs` directory for whole restart files.

While fixing this behavior, another bug was discovered -- we pulled the first file from a presumed list of partition-indexed files (getting the `.000.nc` one), but we needed an additional filter for the last timestamp if multiple timestamps were being listed.

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- Fixes a bug where the `continue-from` directive would look in the wrong directory if PIO is enabled
- Fixes a bug where the `continue-from` directive would obtain the wrong file if partitioning was enabled and multiple restart file timestamps were present

## Improvements

- N/A

## Miscellaneous

- N/A

## Code Review Checklist

- [ ] Closes #xxxx

### DIFF
--- a/cstar/applications/roms_marbl/transforms.py
+++ b/cstar/applications/roms_marbl/transforms.py
@@ -202,6 +202,8 @@ class RestartFile(BaseModel):
     """The expected file extension for a restart file."""
     FMT_TS: t.ClassVar[t.Literal["%Y%m%d%H%M%S"]] = "%Y%m%d%H%M%S"
     """The expected timestamp format in the restart file name"""
+    TS_GLOB: t.ClassVar[str] = "[0-9]" * 14
+    """Glob fragment matching the fixed-width (14-digit) timestamp (see `FMT_TS`)."""
     PATTERN_RST: t.ClassVar[t.Literal[r"^(.*?)_rst\.(\d{14})(?:\.(\d{1,9}))?\.nc$"]] = (
         r"^(.*?)_rst\.(\d{14})(?:\.(\d{1,9}))?\.nc$"
     )
@@ -245,29 +247,30 @@ class RestartFile(BaseModel):
             msg = f"No directory or file found at path: {search_path!r}"
             raise ValueError(msg)
 
-        # filter to names that actually parse before constructing (a stray file matching
-        # the glob shape but not `PATTERN_RST` should be skipped, not fatal).
-        matches = [
-            match
-            for match in search_path.rglob(f"*{cls.SUFFIX}.*.*.{cls.EXT}")
-            if re.fullmatch(cls.PATTERN_RST, match.name, flags=re.ASCII)
-        ]
-        if matches:
-            # Prefer pre-partitioned data when available. `matches` may span
-            # several restart timestamps, each with a full set of partition
-            # files, so continue from partition 0 of the most recent timestamp.
-            rst_files = [RestartFile(path=match) for match in matches]
-            latest_ts = max(rst.timestamp for rst in rst_files)
-            return min(
-                (rst for rst in rst_files if rst.timestamp == latest_ts),
-                key=lambda rst: rst.partition or 0,
-            )
+        # Prefer pre-partitioned files (avoids re-partitioning), falling back to
+        # joined (unpartitioned) files. Both globs pin the fixed-width timestamp
+        # but not the variable-width partition segment, so `PATTERN_RST` still
+        # validates each candidate -- skipping stray files that match the glob
+        # shape but not the naming convention rather than treating them as fatal.
+        partitioned_glob = f"*{cls.SUFFIX}.{cls.TS_GLOB}.*.{cls.EXT}"
+        joined_glob = f"*{cls.SUFFIX}.{cls.TS_GLOB}.{cls.EXT}"
 
-        # Joined (unpartitioned) restart files carry no partition segment;
-        # continue from the most recent timestamp.
-        matches = sorted(search_path.rglob(f"*{cls.SUFFIX}*.{cls.EXT}"), reverse=True)
-        if matches:
-            return RestartFile(path=matches.pop(0))
+        for glob_pattern in (partitioned_glob, joined_glob):
+            rst_files = [
+                RestartFile(path=match)
+                for match in search_path.rglob(glob_pattern)
+                if re.fullmatch(cls.PATTERN_RST, match.name, flags=re.ASCII)
+            ]
+            if rst_files:
+                # A directory may hold several restart timestamps (each a full
+                # partition set); continue from partition 0 of the most recent.
+                # Partition 0 is required: downstream reconstructs the remaining
+                # partitions by substituting its suffix in the file name.
+                latest_ts = max(rst.timestamp for rst in rst_files)
+                return min(
+                    (rst for rst in rst_files if rst.timestamp == latest_ts),
+                    key=lambda rst: rst.partition or 0,
+                )
 
         if not notfound_ok:
             msg = f"No restart files located. Unable to continue from {search_path!r}"
@@ -704,6 +707,9 @@ class ContinuanceDirective(OverrideDirective):
             if name in self.workplan:
                 step = self.workplan[name]
                 fsm = RomsFileSystemManager(step.fsm.root_dir)
+
+                search_path = fsm.output_dir
+
                 # With ParallelIO there are no partitioned restart files to
                 # reuse; the joined restart files live in `joined_output`.
                 blueprint = step.blueprint
@@ -712,10 +718,7 @@ class ContinuanceDirective(OverrideDirective):
                     and blueprint.model_params.use_pio
                 ):
                     search_path = fsm.joined_output_dir
-                else:
-                    # fall back to regular output_dir for non-roms blueprints.
-                    # seems unlikely this is a relevant use case, but easy to guard
-                    search_path = fsm.output_dir
+
             else:
                 msg = f"Unable to locate step {name!r} in workplan"
                 raise KeyError(msg)
--- a/cstar/applications/roms_marbl/transforms.py
+++ b/cstar/applications/roms_marbl/transforms.py
@@ -26,6 +26,7 @@ from cstar.base.utils import (
     min_padded_index,
     slugify,
 )
+from cstar.execution.file_system import RomsFileSystemManager
 from cstar.orchestration.orchestration import LiveStep, LiveWorkplan
 from cstar.orchestration.serialization import serialize
 from cstar.orchestration.transforms import (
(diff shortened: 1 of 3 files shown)


cstar/applications/roms_marbl/transforms.py
```diff
@@ -686,7 +703,19 @@ def _generate_overrides(self) -> dict[str, t.Any]:
         if name := self._config.get(self.KEY_STEP, None):
             if name in self.workplan:
                 step = self.workplan[name]
-                search_path = step.fsm.output_dir
+                fsm = RomsFileSystemManager(step.fsm.root_dir)
```
ankona, line 701: we talked about making the steps have the right FSM based on their application before and this reminds me of that conversation... Should we create a follow-up ticket to modify the `LiveWorkplan` and `LiveStep` by making the step a generic, like `LiveStep[RomsFileSystemManager]`? Then, we can rely on the tighter typehints in any roms-specific transforms and won't have to do all this casting and type-interrogation elsewhere

cstar/applications/roms_marbl/transforms.py
```diff
@@ -686,7 +703,19 @@ def _generate_overrides(self) -> dict[str, t.Any]:
...
+                # reuse; the joined restart files live in `joined_output`.
+                blueprint = step.blueprint
+                if (
+                    isinstance(blueprint, RomsMarblBlueprint)
+                    and blueprint.model_params.use_pio
+                ):
+                    search_path = fsm.joined_output_dir
+                else:
+                    # fall back to regular output_dir for non-roms blueprints.
+                    # seems unlikely this is a relevant use case, but easy to guard
```
ankona: i dislike this comment for 2 reasons:

1. "use the default" shouldn't need a block of comment explaining it.
2. it's lying - it's not only "if it's not an RM bp" but "if it's a non-PIO bp"

move `search_path = step.fsm.output_dir` up under `blueprint = step.blueprint` and just let the modification speak for itself?

cstar/applications/roms_marbl/transforms.py
```diff
@@ -244,10 +245,26 @@ def find(cls, search_path: Path, notfound_ok: bool = True) -> "RestartFile | Non
...
 
-        if matches := sorted(search_path.rglob(f"*{cls.SUFFIX}.*.*.{cls.EXT}")):
-            # prefer use of pre-partitioned data when available
-            return RestartFile(path=matches[0])
+        # filter to names that actually parse before constructing (a stray file matching
+        # the glob shape but not `PATTERN_RST` should be skipped, not fatal).
+        matches = [
+            match
+            for match in search_path.rglob(f"*{cls.SUFFIX}.*.*.{cls.EXT}")
+            if re.fullmatch(cls.PATTERN_RST, match.name, flags=re.ASCII)
```
ankona: We have two ways of solving this that should make this less repetitive: use another regex, use another glob. Since you went the glob route, let's encode preference in search order:

```
nopart_glob = f"*{cls.SUFFIX}.*.{cls.EXT}"
parted_glob = f"*{cls.SUFFIX}.*.*.{cls.EXT}"

search_priority = [parted_glob, nopart_glob]

for glob_pattern in search_priority:
  matches = sorted(x for x in search_path.rglob(glob_pattern) if re.fullmatch(cls.PATTERN_RST), reverse=True)
```

I think the reversed sort should handle getting the maximum if the timestamps are 0-padded and what was really missing was the prioritization of non-partitioned matches?

[update] this only works if you use the exact-match glob patterns discussed below. i think the entire comment is subsumed in my follow-up, but leaving just in case

ankona (changes requested):
Let's see if we can simplify the search prioritization

cstar/applications/roms_marbl/transforms.py
ankona: Wait.. Can you confirm that simply fixing the order of the pre-existing assignments to `matches` makes this particular testcase work?

```
# this was second before...
if matches = [search_path.rglob(unpartitioned_glob_pattern, ...)]:
  return RestartFile(path=next(sorted(matches, reverse=True)))

if matches = [search_path.rglob(partitioned_glob_pattern, ...)]:
  # search for partitioned files if we couldn't find unpartitioned data
  return RestartFile(path=next(sorted(matches, reverse=True)))
```

There is a comment in the prior code "# prefer use of pre-partitioned data when available" that points to the need to know the value of `use_pio` to _dynamically set the priority_. 

We could add something like `prioritize: Literal["partitioned", "unpartitioned"] = "partitioned"` as a parameter to `find` and conditionally do `reverse(search_priority)` 
 - but, this is called at instantiation of the directive to create "system overrides", so we must also...
 - Modify `__call__` in `class ContinuanceDirective`:

```
  def __call__(self, step: LiveStep) -> Sequence[LiveStep]:
    priority = "partitioned" if cast("RomsMarblBlueprint", step.blueprint).model_params.use_pio else "unpartitioned"
    self._system_overrides.update(
      self._generate_overrides(prioritize=priority)
    )
    return super().__call__(step)
```

This is a "more complex" fix, but it fixes the actual underlying problem of creating the overrides as system-level overrides before the step is available to determine the priority.

Note: I'd like to tweak `OverrideDirective` to have both `def _generate_overrides` and `def generate_step_overrides` so we could put this machinery onto `OverrideDirective` and it can be re-used instead of needing to add it to each `XxxDirective` that needs to update overrides based on the specifics of the step.

cstar/tests/unit_tests/orchestration/test_transforms.py
```diff
@@ -310,6 +319,13 @@ async def test_continuance_directive_step_resolution(
 
     await create_mocked_simulation_outputs(wp_template_path, live_wp_path, run_id)
 
+    # `read_yaml_intercept` rewrites `local_bp` with the default template on every
+    # workplan read (the last of which is inside `create_mocked_simulation_outputs`),
+    # so inject `use_pio` here to ensure it survives into the directive's blueprint read.
+    bp = deserialize(local_bp, RomsMarblBlueprint)
```
ankona, line 335: consider mocking instead of touching the file two more times:

`with mock.patch.object(RomsMarblBlueprint, "use_pio", PropertyMock(return_value=use_pio):`

cstar/tests/unit_tests/orchestration/test_transforms.py
```diff
@@ -333,11 +349,21 @@ async def test_continuance_directive_step_resolution(
             assert ContinuanceDirective.KEY_STEP in config
             assert ContinuanceDirective.KEY_PATH not in config
 
-            # confirm the initial conditions have been overridden to reference the named step
+            # confirm the initial conditions were overridden to continue from the
+            # latest restart of the named step: under PIO the joined restart file
+            # in `joined_output`, otherwise partition 0 in `output`.
+            prior_fsm = RomsFileSystemManager(prior_step.fsm.root_dir)
```
ankona, line 365: move the `expected_value` into the parameterization and get rid of the conditional `if use_pio` in the test (and let the `pytest.param` be the source of truth about what should happen instead of the embedded comment)

cstar/tests/unit_tests/orchestration/test_transforms.py
```diff
@@ -1005,6 +1031,41 @@ def test_restart_file_find(tmp_path: Path, pad_size: int) -> None:
...
+
+def test_restart_file_find_skips_malformed(tmp_path: Path) -> None:
+    """Verify that a stray file matching the restart glob shape but not the
+    strict naming convention is skipped rather than aborting the search.
+    """
+    search_path = tmp_path / "output"
+    search_path.mkdir(parents=True)
+
+    (search_path / "foo_rst.20120101000000.000.nc").touch()
+    # matches the loose glob `*_rst.*.*.nc` but not `PATTERN_RST` (bad timestamp)
```
ankona: modify the glob pattern in the class to specify exactly the correct number of characters instead of allowing it to match undesirable outputs, e.g. `glob_pattern = "*_rst.??????????????.???.nc"`

(or `datepart = "[0-9]" * 14; gp = "*_rst.{datepart}.???.nc"` if you want to be more specific on numbers in the date)

cstar/applications/roms_marbl/transforms.py
ScottEilerman: I think "prefer pre-partitioned" still holds for PIO because those files simply won't be there if PIO is used, so it will fall back to the joined ones. I guess you could save a glob command by knowing ahead of time, but I don't think that's worth extra prioritization logic and optionality.

cstar/tests/unit_tests/orchestration/test_transforms.py
ScottEilerman: skipping this one for now, seems like mocking a pydantic field is potentially tricky

cstar/tests/unit_tests/orchestration/test_transforms.py
ScottEilerman: the number of digits in the partition portion is truly variable, not always 3 digits. new code locks down the date portion though.

cstar/applications/roms_marbl/transforms.py
ScottEilerman: cleaned up the loops / patterns a bit

cstar/applications/roms_marbl/transforms.py
ScottEilerman: could do, i'll make a ticket.

cstar/tests/unit_tests/orchestration/test_transforms.py
```diff
@@ -310,6 +331,13 @@ async def test_continuance_directive_step_resolution(
 
     await create_mocked_simulation_outputs(wp_template_path, live_wp_path, run_id)
 
+    # `read_yaml_intercept` rewrites `local_bp` with the default template on every
+    # workplan read (the last of which is inside `create_mocked_simulation_outputs`),
+    # so inject `use_pio` here to ensure it survives into the directive's blueprint read.
```
ankona: you should be able to cut this down to "Inject `use_pio` into the serialized blueprint" - imagine if every place that used `read_yaml_intercept` regurgitated what it did... tons of useless comments.

cstar/applications/roms_marbl/transforms.py
```diff
@@ -244,13 +247,30 @@ def find(cls, search_path: Path, notfound_ok: bool = True) -> "RestartFile | Non
...
             raise ValueError(msg)
 
-        if matches := sorted(search_path.rglob(f"*{cls.SUFFIX}.*.*.{cls.EXT}")):
-            # prefer use of pre-partitioned data when available
-            return RestartFile(path=matches[0])
-
-        matches = sorted(search_path.rglob(f"*{cls.SUFFIX}*.{cls.EXT}"), reverse=True)
-        if matches:
-            return RestartFile(path=matches.pop(0))
+        # Prefer pre-partitioned files (avoids re-partitioning), falling back to
```
ankona: These are probably good details to keep around but we should put them in the docstring so consumers know this instead of having to inspect internals.

cstar/applications/roms_marbl/transforms.py
```diff
@@ -244,13 +247,30 @@ def find(cls, search_path: Path, notfound_ok: bool = True) -> "RestartFile | Non
...
+        joined_glob = f"*{cls.SUFFIX}.{cls.TS_GLOB}.{cls.EXT}"
+
+        for glob_pattern in (partitioned_glob, joined_glob):
+            rst_files = [
+                RestartFile(path=match)
+                for match in search_path.rglob(glob_pattern)
+                if re.fullmatch(cls.PATTERN_RST, match.name, flags=re.ASCII)
+            ]
+            if rst_files:
+                # A directory may hold several restart timestamps (each a full
```
ankona: ditto for moving to docstring

cstar/applications/roms_marbl/transforms.py
```diff
@@ -686,7 +706,19 @@ def _generate_overrides(self) -> dict[str, t.Any]:
         if name := self._config.get(self.KEY_STEP, None):
             if name in self.workplan:
                 step = self.workplan[name]
-                search_path = step.fsm.output_dir
+                fsm = RomsFileSystemManager(step.fsm.root_dir)
+
+                search_path = fsm.output_dir
+
+                # With ParallelIO there are no partitioned restart files to
```
ankona, line 705: consider just cutting to "Unpartitioned ParallelIO outputs are written to an alternate location"

ankona (approved):
Approved, but let's clean up a few comments that are overly verbose before merging.

PR #632, opened 2026-08-19
Normalize run-id casing at intake to fix mixed-case run breakage

ScottEilerman:
# Summary

A mixed-case --run-id was used raw for run directories and tracking records but casefolded (slugify) for CSTAR_RUNID by configure_environment, producing two runs directories, mismatched cache/tracking entries, and status/rerun lookups that could not find the run record. Slugify the run-id once at CLI intake and in build_and_run_dag, and normalize TrackingRepository lookups so mixed-case queries resolve to the stored record.

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- Normalize run-id casing at intake to fix mixed-case run breakage

## Improvements

- N/A

## Miscellaneous

- N/A

## Code Review Checklist

- [ ] Closes #xxxx

### DIFF
--- a/cstar/cli/common.py
+++ b/cstar/cli/common.py
@@ -16,6 +16,7 @@ from cstar.base.env import (
 )
 from cstar.base.feature import is_flag_enabled
 from cstar.base.log import LogLevelChoices, get_logger, reset_log_level
+from cstar.base.utils import slugify
 from cstar.execution.file_system import DirectoryManager, is_remote_resource
 from cstar.orchestration.models import BlueprintCore
 from cstar.orchestration.serialization import (
@@ -127,6 +128,35 @@ def set_flag(
         return value
 
     return _callback
+
+
+def normalize_runid(_context: typer.Context, run_id: str) -> str:
+    """Normalize a user-supplied run-id (slugify) so the same identifier is
+    used everywhere (directories, tracking records, environment).
+
+    Empty values are returned unchanged so callers can apply their own
+    presence/default handling.
+
+    Parameters
+    ----------
+    _context : typer.Context
+        The typer context (unused).
+    run_id : str
+        The run-id value received from the user.
+
+    Returns
+    -------
+    str
+    """
+    if not (run_id := run_id.strip()):
+        return run_id
+
+    normalized = slugify(run_id)
+    if normalized != run_id:
+        msg = f"Normalized run-id `{run_id}` to `{normalized}`"
+        log.info(msg)
+
+    return normalized
 
 
 def set_env(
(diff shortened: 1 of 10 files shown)


cstar/cli/common.py
```diff
@@ -129,6 +130,35 @@ def _callback(ctx: typer.Context, value: bool) -> bool:
...
+    -------
+    str
+    """
+    if not (run_id := run_id.strip()):
+        return run_id
+
+    normalized = slugify(run_id)
+    if normalized != run_id:
+        msg = f"Normalized run-id `{run_id}` to `{normalized}`"
+        log.info(msg)
```
ankona: do you want this at `info` level or `debug`?

cstar/orchestration/tracking.py
ankona, line 292: slugify should already strip whitespace. consider replacing this line instead of adding below

cstar/orchestration/tracking.py
```diff
@@ -326,6 +328,8 @@ def get_workplan_run_sync(
             msg = "A valid run-id was not provided; unable to retrieve run"
             raise ValueError(msg)
 
+        run_id = slugify(run_id)
```
ankona, line 331: consider moving to before `if not run_id` check like `get_workplan_run`

cstar/cli/workplan/run.py
```diff
@@ -279,8 +282,8 @@ def preprocess_runid(ctx: typer.Context, run_id: str) -> str:
     -------
     str
     """
-    if run_id := run_id.strip():
-        return run_id
+    if normalized := normalize_runid(ctx, run_id):
```
ankona: consider putting into the pre-existing `cb_pipeline(...)` instead of nesting calls

ankona (approved):
LGTM. nitpicks only...

- considered renaming callback since it's not _really_ runid specific, but who cares.
- considered suggesting adding the value logging in the `cb_pipeline` so all values are logged at the end

cstar/orchestration/tracking.py
ScottEilerman: I guess it would drop a string with spaces to an empty string that would trigger the next error line? slugify has an error too if the string is empty, but it doesn't provide any error message. leaving for now.

cstar/orchestration/tracking.py
ScottEilerman: same, I think we'd lose the helpful error message. maybe slugify shouldn't raise and just return an empty string if it gets an empty input?

ScottEilerman:
Fixed two of the comments, punted on two that would result in less informative error messages. We might want to revisit the bare ValueError from slugify, but that can be separate from this PR if desired.

PR #633, opened 2026-08-19
Prune already-completed SLURM dependency jobs on every launch

ScottEilerman:
# Summary

On workplan rerun, a finished step is not resubmitted (Prefect cache returns its old job id), but a dependent step submitted for the first time still passed `--dependency=afterok:<old_jobid>`, which SLURM rejectsor kills once the job is gone. The completed-dependency pruning only ran when the dependent step itself had a failed prior sentinel; it now runs unconditionally before every submission, querying the dependency job ids directly.

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- Depending on already-completed SLURM jobs should now seamlessly drop the dependencies

## Improvements

- N/A

## Miscellaneous

- N/A

## Code Review Checklist

- [ ] Closes #xxxx

### DIFF
--- a/cstar/orchestration/launch/slurm.py
+++ b/cstar/orchestration/launch/slurm.py
@@ -1,7 +1,6 @@
 import asyncio
 import os
 import typing as t
-from collections.abc import Mapping
 
 from prefect import State, task
 from prefect import Task as PrefectTask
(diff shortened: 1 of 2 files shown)


cstar/orchestration/launch/slurm.py
```diff
@@ -380,22 +403,10 @@ async def launch(
...
-                if dependencies and successes:
-                    reusable = set(x.pid for x in dependencies).intersection(successes)
-                    msg = f"Dependencies previously satisfied: {', '.join(reusable)}"
-                    log.info(msg)
-
-                    # only keep dependencies that are not re-usable
-                    active = set(x.pid for x in dependencies).difference(successes)
-                    dependencies = list(filter(lambda x: x.pid in active, dependencies))
+        # always drop dependencies on jobs that already completed successfully
+        # (e.g. steps satisfied by a prior run) so SLURM does not reject the
```
ankona: the docstring already tells us every word in this and the new method name is much nicer. let's dump this whole comment block.

ankona (approved):
one annoying comment block duplicating a docstring to remove, all else G2G.

PR #635, opened 2026-08-19
Add --clobber <step> to cstar workplan run for per-step reruns

ScottEilerman:
# Summary

Add the option to clobber individual 

## Breaking Changes

- CSTAR_CLOBBER_WORKING_DIR is no longer used to clobber all workplan steps. Use `cstar workplan run workplan.yaml --clobber all` instead.

## New Features

- Add the option to clobber individual steps with `--clobber <step name>`. Clobbering all steps is achieved via `--clobber all`.

## Bug Fixes

- N/A

## Improvements

- N/A

## Miscellaneous

- N/A

## Code Review Checklist

- [ ] Closes #xxxx

### DIFF
--- a/cstar/base/env.py
+++ b/cstar/base/env.py
@@ -2,7 +2,7 @@ import os
 import sys
 import typing as t
 from collections import defaultdict
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from functools import lru_cache
@@ -152,6 +152,28 @@ def generate_run_id() -> str:
     return datetime.now(UTC).strftime("%Y%m%d_%H%M%S")
 
 
+def get_clobber_steps() -> frozenset[str]:
+    """Return the set of step safe_names selected for per-step clobber.
+
+    Returns
+    -------
+    frozenset[str]
+    """
+    raw = os.environ.get(ENV_CSTAR_CLOBBER_STEPS, "")
+    return frozenset(x for entry in raw.split(",") if (x := entry.strip()))
+
+
+def set_clobber_steps(values: Iterable[str]) -> None:
+    """Set `CSTAR_CLOBBER_STEPS` to the comma-encoding of `values`.
+
+    Parameters
+    ----------
+    values : Iterable[str]
+        The step names or safe_names to encode.
+    """
+    os.environ[ENV_CSTAR_CLOBBER_STEPS] = ",".join(values)
+
+
 ENV_CSTAR_LOG_LEVEL: t.Annotated[
     t.Literal["CSTAR_LOG_LEVEL"],
     EnvVar(
@@ -181,6 +203,16 @@ ENV_CSTAR_CLI_VERBOSE: t.Annotated[
     ),
 ] = "CSTAR_CLI_VERBOSE"
 """Set to `1` to produce verbose CLI outputs."""
+
+ENV_CSTAR_CLOBBER_STEPS: t.Annotated[
+    t.Literal["CSTAR_CLOBBER_STEPS"],
+    EnvVar(
+        "Comma-separated list of step safe_names whose prior state is cleared and re-executed on rerun.",
+        GROUP_SIM,
+        default="",
+    ),
+] = "CSTAR_CLOBBER_STEPS"
+"""Comma-separated list of step safe_names whose prior state is cleared and re-executed on rerun."""
 
 ENV_CSTAR_CLOBBER_WORKING_DIR: t.Annotated[
     t.Literal["CSTAR_CLOBBER_WORKING_DIR"],
--- a/cstar/base/env.py
+++ b/cstar/base/env.py
@@ -2,7 +2,7 @@ import os
 import sys
 import typing as t
 from collections import defaultdict
-from collections.abc import Callable, Iterable
+from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from functools import lru_cache
@@ -152,28 +152,6 @@ def generate_run_id() -> str:
     return datetime.now(UTC).strftime("%Y%m%d_%H%M%S")
 
 
-def get_clobber_steps() -> frozenset[str]:
-    """Return the set of step safe_names selected for per-step clobber.
-
-    Returns
-    -------
-    frozenset[str]
-    """
-    raw = os.environ.get(ENV_CSTAR_CLOBBER_STEPS, "")
-    return frozenset(x for entry in raw.split(",") if (x := entry.strip()))
-
-
-def set_clobber_steps(values: Iterable[str]) -> None:
-    """Set `CSTAR_CLOBBER_STEPS` to the comma-encoding of `values`.
-
-    Parameters
-    ----------
-    values : Iterable[str]
-        The step names or safe_names to encode.
-    """
-    os.environ[ENV_CSTAR_CLOBBER_STEPS] = ",".join(values)
-
-
 ENV_CSTAR_LOG_LEVEL: t.Annotated[
     t.Literal["CSTAR_LOG_LEVEL"],
     EnvVar(
@@ -203,16 +181,6 @@ ENV_CSTAR_CLI_VERBOSE: t.Annotated[
     ),
 ] = "CSTAR_CLI_VERBOSE"
 """Set to `1` to produce verbose CLI outputs."""
-
-ENV_CSTAR_CLOBBER_STEPS: t.Annotated[
-    t.Literal["CSTAR_CLOBBER_STEPS"],
-    EnvVar(
-        "Comma-separated list of step safe_names whose prior state is cleared and re-executed on rerun.",
-        GROUP_SIM,
-        default="",
-    ),
-] = "CSTAR_CLOBBER_STEPS"
-"""Comma-separated list of step safe_names whose prior state is cleared and re-executed on rerun."""
 
 ENV_CSTAR_CLOBBER_WORKING_DIR: t.Annotated[
     t.Literal["CSTAR_CLOBBER_WORKING_DIR"],
--- a/cstar/cli/workplan/run.py
+++ b/cstar/cli/workplan/run.py
@@ -9,11 +9,9 @@ from pydantic import BaseModel
 
 from cstar.base.env import (
     ENV_CSTAR_CLI_DRY_RUN,
-    ENV_CSTAR_CLOBBER_STEPS,
     ENV_CSTAR_CLOBBER_WORKING_DIR,
     ENV_CSTAR_LOG_LEVEL,
     ENV_CSTAR_RUNID,
-    set_clobber_steps,
 )
 from cstar.base.exceptions import CstarExpectationFailed
 from cstar.base.log import LogLevelChoices, get_logger
@@ -308,42 +306,24 @@ def preprocess_runid(ctx: typer.Context, run_id: str) -> str:
 
 def preprocess_clobber_steps(
     ctx: typer.Context,
-    values: list[str] | None,
-) -> list[str] | None:
+    values: list[str],
+) -> list[str]:
     """Perform validation and formatting of the `--clobber-step` option.
-
-    - clears `CSTAR_CLOBBER_STEPS` when no entries are supplied, so a value
-      preset in the shell cannot leak into an unrelated run
 
     Parameters
     ----------
     ctx : typer.Context
         A context object containing state for the typer app.
-    values : list[str] | None
+    values : list[str]
         The user-supplied step names or safe_names, one per `--clobber-step` use.
 
     Returns
     -------
-    list[str] | None
-        The original input.
-
-    Raises
-    ------
-    typer.BadParameter
-        - If an entry contains a comma, the delimiter used to encode values.
+    list[str]
+        The cleaned step names or safe_names, stripped of whitespace with any
+        empty entries dropped.
     """
-    entries = [x for value in values or [] if (x := value.strip())]
-    if not entries:
-        os.environ.pop(ENV_CSTAR_CLOBBER_STEPS, None)
-        return values
-
-    if bad_entries := [x for x in entries if "," in x]:
-        msg = f"{ARG_CLOBBER_STEP} values must not contain commas: {bad_entries!r}"
-        raise typer.BadParameter(msg, param_hint="clobber_step")
-
-    set_clobber_steps(entries)
-
-    return values
+    return [x for value in values if (x := value.strip())]
 
 
 def preprocess_path(workplan_path: str | None) -> str | None:
@@ -483,13 +463,13 @@ def run(
         ),
     ] = False,
     clobber_step: t.Annotated[
-        list[str] | None,
+        list[str],
         typer.Option(
             ARG_CLOBBER_STEP,
             help=ARG_CLOBBER_STEP_HELP,
             callback=preprocess_clobber_steps,
         ),
-    ] = None,
+    ] = [],
 ) -> None:
     """Execute a workplan.
 
@@ -510,6 +490,7 @@ def run(
                     run_id,
                     user_variables=t.cast("Mapping[str, str]", ctx.obj),
                     dry_run=dry_run,
+                    clobber_steps=clobber_step,
                 ),
             )
             console.print(get_run_summary_display(summary))
--- a/cstar/cli/workplan/run.py
+++ b/cstar/cli/workplan/run.py
@@ -35,15 +35,17 @@ from cstar.entrypoint.utils import (
     ARG_LOGLEVEL_HELP,
     ARG_LOGLEVEL_LONG,
     ARG_LOGLEVEL_SHORT,
+    OPT_CLOBBER_ALL,
 )
 from cstar.execution.file_system import local_copy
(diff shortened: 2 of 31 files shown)


cstar/entrypoint/utils.py
```diff
@@ -4,6 +4,11 @@
 ARG_CLOBBER_HELP: t.Final[str] = (
     "Set this flag to remove any pre-existing files in the working directory."
 )
+ARG_CLOBBER_STEP: t.Final[str] = "--clobber-step"
```
ankona: Should we use `--clobber` with a default "all" and convert it from bool to list[str], so we can also `--clobber a --clobber b` to specify steps? 

We haven't reached argument overload, but it seems to make good sense to let `--clobber` do this

cstar/cli/workplan/run.py
```diff
@@ -298,6 +302,46 @@ def preprocess_runid(ctx: typer.Context, run_id: str) -> str:
...
+    entries = [x for value in values or [] if (x := value.strip())]
+    if not entries:
+        os.environ.pop(ENV_CSTAR_CLOBBER_STEPS, None)
+        return values
+
+    if bad_entries := [x for x in entries if "," in x]:
+        msg = f"{ARG_CLOBBER_STEP} values must not contain commas: {bad_entries!r}"
+        raise typer.BadParameter(msg, param_hint="clobber_step")
+
+    set_clobber_steps(entries)
```
ankona: `set_env(ENV_CSTAR_CLOBBER_STEPS, ','.join(entries))` avoids an extra method...

cstar/cli/workplan/run.py
```diff
@@ -298,6 +302,46 @@ def preprocess_runid(ctx: typer.Context, run_id: str) -> str:
     return run_id
 
 
+def preprocess_clobber_steps(
+    ctx: typer.Context,
+    values: list[str] | None,
```
ankona: Can we set `default_factory=list` on the `typer.Option` and avoid unnecessary null handling (e.g. values or [])

cstar/cli/workplan/run.py
```diff
@@ -298,6 +302,46 @@ def preprocess_runid(ctx: typer.Context, run_id: str) -> str:
...
+        - If an entry contains a comma, the delimiter used to encode values.
+    """
+    entries = [x for value in values or [] if (x := value.strip())]
+    if not entries:
+        os.environ.pop(ENV_CSTAR_CLOBBER_STEPS, None)
+        return values
+
+    if bad_entries := [x for x in entries if "," in x]:
+        msg = f"{ARG_CLOBBER_STEP} values must not contain commas: {bad_entries!r}"
+        raise typer.BadParameter(msg, param_hint="clobber_step")
```
ankona: i think i've needlessly hinted before... if this is only used for one parameter, the hint isn't needed and typer should print the right thing. It only matters to avoid any name changes to the variable not being done in the string.

cstar/orchestration/dag_runner.py
```diff
@@ -518,6 +524,60 @@ async def from_run(
         )
 
 
+def _resolve_clobber_steps(steps: "list[LiveStep]") -> frozenset[str]:
```
ankona: I'd prefer this existed in the `LiveWorkplan` instead of `dag_runner`.

Could we solve this problem by adding a `clobber: bool = False` attribute to the `LiveStep` and `LiveWorkplan`? Then, use the plan-level flag to clobber all and step-level flag to clobber specific steps.

I think that would give us the ability to carry the information in the serialized workplan by populating `workflow_overrides`.

cstar/cli/workplan/run.py
ankona: this really belongs as an action in `prepare_workplan`. 

```python
def prepare_workplan(...):
  ...
  clobber_key = ENV_CSTAR_CLOBBER_WORKING_DIR
  if os.getenv(clobber_key,"") == "all":
    clobbers = {slugify(step_name): {"clobber": True} for name in wp.steps}
  else:
    clobbers = {slugify(step_name): {"clobber": True} for name in os.getenv(clobber_key, "").split(",")}

  wp.workflow_overrides = deep_merge(wp.workflow_overrides, clobbers)
```
of course, we update `LiveStep`

```python
class LiveStep:
  clobber: bool = False

  def apply_overrides(self, overrides):
    step.clobber = overrides.get("clobber", False)
    ... # any other overrides 
    
```

and have the orchestrator feed it workflow customizations 

```python

async def process_node(self, node: str) -> Task[ProcessHandle] | None:
  wp = self.planner.workplan
  step = planner.retrieve(node, KEY_STEP, None)
  ...
  step.apply_overrides(wp.workflow_overrides.get(step.safe_name, {})

```

then, passing the flag to the `cstar blueprint run` is done in the  `StepToRunRequestAdapter`

```python
class StepToRunRequestAdapter:
  def adapt(self, model: LiveStep):
    ...
    if step.clobber:
      cmd.append(ARG_CLOBBER)
      
```

ankona (changes requested):
Let's consider re-using the existing clobber variable with either "all" or a CSV of step names instead of just `FLAG_ON` instead of introducing a second variable

Also, i think there is a cleaner approach to integrating this via our unused workflow overrides.

cstar/base/env.py
```diff
@@ -152,6 +152,28 @@ def generate_run_id() -> str:
...
+    """Return the set of step safe_names selected for per-step clobber.
+
+    Returns
+    -------
+    frozenset[str]
+    """
+    raw = os.environ.get(ENV_CSTAR_CLOBBER_STEPS, "")
+    return frozenset(x for entry in raw.split(",") if (x := entry.strip()))
+
+
```
ankona: these two methods should not exist as-is in `env.py`, but should be converted into generic `get_csv` and `set_csv` that take a key

cstar/base/env.py
```diff
@@ -152,6 +152,28 @@ def generate_run_id() -> str:
...
+def set_clobber_steps(values: Iterable[str]) -> None:
+    """Set `CSTAR_CLOBBER_STEPS` to the comma-encoding of `values`.
+
+    Parameters
+    ----------
+    values : Iterable[str]
+        The step names or safe_names to encode.
+    """
+    os.environ[ENV_CSTAR_CLOBBER_STEPS] = ",".join(values)
```
ankona: I think this could do the cleaning, too...

```python
def set_csv(key: str, values: Sequence[str], *, slugify: bool = False):
  if slugify:
    cleaned = [slugify(x) for x in values]
  else:
    cleaned = [(x).strip() for x in values]

  os.environ[key] = ",".join(cleaned)
```

cstar/orchestration/launch/slurm.py
```diff
@@ -397,6 +398,10 @@ async def launch(
                     active = set(x.pid for x in dependencies).difference(successes)
                     dependencies = list(filter(lambda x: x.pid in active, dependencies))
 
+        if step.safe_name in get_clobber_steps():
+            step.fsm.clear_prior()
```
ankona: clobber flag is sent to the blueprint runner already, so we could avoid having it also do the clearing here. I think the workflow overrides would end up letting us also remove a similar thing in the `SlurmLauncher.launch`

bonus points for the clobber will be written into `script.sh` so we can verify it was intended without parsing logs

ScottEilerman:
Did another pass based on your comments. Looks like it's going through a much cleaner route now, through workflow_overrides. The intermediate env-var is gone entirely. Seems like typer won't allow `--clobber` to interpret a non-existent arg as an "all" default, so we still have `--clobber-steps <step>` on the CLI, but it passes through more cleanly, gets recorded in yamls, etc.

ankona:
> Did another pass based on your comments. Looks like it's going through a much cleaner route now, through workflow_overrides. The intermediate env-var is gone entirely. Seems like typer won't allow `--clobber` to interpret a non-existent arg as an "all" default, so we still have `--clobber-steps <step>` on the CLI, but it passes through more cleanly, gets recorded in yamls, etc.

I wasn't sure if it could be done (empty as default), so my follow-up is ... would `--clobber all` be too inconvenient that we want to create a second parameter?

ScottEilerman:
I asked our prime user and feature requester who is sitting right next to me which she prefers, and we agreed on `--clobber all` being preferable because it forces you to really ask it to clobber _everything_. And `--clobbber-steps` is _so long_ to type out. I shall make it so.

cstar/cli/workplan/run.py
```diff
@@ -302,6 +301,29 @@ def preprocess_runid(ctx: typer.Context, run_id: str) -> str:
...
+        The user-supplied step names, safe_names, or the literal token
+        `all`, one per `--clobber` use.
+
+    Returns
+    -------
+    list[str]
+        The cleaned step names or safe_names, stripped of whitespace with any
+        empty entries dropped.
+    """
+    return [x for value in values if (x := value.strip())]
```
ankona, line 329: TIL that I could get that walrus `x` in a comprehension. I've always disliked doubled `.strip()` calls but i still experience some revulsion at the syntax.

cstar/cli/workplan/run.py
```diff
@@ -430,14 +452,13 @@ def run(
...
     clobber: t.Annotated[
-        bool,
+        list[str],
         typer.Option(
             ARG_CLOBBER,
-            callback=set_flag(ENV_CSTAR_CLOBBER_WORKING_DIR),
-            help=ARG_CLOBBER_HELP,
-            envvar=ENV_CSTAR_CLOBBER_WORKING_DIR,
+            help=ARG_CLOBBER_WORKPLAN_HELP,
+            callback=preprocess_clobber_steps,
```
ankona, line 511: i think you still want `envvar=ENV_CSTAR_CLOBBER_WORKING_DIR` here. It's pretty cool that typer will show the user in the help text that it can be populated from that envvar if it's not provided

cstar/orchestration/dag_runner.py
```diff
@@ -519,6 +530,75 @@ async def from_run(
...
+    """
+    tokens = [x for token in clobber_steps or [] if (x := token.strip())]
+    if not tokens:
+        return
+
+    steps = wp.steps
+
+    if "all" in tokens:
+        if len(tokens) > 1:
+            msg = f"{ARG_CLOBBER} 'all' cannot be combined with step names"
```
ankona: good choice for fail fast.... presence of any step name sort of indicates that all was added by accident.

cstar/orchestration/dag_runner.py
```diff
@@ -519,6 +530,75 @@ async def from_run(
...
+    matched = [by_token[token] for token in tokens]
+    targeted_names = {step.name for step in matched}
+
+    for step in matched:
+        step.workflow_overrides[KEY_CLOBBER] = True
+
+    for step in steps:
+        if step.name in targeted_names:
+            continue
+        if stale := sorted(targeted_names.intersection(step.depends_on)):
```
ankona: Do you really want to output this message for each step. It feels like it may be harder to go through for the user.

If it computes steps with stale deps in one go, it can spit out:

```python
msg = "WARNING - Some dependencies on clobbered steps were found. It "
      "may result in the unintended use of outputs from steps that "
      "will not be re-run with updated inputs. Review the following "
      f"steps: {', '.join(names)}"
```

your call on 1 or N messages

Alternatives you probably considered:

- if you clobber A and it has a dependent B, automatically clobber B
- if you clobber A and it has a dependent B, report & fail, requiring `--ignore-stale-results` to execute

cstar/orchestration/models.py
```diff
@@ -305,6 +310,17 @@ def safe_name(self) -> str:
         """
         return slugify(self.name)
 
+    @property
+    def clobber(self) -> bool:
+        """Return `True` if this step's prior state should be cleared and
```
ankona, line 315:
```suggestion
        """Return `True` if all workflow steps will be re-executed."""
```
copypasta?

cstar/orchestration/dag_runner.py
```diff
@@ -519,6 +530,75 @@ async def from_run(
...
+        if len(tokens) > 1:
+            msg = f"{ARG_CLOBBER} 'all' cannot be combined with step names"
+            raise ValueError(msg)
+        for step in steps:
+            step.workflow_overrides[KEY_CLOBBER] = True
+        return
+
+    by_token = {key: step for step in steps for key in (step.name, step.safe_name)}
+
+    if unknown := set(tokens).difference(by_token):
```
ankona: Consider doing this block in the CLI for fast-failure. You can extract each check (valid names, stale dependency) to a separate method and reuse in CLI and here.

It provides immediate feedback to a user and requires no log parsing to be notified of the issue (I'm sensitive to this because I created the problem).

```python
# dag_runner.py

def check_clobber_targets(wp: Workplan, clobbers: set[str]) -> Sequence[str]:
    """Return a list of invalid clobber targets."""
    if not clobbers:
      return []
    actual = set(s.safe_name for s in wp.steps)
    return list(clobbers.difference(actual)):


def check_clobber_deps(wp: Workplan, clobbers: set[str]) -> Sequence[str]:
    """Return a list of stale dependencies on clobbers."""
    if not clobbers or set(s.safe_name for s in wp) == clobbers:
      return []
    x = ... # magic that exists
    return x
```

and the CLI

```python
# cli/workplan/run.py

def preload_workplan(ctx, path: Path):
  # cb on path arg
  ctx.obj = {'wp': deserialize(path, Workplan)}
  return path

def clean_clobbers(ctx, clobbers: list[str]):
  if clobbers:
    clobbers = [slugify(x) for x in clobbers]
  return clobbers

def handle_all_clobbers(ctx, clobbers: list[str]):
  OPT_ALL = "all"
  if clobbers and OPT_ALL in clobbers:
    clobbers = set(values)
    if len(clobbers) > 1:
      problems = clobbers - OPT_ALL
      msg = f"step names should not be specified when using clobber all"
      raise typer.BadParameter(msg)
    wp = ctx.obj['wp']
    return [s.safe_name for s in wp]
    
def validate_clobbers(ctx, clobbers: list[str]):
  if clobbers:
    wp = ctx.obj['wp']

    if names := check_clobber_targets(wp, clobbers):
      msg = f"Invalid steps specified: {','.join(names)"
      raise typer.BadParameter(msg)

    if deps := check_clobber_deps(wp, clobbers):
      msg = f"Stale items, blah blah, specified: {','.join(deps)"
      raise typer.BadParameter(msg)

  return clobbers

def run(clobbers=typer.Option(
  callback=cb_pipeline(
    clean_clobbers, 
    handle_all_clobbers, 
    validate_clobbers),
)):
```

cstar/orchestration/dag_runner.py
ankona: - This exposes CLI-level variables in a system-level module and is going to be a circular dependency issue sooner or later (`ARG_CLOBBER`). 
- At this point, you don't really know if the CLI was involved and your message is also wrong. Compute results separate from display. See suggestion below.

cstar/orchestration/dag_runner.py
```diff
@@ -519,6 +530,75 @@ async def from_run(
...
+        If `all` is combined with other tokens, or if any other token does
+        not match a step's `name` or `safe_name`.
+    """
+    tokens = [x for token in clobber_steps or [] if (x := token.strip())]
+    if not tokens:
+        return
+
+    steps = wp.steps
+
+    if "all" in tokens:
```
ankona: we're checking a CLI-specific user-input in the bowels of the system. instead, the CLI should look for "all" as an input and produce the list of all step names.

cstar/cli/workplan/run.py
ankona: per offline convo, i am wrong

ankona (changes requested):
Almost there. We have a much better UX possible by extracting your validation so it can be run in the CLI to consider.

Also, there is a very real circular dependency that requires a change.

cstar/orchestration/models.py
ScottEilerman: this is on the Step itself, so it's accurate

cstar/base/env.py
ankona: refactored out

cstar/base/env.py
ankona: refactored out

cstar/orchestration/dag_runner.py
ankona: converted to single msg

cstar/orchestration/models.py
ankona: comment targeting wrong type (Workplan)

cstar/orchestration/dag_runner.py
ankona: check relocated to `cstar.cli.workplan.run` and step names resolved before workplan processing

cstar/orchestration/dag_runner.py
ankona: - token validation relocated
- circular reference removed by shifting token validation

cstar/orchestration/dag_runner.py
ankona: created CSD-1122 and CSD-1123 as follow-ups where this wasn't implemented as noted

ankona (approved):
Updated final couple items and looks ready

PR #641, opened 2026-08-24
Update default workplan output directory

ankona:
# Summary

This change modifies the default output location for orchestrated workplans from `~/.local/share` to `~/cstar`. 

## Breaking Changes

- Change creates a variation from XDG spec for `CSTAR_DATA_HOME` 

## New Features

- N/A

## Bug Fixes

- N/A

## Improvements

- Avoid default user-facing output directory that is hidden.
- Avoid changing env var in tests where an autouse fixture already handles it

## Miscellaneous

- N/A

## Code Review Checklist

- [x] Closes #CSD-1120

### DIFF
--- a/cstar/base/env.py
+++ b/cstar/base/env.py
@@ -273,7 +273,7 @@ ENV_CSTAR_DATA_HOME: t.Annotated[
     EnvVar(
         "Environment variable used to override the home directory for C-Star dataset storage.",
         GROUP_FS,
-        "~/.local/share",
+        "~/cstar",
         indirect_var="XDG_DATA_HOME",
         default_factory=lambda x: hpc_data_directory() or indirect_default_factory(x),
     ),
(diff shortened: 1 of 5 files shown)


cstar/base/env.py
```diff
@@ -273,7 +273,7 @@ def generate_run_id() -> str:
     EnvVar(
         "Environment variable used to override the home directory for C-Star dataset storage.",
         GROUP_FS,
-        "~/.local/share",
+        "~/cstar",
         indirect_var="XDG_DATA_HOME",
```
ScottEilerman, line 277: I'm guessing this means if the user _does_ explicitly specify XDG_DATA_HOME, we still respect it, and just fall back to `"~/cstar"` if they don't (which will be the case for 99.7% of our users)?

ScottEilerman (approved):
one question but LGTM regardless of the answer!

cstar/base/env.py
ankona: yes, exactly - it will honor user settings if configured.

PR #643, opened 2026-08-25
Schema update for namelist_overrides section, partitioning rework; migration promotion

ScottEilerman:
# Summary

Blueprint schema 3.0.0 for `roms_marbl`: users can now override any `namelist.nml` value from the blueprint via a new `namelist_overrides` mapping, `model_params` is dissolved (the time step now lives in the namelist; `use_pio` moves under `partitioning`), and partitioning gains in-development auto-tiling fields. Alongside the schema bump, blueprint schema migration is promoted from feature-flagged to default-on: `cstar blueprint run` transparently migrates out-of-date blueprints, and the `migrate` command is always available.

## Breaking Changes
- `roms_marbl` blueprint schema is now **3.0.0** (single migration step from 2.1.0; `cstar blueprint run` migrates automatically, or run `cstar blueprint migrate`):
  - `model_params` is removed. `time_step` is now the namelist's `time_stepping.dt` (migration seeds it into `namelist_overrides`); `use_pio` moves to `partitioning.use_pio`.
- `time_step` is removed from `Discretization`/`ROMSDiscretization`; previously serialized `ROMSSimulation.to_dict()` payloads carrying `discretization.time_step` no longer load.
- `roms_runtime_settings` derives `ntimes` from the effective (post-override) `dt` and now also writes `param_settings.np_xi`/`np_eta` from the partitioning, so the runtime namelist always matches the scheduled cores.

## New Features
- `namelist_overrides` blueprint field: a `group -> {key: value}` mapping deep-merged onto the runtime namelist **after** C-Star's derived settings, so user values win. Partial nested overrides touch only the keys they name; list values replace the existing list wholesale; unknown groups/keys fail loudly against the versioned namelist schema; overriding `np_xi`/`np_eta` inconsistently with `partitioning` is an error.
- Early (blueprint-validation-time) checking of `namelist_overrides` names: unknown groups/keys are an error when `code.roms` pins a release version, and a warning against the best-guess schema for unpinned refs; `np_xi`/`np_eta` conflicts with `partitioning` error at validation time regardless. Values are still validated at runtime against the merged namelist.
- In-development auto-tiling support on `partitioning`: `auto_tiling` (requires `use_pio`) and `n_cores` (usable in place of `n_procs_x`/`n_procs_y`; consistency enforced when all three are given). Build-time check requires `#define MPI_MASKING` when `auto_tiling` is enabled.
- Automatic blueprint schema migration during `cstar blueprint run` is now the default, and `cstar blueprint migrate` is always mounted (feature flags `CSTAR_FF_CLI_BP_MIGRATE_AUTO`/`CSTAR_FF_CLI_BP_MIGRATE_SHOW` removed).
- New `CSTAR_DISABLE_MIGRATION=1` escape hatch: guarantees a blueprint is never modified — commands fail early if the blueprint is not at the current schema version.

## Bug Fixes
- `RuntimeParameterSet`'s model validator silently disabled the inherited `ParameterSet` locked/hash check (Pydantic v2 same-name validator shadowing): a `locked: true` parameter set with no hash validated cleanly.
- `cstar blueprint run` exited with code 0 when the blueprint failed validation; it now exits 1.
- Serialized blueprints/workplans omitted `schema_version` whenever it equaled the current default (`exclude_defaults`); it is now always written so persisted artifacts self-describe.

## Improvements
- `deep_merge` gains a `replace_lists` option (used by the namelist override path) so a shorter override list is not element-wise merged with stale values.
- The `use_pio`/`PARALLEL_IO` cppdefs check is generalized into `_validate_cppdef_flag(define, enabled, ...)`, shared by the new `MPI_MASKING` check.
- Auto-migration of an already-current blueprint is a no-op: nothing is persisted to the state directory and the original file is used as-is.

## Miscellaneous
- New versioned template `blueprint.3.0.0.yaml` and generated `roms_marbl_schema.3.0.0.json`; `docs/schemas/index.rst`, `docs/blueprints.rst`, and the tutorial blueprints/notebook updated to the 3.0.0 shape; stale `ModelParameterSet` API listing removed.
- Test fixtures (`blueprint_complete.yaml`, `blueprint_template.yaml`, unit-test conftest data) migrated to 3.0.0.

## Notes for Reviewers
- The field name `auto_tiling` was workshopped from `optimized_tiling`; the literal is confined to the blueprint field, the simulation attribute, and error messages, so a further rename is one grep.
- Two `cstar blueprint check` tests are `xfail(strict=False)`: they fetch the wales-toy blueprint live from `cstar_blueprint_roms_marbl_example`, which is still schema 2.0.0 (`check` does not auto-migrate). They self-clear once that repo publishes a 3.0.0 blueprint. The corresponding `run` test passes via auto-migration.
- `CSTAR_DISABLE_MIGRATION` is enforced at the single choke point in `execute_migration`, so it also blocks explicit `cstar blueprint migrate` — "disabled" means disabled everywhere.
- The migration discards `model_params`' `locked`/`hash` metadata (noted in the adapter docstring); `namelist_overrides` is a plain mapping and is not lockable.
- Follow-up (separate PR, sibling repo): cstar-forge still emits `model_params` blueprints and re-validates against the installed C-Star models at emit time, so Forge environments should pin their C-Star ref until the Forge-side change lands. Already-emitted 2.x blueprints run fine against this branch via auto-migration.

## Code Review Checklist
- [ ] Closes #xxxx

### DIFF
--- a/cstar/applications/roms_marbl/models.py
+++ b/cstar/applications/roms_marbl/models.py
@@ -1,4 +1,5 @@
 import typing as t
+import warnings
 from datetime import datetime
 
 from pydantic import (
@@ -292,11 +293,14 @@ class RomsMarblBlueprint(Blueprint):
 
     These overrides are layered onto the runtime namelist LAST, after C-Star's
     own derived settings, so they take precedence over anything C-Star would
-    otherwise compute; list values replace the existing list wholesale. Keys
-    are validated at runtime against the namelist
-    schema for the pinned ucla-roms version. Overriding `param_settings.np_xi`/
-    `param_settings.np_eta` to values that conflict with `partitioning` is an
-    error.
+    otherwise compute; list values replace the existing list wholesale.
+
+    Group and key names are checked when the blueprint is validated: unknown
+    names are an error when `code.roms` pins a release version, and a warning
+    (against the best-guess schema) otherwise. Values are validated at runtime
+    against the namelist schema for the pinned ucla-roms version. Overriding
+    `param_settings.np_xi`/`param_settings.np_eta` to values that conflict
+    with `partitioning` is an error.
     """
 
     @model_validator(mode="after")
@@ -318,7 +322,62 @@ class RomsMarblBlueprint(Blueprint):
             msg = "code.pio was supplied but partitioning.use_pio is false"
             raise ValueError(msg)
 
+        self._validate_namelist_overrides()
+
         return self
+
+    def _validate_namelist_overrides(self) -> None:
+        """Check `namelist_overrides` names and partitioning consistency early.
+
+        Unknown group/key names are an error when `code.roms` pins a release
+        version (the namelist schema is then certain) and a warning otherwise,
+        where the latest schema is only a best guess. `np_xi`/`np_eta`
+        conflicts with `partitioning` are schema-independent and always an
+        error. Values are not checked here — the merge onto the real namelist
+        in `roms_runtime_settings` re-validates against the full schema.
+        """
+        if not self.namelist_overrides:
+            return
+
+        # deferred import: the cstar.roms package init pulls in ROMSSimulation,
+        # which imports this module back (circular at module import time)
+        from cstar.roms.namelist import _parse_semver, namelist_schema_for_ref
+
+        param_overrides = self.namelist_overrides.get("param_settings", {})
+        violations = [
+            f"param_settings.{key}={value!r} conflicts with partitioning value "
+            f"{expected!r}"
+            for key, expected in (
+                ("np_xi", self.partitioning.n_procs_x),
+                ("np_eta", self.partitioning.n_procs_y),
+            )
+            if (value := param_overrides.get(key)) is not None
+            and expected is not None
+            and value != expected
+        ]
+
+        ref = self.code.roms.checkout_target
+        pinned_release = _parse_semver(ref) is not None
+        if pinned_release:
+            schema = namelist_schema_for_ref(ref)
+            violations += schema.unknown_override_keys(self.namelist_overrides)
+        else:
+            with warnings.catch_warnings():
+                # suppress the unparseable-ref fallback warning; unknown names
+                # are reported below only if any are found
+                warnings.simplefilter("ignore", UserWarning)
+                schema = namelist_schema_for_ref(ref)
+            if unknown := schema.unknown_override_keys(self.namelist_overrides):
+                warnings.warn(
+                    f"namelist_overrides contain names not in {schema.__name__} "
+                    f"(assumed for unpinned ucla-roms ref {ref!r}): "
+                    + "; ".join(unknown),
+                    UserWarning,
+                    stacklevel=2,
+                )
+
+        if violations:
+            raise ValueError("invalid namelist_overrides: " + "; ".join(violations))
 
     @property
     def cpus_needed(self) -> int:
--- a/cstar/applications/roms_marbl/models.py
+++ b/cstar/applications/roms_marbl/models.py
@@ -19,6 +19,7 @@ from cstar.orchestration.models import (
     DocLocMixin,
     PathFilter,
 )
+from cstar.roms.namelist import _parse_semver, namelist_schema_for_ref
 
 APP_NAME: t.Literal["roms_marbl"] = "roms_marbl"
 
@@ -184,7 +185,12 @@ class RuntimeParameterSet(ParameterSet):
 
 
 class PartitioningParameterSet(ParameterSet):
-    """Parameters for the partitioning of the model."""
+    """Parameters for the partitioning of the model.
+
+    Namelist overrides of the processor grid (`np_xi`/`np_eta` in
+    `param_settings`) must agree with these parameters; the partitioning is
+    authoritative for the processor grid at runtime.
+    """
 
     n_procs_x: PositiveInt | None = None
     """Number of processes used to subdivide the domain on the x-axis."""
@@ -204,6 +210,13 @@ class PartitioningParameterSet(ParameterSet):
     n_cores: PositiveInt | None = None
     """Total number of cores to use when `auto_tiling` selects the tiling layout."""
 
+    @property
+    def dims(self) -> tuple[PositiveInt, PositiveInt] | None:
+        """The `(n_procs_x, n_procs_y)` pair, or `None` unless both are set."""
+        if self.n_procs_x is None or self.n_procs_y is None:
+            return None
+        return (self.n_procs_x, self.n_procs_y)
+
     @model_validator(mode="after")
     def _validate_partitioning(self) -> "PartitioningParameterSet":
         """Perform validation on the model after field-level validation is complete.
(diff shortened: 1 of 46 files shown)


cstar/applications/roms_marbl/models.py
```diff
@@ -181,30 +186,73 @@ def _model_validator(self) -> "RuntimeParameterSet":
 class PartitioningParameterSet(ParameterSet):
     """Parameters for the partitioning of the model."""
 
-    n_procs_x: PositiveInt
+    n_procs_x: PositiveInt | None = None
```
ankona, line 195: I'd suggest swapping `n_procs_x` and `n_procs_y` into a 2-tuple that is instead nullable. That would make it easier to say you have both x and y and perform any other checks with a single comparison to `partitioning.dims` instead of `if p.npx is None or p.npx is None`

cstar/applications/roms_marbl/models.py
```diff
@@ -181,30 +186,73 @@ def _model_validator(self) -> "RuntimeParameterSet":
...
+    @model_validator(mode="after")
+    def _validate_partitioning(self) -> "PartitioningParameterSet":
+        """Perform validation on the model after field-level validation is complete.
+
+        Named distinctly from `ParameterSet._model_validator` so the inherited
+        locked/hash check is not shadowed and both validators run.
+        """
+        violations: list[str] = []
+
+        if self.auto_tiling and not self.use_pio:
```
ankona: can we combine all the auto-tiling tests into a single conditional? It feels infinitely more readable as:

```python

has_xy = self.n_procs_x and self.n_procs_y

if self.auto_tiling:
  if not self.use_pio:
    violations.append("tiling requires PIO")

  if not self.n_cores and (self.n_cores is None or not has_xy):
    violations.append("tiling requires ncores or dims")
else:
  if not has_xy:
    violations.append("dims are required unless using tiling")
  
  if self.n_cores:
    violations.append("cores can only be specified with tiling")

```

cstar/applications/roms_marbl/models.py
```diff
@@ -243,6 +288,21 @@ class RomsMarblBlueprint(Blueprint):
...
+    These overrides are layered onto the runtime namelist LAST, after C-Star's
+    own derived settings, so they take precedence over anything C-Star would
+    otherwise compute; list values replace the existing list wholesale.
+
+    Group and key names are checked when the blueprint is validated: unknown
+    names are an error when `code.roms` pins a release version, and a warning
+    (against the best-guess schema) otherwise. Values are validated at runtime
+    against the namelist schema for the pinned ucla-roms version. Overriding
+    `param_settings.np_xi`/`param_settings.np_eta` to values that conflict
+    with `partitioning` is an error.
```
ankona: Consider dumping the final sentence and replacing with `see: class:PartitioningParameterSet documentation for more information on potential conflicts`

cstar/applications/roms_marbl/models.py
```diff
@@ -258,13 +318,76 @@ def _model_validator(self) -> "RomsMarblBlueprint":
             msg = "start_date is outside the valid range"
             raise ValueError(msg)
 
-        if self.code.pio is not None and not self.model_params.use_pio:
-            msg = "code.pio was supplied but model_params.use_pio is false"
+        if self.code.pio is not None and not self.partitioning.use_pio:
+            msg = "code.pio was supplied but partitioning.use_pio is false"
```
ankona: Can we avoid dot notation in log messages? Consider `PIO is not enabled in the partitioning configuration for this blueprint"

cstar/applications/roms_marbl/models.py
```diff
@@ -258,13 +318,76 @@ def _model_validator(self) -> "RomsMarblBlueprint":
...
+        conflicts with `partitioning` are schema-independent and always an
+        error. Values are not checked here — the merge onto the real namelist
+        in `roms_runtime_settings` re-validates against the full schema.
+        """
+        if not self.namelist_overrides:
+            return
+
+        # deferred import: the cstar.roms package init pulls in ROMSSimulation,
+        # which imports this module back (circular at module import time)
+        from cstar.roms.namelist import _parse_semver, namelist_schema_for_ref
```
ankona: the namelist shouldn't need `ROMSSimulation`. Can you remove `__all__` from `cstar.roms.__init__.py` and avoid this unnecessary circular dependency? 

Some file relocation will solve this without having a timebomb.

cstar/applications/roms_marbl/models.py
```diff
@@ -258,13 +318,76 @@ def _model_validator(self) -> "RomsMarblBlueprint":
...
+        self._validate_namelist_overrides()
+
         return self
 
+    def _validate_namelist_overrides(self) -> None:
+        """Check `namelist_overrides` names and partitioning consistency early.
+
+        Unknown group/key names are an error when `code.roms` pins a release
+        version (the namelist schema is then certain) and a warning otherwise,
+        where the latest schema is only a best guess. `np_xi`/`np_eta`
```
ankona, line 343: we shouldn't have to perform duplicate validation here (re: np_xi/np_eta because it's obvious, but possibly other dupes to review)

perhaps you can perform `_apply_namelist()` in your `after` validator on the first line. Then, it would validate a single time with the rules in one place (e.g. the partitioning parameter set) instead of duplicating validation for every attribute here).

cstar/cli/common.py
```diff
@@ -381,6 +385,21 @@ def execute_migration(request: MigrationRequest) -> PersistedMigrateResult:
...
+    if not migration_result.plan.adapters and request.target is None:
+        return PersistedMigrateResult(migration_result, request.source)
+
+    if migration_result.plan.adapters and is_flag_enabled(ENV_CSTAR_DISABLE_MIGRATION):
+        print(
+            f"Blueprint at '{request.source}' requires schema migration from "
+            f"{migration_result.plan.source!r} to {migration_result.plan.target!r}, "
+            f"but migration is disabled ({ENV_CSTAR_DISABLE_MIGRATION}=1). Unset "
+            f"{ENV_CSTAR_DISABLE_MIGRATION} to allow migration, or update the "
+            "blueprint to the current schema."
```
ankona, line 402: use `console.print` to get some nice string colorization features from typer

cstar/tests/unit_tests/orchestration/cli/blueprint/test_run_blueprint.py
```diff
@@ -63,6 +69,9 @@ def test_blueprint_run_remote_blueprint_dne() -> None:
 def test_blueprint_run_remote_blueprint() -> None:
     """Verify that a URL to a remote blueprint is handled properly and the
     blueprint is executed.
+
+    The published wales_toy blueprint is still schema 2.0.0; automatic
```
ankona, line 73: do we want to xfail this the same as the other test that calls out the "partner repo publishing an updated version"?

cstar/roms/namelist.py
```diff
@@ -701,6 +701,12 @@ class RomsNamelistBase(BaseModel):
         extra="forbid", validate_assignment=True, use_attribute_docstrings=True
     )
 
+    _NOT_A_USABLE_SCHEMA_MSG: ClassVar[str] = (
```
ankona, line 704: Consider using `ABC` to avoid needing these checks and detached messaging...

There must be something that a namelist can share with subclasses, like a list of valid keys that can be used for overrides...

```python
class NamelistBase(abc.ABC, BaseModel):
    @abc.abstractmethod
    def _valid_keys(self) -> set[str]:
        ...
```

cstar/roms/namelist.py
```diff
@@ -762,6 +764,43 @@ def _reject_direct_use(cls, data: Any) -> Any:
     pipe_frc_settings: PipeFrcSettings
     particles_settings: _ParticlesSettingsCommon
 
+    @classmethod
+    def unknown_override_keys(cls, overrides: dict[str, dict[str, Any]]) -> list[str]:
```
ankona, line 768: I feel like we could get this behavior for free if namelist subclasses had `extras="forbid"` and we did `overridden = NamelistXXX(**og_model.model_dump(), **overrides.model_dump())`

ankona: unless there is an adaptation (flat attrs on namelist to deep attrs on obj)... in that case, i'd like to see us create a `ModelAdapter[tuple[XxxNamelist,Xxx]], Xxx]` instead of building the adaptation functionality into the namelist where it might cause problems with subclasses

ankona (changes requested):

cstar/cli/common.py
ankona: rich probably handles creating a single `Console` but there is an instance in `common.py` already to use

ankona (approved):

PR #646, opened 2026-08-27
Implement deferred blueprints

ScottEilerman:
# Summary

A workplan step can now consume a blueprint *generated by an earlier step*, instead of requiring every blueprint to exist before the workplan is scheduled. Point the step's `blueprint` field at the producer (`{from_step: <name>, filename: <optional>}`) and C-Star finds and runs the generated blueprint at runtime, with `blueprint_overrides` still supported. This enables workplans like "generate model inputs, then run the simulation they describe" in a single DAG.

Alongside the feature, this PR reworks how *all* steps' overrides are handled: rather than baking overrides into rewritten blueprint files at schedule time, every step keeps its original blueprint path in the transformed workplan and carries its overrides in a runtime `apply-overrides` directive. Overrides are still validated at schedule time (bad overrides fail `prepare_workplan`, not the compute node), and each step's CPU requirement is recorded into its `compute_overrides` so schedulers never need to read blueprint content.

Validated end to end with cstar-forge as the producer (forge generates ROMS-MARBL inputs and the blueprint; the next step resolves and runs it to completion). 

## Breaking Changes
- Transformed workplans no longer reference rewritten `.ovrd.yaml` blueprint files; they keep the original blueprint paths and record overrides as `apply-overrides` directives. The merged blueprint is written at runtime into each step's own run directory. Re-preparing a run is now idempotent (no more compounded `.ovrd.ovrd.yaml` files).

## New Features
- Deferred blueprint references in workplan steps: `blueprint: {from_step: <name>, filename: <optional>}`. The producer must be listed in `depends_on`; when `filename` is omitted, exactly one blueprint file must be present in the producer's output directory.
- `cstar blueprint run` accepts `step://<step>[/<filename>]` URIs, resolving them against the current run at runtime (including automatic schema migration of the resolved blueprint, when enabled).
- Each step's CPU requirement is recorded in the transformed workplan (`compute_overrides.slurm.num_cpus`, read from the blueprint when available); a declared value always wins, and a deferred step with no declaration is scheduled at 1 CPU.
- Unsupported combinations fail loudly at schedule time (e.g. deferring a step whose application uses schedule-time transforms, such as ROMS-MARBL time splitting).

## Bug Fixes
- Workplans using external applications (e.g. cstar-forge) crashed during preparation when the application's blueprint model used its own enums.
- Values set via `blueprint_overrides` (e.g. `use_pio`) were invisible to downstream steps inspecting a producer's configuration, sending restart-file searches to the wrong directory.
- The local launcher logged an adaptation-failure traceback for steps whose compute overrides carried only scheduler (slurm) keys.
- A SLURM submission whose declared `compute_overrides` fail validation now aborts with an error instead of submitting with default resources and a warning.

## Miscellaneous
- Workplan documentation gains a Deferred Blueprints example; the workplan JSON schema accepts the new `blueprint` mapping form.

## Code Review Checklist
- [ ] Closes #xxxx

## Notes for Reviewers
- Design shape: one override-persistence path for all steps; `LiveStep.blueprint` returns `Blueprint` and raises a typed `BlueprintDeferredError` for deferred steps (no `Blueprint | None`); `DeferredBlueprintRef` is referenced only at its definition/validator, the accessor, the transformer's single guard region, the runtime resolver, and the CLI URI boundary. A new `effective_blueprint(step)` merges a step's packaged directive payload in memory for consumers that need post-override content (ContinuanceDirective).
- Deliberate scope cut: the feature-flagged time-splitter pipeline (`ENV_FF_ORCH_TRX_TIMESPLIT`, off by default) still materializes merged blueprints at schedule time; unifying it onto runtime directives is a larger rework of a disabled feature.
- Pre-existing quirk noticed, not changed: `OverrideTransform.apply` replaces `description` with its audit trail, so a user override of `description` is discarded.
- No automated two-step integration test yet: no in-repo application generates a blueprint into its output directory, so the end-to-end path was validated manually with cstar-forge. The CLI resolution path has unit coverage with a real hello_world execution.
- When `filename` is omitted, resolution globs the producer's output directory; re-running the same run-id deliberately reuses a completed producer's published blueprint (resume semantics), and fresh run-ids get fresh directories.


cstar/orchestration/models.py
```diff
@@ -246,6 +247,87 @@ class WorkplanState(StrEnum):
...
+        Parameters
+        ----------
+        uri : str
+            The URI to inspect.
+
+        Returns
+        -------
+        bool
+        """
+        return uri.startswith(cls.SCHEME)
```
ankona: may fail due to case-sensitivity. should we add `uri.strip().casefold()`?

cstar/orchestration/models.py
```diff
@@ -452,6 +547,37 @@ def _check_dependencies(cls, value: list[Step]) -> list[Step]:
...
+    def _check_deferred_blueprints(cls, value: list[Step]) -> list[Step]:
+        """Verify deferred blueprint references point at valid producer steps.
+
+        Parameters
+        ----------
+        value : list[Step]
+            The steps in the workplan.
+        """
+        names = {step.name for step in value}
+        for step in value:
```
ankona: can we use 

`deferrals = (s for s in value if s.is_deferred)`

to avoid this reference to DeferredBlueprintRef?

cstar/orchestration/models.py
```diff
@@ -452,6 +547,37 @@ def _check_dependencies(cls, value: list[Step]) -> list[Step]:
 
         return value
 
+    @field_validator("steps", mode="after")
+    @classmethod
+    def _check_deferred_blueprints(cls, value: list[Step]) -> list[Step]:
```
ankona: i think i did the wrong thing above, but do we want to swap to `Sequence[Step]` to be sure we don't try to modify the value

cstar/orchestration/transforms.py
```diff
@@ -514,6 +534,21 @@ def apply(self) -> Workplan:
                 trx_step.depends_on.clear()
                 trx_step.depends_on.extend(depends_on)
 
+        # deferred blueprint references must point at steps that still exist;
```
ankona: It doesn't make a ton of sense to run a time-splitter on a step that produces a blueprint - they seem completely orthogonal. 

I think there is a solution that can get rid of all these `if isinstance(x, DeferredBlueprint)` everywhere... 

- maybe a `class DeferredStep` - no more `s.blueprint_path is None` 
- would have to couple with transforms that apply to deferred steps

cstar/orchestration/transforms.py
```diff
@@ -658,6 +700,36 @@ def apply_automatic_overrides(step: LiveStep) -> LiveStep:
     return overridden_step_result[0]
 
 
+def package_deferred_overrides(step: LiveStep) -> LiveStep:
```
ankona: this would be a `DeferredOverrideTransform`

cstar/orchestration/transforms.py
```diff
@@ -473,8 +474,13 @@ def apply(self) -> Workplan:
             live_steps = [filled for step in live_steps for filled in fill(step)]
 
         # apply user blueprint_overrides and ensure consistent output targets;
-        # must happen before time-splitting so the splitter reads correct blueprint values
-        steps = [apply_automatic_overrides(step) for step in live_steps]
+        # must happen before time-splitting so the splitter reads correct blueprint values;
+        # deferred blueprints do not exist yet, so their overrides are packaged
+        # into a runtime directive instead
+        steps = [
+            step if step.is_deferred else apply_automatic_overrides(step)
```
ankona: could `apply_automatic_overrides` perform `package_deferred_overrides` (since it's technically _the_ automatic override for a deferred step?

cstar/orchestration/launch/local.py
```diff
@@ -257,11 +257,12 @@ def adapt_step(
...
 
-        if step.compute_overrides:
+        # overrides for other launchers (e.g. the slurm cpu requirement the
+        # workplan transformer records) are not this launcher's concern
+        if step.compute_overrides.get("local"):
             try:
-                if step.compute_overrides:
-                    compute = LocalComputeAdapter().adapt(step.compute_overrides)
-                    enricher = TimeConstrainedRunRequestEnricher(compute)
+                compute = LocalComputeAdapter().adapt(step.compute_overrides)
```
ankona, line 266: ty for fixing my copypasta fail (I wanted to be doubly sure it had overrides 🦞 )

cstar/orchestration/launch/slurm.py
```diff
@@ -182,12 +182,26 @@ def configured_account() -> str:
...
+            The step being scheduled. Reserved for parity with the rest of
+            the compute-spec API; not consulted for `num_cpus`.
+
         Returns
         -------
         SlurmComputeSpec
         """
         return SlurmComputeSpec(
-            num_cpus=step.blueprint.cpus_needed,
+            num_cpus=1,
```
ankona: There is a downstream call (`SchedulerJob.__init__`) that has optional parameters triggering the use of some default for everything except `cpus`. That call drove me to set `num_cpus` from the blueprint on the `default_compute` instance of `SlurmComputeSpec`, even though i did not want to touch the blueprint here.

I like being able to have a real default, but also think it highlights the issue of finding default compute specs two ways. We should create a follow-up task to pull those defaults out of the methods mentioned above so the `SlurmComputeSpec` and `LocalComputeSpec` are the source-of-truth for these values. 

For now:
- if `1` is the true default for `num_cpus`, let's tweak `SlurmComputeSpec` with `num_cpus: int = Field(default=1)` (remember, not using kwarg form for pydantic defaults results in poor IDE behavior when generating intellisense for constructors)

For follow-up ticket:
- `get_default_compute_spec` can be converted into a `@classmethod` exposed directly from `XxxComputeSpec` classes (side note: `SlurmLauncher.configured_xxx` will likely be dropped and implemented here to avoid creating a circular reference and satisfy DRY principle).
- `SchedulerJob.__init__` can be tweaked to take a `SlurmComputeSpec` in place of the independent parameters to reduce param count (yes, i know we have `PBSJob` that also runs through here but we really need to just consider dropping PBS support or creating a `PBSComputeSpec`

cstar/orchestration/launch/local.py
```diff
@@ -257,11 +257,12 @@ def adapt_step(
 
         enricher: ModelEnricher[RunRequest] | None = None
 
-        if step.compute_overrides:
+        # overrides for other launchers (e.g. the slurm cpu requirement the
+        # workplan transformer records) are not this launcher's concern
```
ankona: I don't disagree with this comment in spirit, but the purpose of the adapter is to hide how compute overrides are converted into the `XxxComputeSpec`, including which keys to look for. Unless there is another reason besides the empty check, I'd revert to the prior conditional `if step.compute_overrides` in case we need to adjust the way the generic `KeyValueStore` on the `Step` is populated.

cstar/orchestration/launch/slurm.py
```diff
@@ -216,9 +230,11 @@ def _get_compute_spec(step: "LiveStep") -> SlurmComputeSpec:
                 compute = default_compute.model_copy(
                     update=overrides.model_dump(exclude_defaults=True)
                 )
-            except CstarAdaptationError:
+            except CstarAdaptationError as ex:
+                # the default spec is only a 1-cpu floor; submitting with it
+                # in place of the declared resources must fail, not warn
                 msg = f"SLURM overrides did not result in valid compute spec: {step.compute_overrides}"
-                log.warning(msg, exc_info=True)
+                raise CstarError(msg) from ex
```
ankona, line 241: great catch. failure to configure is way better as fail-fast

cstar/orchestration/models.py
```diff
@@ -246,6 +247,87 @@ class WorkplanState(StrEnum):
     """A workflow that has been validated."""
 
 
+class DeferredBlueprintRef(BaseModel):
```
ankona, line 250: i like to add `str_strip_whitespace` and `use_attribute_docstring` as minimum model config if we can't inherit to:
- reduce annoying need to do manual `.strip()` on any attrs
- improve field documentation in generated/published schemas

We probably should also try to be strict on fields, like other models...

```
    model_config: t.ClassVar[ConfigDict] = ConfigDict(
        extra="forbid",
        str_strip_whitespace=True,
        use_attribute_docstrings=True,
    )
```

cstar/orchestration/models.py
```diff
@@ -246,6 +247,87 @@ class WorkplanState(StrEnum):
...
+    located in the producing step's output directory at runtime.
+    """
+
+    SCHEME: t.ClassVar[str] = "step://"
+    """URI scheme used as the command-line wire form of a deferred reference."""
+
+    from_step: RequiredString
+    """The name of the step that generates the blueprint."""
+
+    filename: RequiredString | None = None
```
ankona: Should this just use an empty-string default, e.g. ` = Field(default="")` instead of `None`? We get the same truthiness checks without the hassle of sometimes  checking None.

cstar/orchestration/transforms.py
```diff
@@ -514,6 +538,20 @@ def apply(self) -> Workplan:
                 trx_step.depends_on.clear()
                 trx_step.depends_on.extend(depends_on)
 
+        # deferred blueprint references must point at steps that still exist;
```
ankona, line 541: Is this true? 

A deferred step is the same as any other step except for when the blueprint is available.

If I run the time-splitter, I don't see anything unreasonable about wanting to chain a deferred step to a split step. In this example, the `forge` app generates a blueprint that we want to use for both spin-up and the main simulation.

```yaml
- name: Generate Blueprint
  application: forge
- name: Spin-up
  application: roms_marbl
  blueprint: step://generate-spin-up
  blueprint_overrides:
    start_date: <sd>-90 days
    end_date: <sd>
- name: Main Simulation
  application: roms_marbl
  blueprint: step://blueprint-generation
  depends_on: ["Generate Blueprint", "Spin-up"]
  continue-from:
    step: Spin-up
  blueprint_overrides:
    start_date: <sd>
    end_date: <ed>
```

If spin-up is long enough to warrant monthly time-splitting, the `Main Simulation` deferred step needs to be transformed properly (ignoring that i'm not time-splitting main sim just for ease of description):

```yaml
...
- name: Spin-up 1
  ...
- name: Spin-up 2
  ...
- name: Spin-up 3
  ...
- name: Main Simulation
  application: roms_marbl
  blueprint: step://blueprint-generation
  depends_on: ["Generate Blueprint", "Spin-up 3"] # <--- mapped dep
  continue-from:
    step: spin-up-3  # <----------- mapped step name in directive
  blueprint_overrides:
    start_date: <sd>
    end_date: <ed>
```

We can add this as a follow-up task, but the time-splitter definitely needs to be updating any "step references" and the initial flaw here is that it only updates `depends_on`.

cstar/orchestration/transforms.py
```diff
@@ -756,23 +1056,23 @@ def apply_directives(
...
             step = LiveStep(
                 name="directive-step",
-                application=Application.ROMS_MARBL,
+                application=app.name,
                 blueprint=local_bp,
                 working_dir=blueprint.working_dir,
             )
 
-            transforms = {
+            transforms = [
```
ankona, line 1069: good catch. i used a set to de-dupe and lost prioritiztion.

cstar/orchestration/transforms.py
```diff
@@ -818,3 +1118,69 @@ def load_workplan(
     @classmethod
     def register(cls, key: str, directive: type[Directive]) -> None:
         DirectiveConfig.directive_map[key] = directive
+
+
+def resolve_deferred_blueprint(ref: DeferredBlueprintRef) -> Path:
```
ankona, line 1123: If we convert this method into `class DeferedBlueprintResolutionDirective` (what a mouthful), can we get automatic resolution without manually calling the method using the pre-existing directive processing?

We could do this during the normal `WorkplanTransformer` processing.

```python
class DeferredSourceTransform:
   def apply(step: LiveStep):
      if step.is_deferred:
        step.directives = {"bp-resolve": {}, **step.directives}

# - or - 

class WorkplanTransformer:
  ...
  def apply(...):
    for step in steps:
      if step.is_deferred:
        workplan_overrides["directives"]["bp-resolve"] = {}
```

ankona (approved):
LGTM with those follow-ups

PR #647, opened 2026-08-27
Local launcher fixes and promotion

ScottEilerman:
# Summary

Local workplan runs could hang forever and lost track of step status once `cstar workplan run` returned. This PR makes several fixes and promotes proxy features that were previously behind a feature flag: steps now run in sequence, failures stop the steps that depend on them, and `cstar workplan status` reports accurate progress at any time — during the run and after it finishes.

Verified with real two-step local workplan runs (forge → ROMS-MARBL); unit suite passes (1461 passed) and `pre-commit run --all-files` is clean.

## Breaking Changes
- `cstar workplan run` on a local machine now schedules all steps and returns; use `cstar workplan status` to follow progress.
- A step whose dependency fails is now aborted and marked Failed instead of running against incomplete upstream output.
- The experimental `CSTAR_FF_ENABLE_LOCAL_PROXY` flag is removed; its behavior is now the default.

## Bug Fixes
- Local workplan runs hung indefinitely because a step that finished was still reported as running.
- `cstar workplan status` showed local steps as Running forever after the launching process exited.

## Code Review Checklist
- [ ] Closes #xxxx

## Notes for Reviewers
- Root causes: step status was read from `Popen.returncode` without ever calling `poll()` (so self-exited processes stayed "Running"), and a fresh status process had no way to observe outcomes at all. Each step now runs under the proxy script, which waits on its dependencies, checks their recorded outcome, and writes status transitions to per-run sentinel files that `load_run_state` reads back.
- An adversarial review caught that removing the in-parent dependency wait initially dropped failure propagation; the dependency-outcome check in the proxy script is the fix, tested by executing the generated script for done/failed/vanished dependencies. The unit suite now exercises the proxy path throughout.
- `test_blueprint_migrate_file_dne` fails on `main` as well; unrelated to this PR.
- Pairs with the deferred-blueprints PR: that feature schedules fine without this one, but real local multi-step runs need these fixes to track the producer step's completion. No coupling on SLURM.

### DIFF
--- a/cstar/additional_files/templates/launchers/local_job_proxy.sh
+++ b/cstar/additional_files/templates/launchers/local_job_proxy.sh
@@ -2,6 +2,7 @@
 SENTINEL_PATH="{sentinel_path}"
 BLUEPRINT_PATH="{blueprint_path}"
 DEP_PIDS=({pids})
+DEP_SENTINELS=({dep_sentinels})
 
 {env_vars}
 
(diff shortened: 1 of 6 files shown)


ankona (approved):
LGTM!

PR #649, opened 2026-08-27
Adopt in-flight SLURM jobs on reload instead of duplicating submission

ScottEilerman:
* fix a bug where in-flight / in-queue jobs would get resubmitted
* fix a few docstring return types
* remove docs mention of prefect server
* ...maybe addressed a potentially noisy log message? here's the reasoning on that one, we can change this back if it seems wrong:

> _ignore_ambient_clobber_env is silent again for CSTAR_CLOBBER_WORKING_DIR=0 (the #635 behavior), still using the branch's new unset() helper, and the warning says "is set but is ignored" instead of "is unset and  ignored". Restored test_ignore_ambient_clobber_env_silent_when_off and dropped FLAG_OFF from the warn-test parametrize

### DIFF
--- a/cstar/orchestration/launch/slurm.py
+++ b/cstar/orchestration/launch/slurm.py
@@ -306,23 +306,6 @@ class SlurmLauncher(Launcher[SlurmHandle]):
         raise RuntimeError(msg)
 
     @staticmethod
-    async def _get_status(job_id: str) -> ExecutionStatus:
-        """Retrieve the status of a step running in SLURM.
-
-        Parameters
-        ----------
-        job_id : str
-            The slurm job ID to retrieve status for.
-
-        Returns
-        -------
-        ExecutionStatus
-            The current status of the step.
-        """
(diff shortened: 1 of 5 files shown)


cstar/orchestration/launch/slurm.py
```diff
@@ -395,15 +378,18 @@ async def launch(
...
             if Status.is_failure(last_status):
-                # force cache refresh for any tasks that didn't succeed
+                # clear prior state and re-run any tasks that didn't succeed
                 log.debug(f"Prior run of {name!r} in fail state. Re-running.")
-                # step.fsm.clear_prior()
                 step.workflow_overrides[KEY_CLOBBER] = True
-            elif Status.is_terminal(last_status):
-                # re-use the result from a prior run if it terminated
-                # successfully and isn't configured to be clobbered
+            elif Status.is_terminal(last_status) or Status.is_in_progress(last_status):
```
ankona, line 384: that's a great catch

PR #654, opened 2026-08-29
MPI_MASKING fixes for toy domains and coarse grids

ScottEilerman:
# Summary

Makes MPI_MASKING auto-tiling usable on small grids and turns several crash/corruption paths around coarse-resolution forcing into clear, actionable startup errors. Previously, `find_optimal_tiling` hard-rejected any candidate tiling carrying a coarse-grid caveat, so small domains (e.g. toy/test grids) could find no valid tiling at all and aborted with a confusing error followed by garbage output from ranks running past `MPI_Abort`. The scan now falls back to the most efficient candidate when the run reads no coarse-resolution forcing, rejects tilings that would break PIO in other ways, and fails fast with a clear message when coarse forcing genuinely rules out every option. (Changes are currently uncommitted on `pio_nml`.)

## Breaking Changes
- N/A

## New Features
- N/A

## Bug Fixes
- MPI_MASKING auto-tiling no longer refuses to run on small grids where every candidate tiling has coarse-grid caveats: when the run reads no coarse-resolution forcing (`interp_*_frc` all off), the most efficient candidate is selected with a prominent warning instead of aborting with "no valid MPI_MASKING tiling found".
- MPI_MASKING tilings that leave a physical domain edge with no ocean tile are now rejected during the scan; previously they passed selection and crashed later in PIO initialization with a cryptic `totiosize <= 0` assert.
- Coarse-resolution forcing (`interp_*_frc`) on a tiling too fine for the coarse grid now fails at the first coarse read with a clear error; previously it produced out-of-range PIO decompositions or silently wrong interpolated forcing near tile seams. This guard applies to all runs, including fixed `np_xi`/`np_eta` tilings without MPI_MASKING.
- A failed tiling scan no longer emits garbage output ("splitting communicator to 0 ranks", a duplicate scan printout, nonsense candidates like "mpirun -n 0" with overflowed points-per-tile): ranks running briefly past `MPI_Abort` were reading past the end of the candidate array, and unpartitionable decompositions were entering the candidate list.

## Improvements
- When coarse-resolution forcing is enabled and no compatible tiling exists for the launched core count, MPI_MASKING now errors at startup with a message naming the conflict and the remedies (disable forcing interpolation or change the rank count), instead of failing mid-run.

## Miscellaneous
- Regenerated `src/Make.depend` for the new dependencies of `main.F90`.

## Notes for Reviewers
- Verified end-to-end against the real failing case (cstar toy blueprint, 20×20 grid, `mpirun -n 10`): with coarse forcing off, the scan selects a 1×10 tiling and the run completes all steps and writes output; with `interp_bulk_frc = .true.`, the run stops at startup with the new error. The pytest golden-master suite was not run (no MPI_MASKING configuration is covered by it).
- Coarse-forcing intent is taken from the namelist flags, which are read before the tiling scan. `main.F90` gates on the CPP-active forcing module because `interp_flux_frc` is inert under `BULK_FRC` (`init_arrays_flux_frc` is `#ifndef BULK_FRC`).
- The edge-ownership rejection is a workaround for PIO's requirement that every boundary decomposition be non-empty; a deeper alternative (making `pio_roms` tolerate a fully-land physical edge) was deliberately left for follow-up.
- `find_optimal_tiling` gained a `coarse_frc` argument (module-internal interface; only caller is `main.F90`).

## Code Review Checklist
- [ ] Closes #xxxx
- [x] Any updates to namelist options or cppdefs are being propagated to C-Star/Forge (no namelist or cppdefs changes in this PR)
- [ ] New functionality has documentation and tests
- [x] Tests and `pre-commit run --all-files` are passing (pre-commit clean; pytest suite not run — see Notes for Reviewers)

### DIFF
--- a/cstar/cli/blueprint/run.py
+++ b/cstar/cli/blueprint/run.py
@@ -1,10 +1,13 @@
 import asyncio
 import typing as t
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _pkg_version
 from pathlib import Path
 
 import typer
 from pydantic import ValidationError
 
+import cstar
 from cstar.applications.core import (
     RunnerRequest,
     get_app_for_blueprint,
@@ -53,6 +56,21 @@ CMD_HELP: t.Final[str] = "Execute a blueprint in a local worker service."
 
 app = typer.Typer()
 log = get_logger(__name__)
+
+
+def _log_startup_versions() -> None:
+    """Best-effort startup log line recording installed CWorthy library versions,
+    so a run's own log records what generated it.
+    """
+    # TODO: warn when installed versions are known-incompatible with each other
+    # (e.g. roms-tools vs cstar-ocean) -- deferred as a follow-up; this only
+    # records what's installed.
+    versions = [f"cstar-ocean=={cstar.__version__}"]
+    try:
+        versions.append(f"roms-tools=={_pkg_version('roms-tools')}")
+    except PackageNotFoundError:
+        pass
+    log.info("Versions: %s", ", ".join(versions))
 
 
 def path_callback(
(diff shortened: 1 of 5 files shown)


ScottEilerman:
wrong repo bro (me)

PR #657, opened 2026-08-31
Add `cstar workplan gather` to consolidate per-step joined output

ScottEilerman:
# Summary

Adds a `cstar workplan gather <run-id>` CLI command that consolidates every step's `joined_output` files into a single run-level `joined_output/` directory of symlinks (next to `tasks/`), so a run's de-partitioned outputs can be browsed in one place.

## Breaking Changes
- N/A

## New Features
- New `cstar workplan gather <run-id>` command builds a run-level `joined_output/` directory of relative symlinks pointing at every file in the per-step `joined_output` directories. It is safe to re-run at any point — including while a workplan is still in progress — and each invocation replaces the symlink set with one reflecting the current state on disk. If two steps produce a file with the same name, the command aborts with an error listing every conflicting filename (and modifies nothing).

## Bug Fixes
- N/A

## Miscellaneous
- Added a "Gathering Workplan Outputs" section to the workplans documentation.

## Notes for Reviewers
- The run-id argument shared by `workplan log` and `workplan gather` (autocompletion, normalization, and run preloading) is now defined once as a reusable `RunIdArgument` in `cli/workplan/shared.py` instead of being duplicated per command.
- Safety semantics: `gather` never touches the per-step directories; it only manages the run-level view. It refuses (exit 1) to wipe the consolidated directory if it contains any non-symlink entries, and refuses to proceed if the destination path exists but is not a directory (e.g. a stray file or dangling symlink) — no real data is ever deleted.
- Symlinks are relative (`../tasks/<step>/joined_output/<file>`), so the view survives the run directory being moved or mounted at a different path.
- Steps without a `joined_output` directory (non-ROMS applications, not-yet-run steps) are skipped quietly; when nothing is found the command leaves an empty directory and says so.

## Code Review Checklist
- [ ] Closes #xxxx
- [x] Tests and `pre-commit run --all-files` are passing (full unit tier: 1658 passed; integration tier skipped per the tiered strategy — no ROMS-build-affecting change)

### DIFF
--- a/cstar/cli/workplan/gather.py
+++ b/cstar/cli/workplan/gather.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 from collections import defaultdict
 from pathlib import Path
 
@@ -89,12 +90,12 @@ def report_conflicts(conflicts: dict[str, list[Path]]) -> None:
 
 
 def relink(dest: Path, links: dict[str, Path]) -> None:
-    """Replace the symlinks in a consolidated joined-output directory.
+    """Replace a consolidated joined-output directory with a fresh set of symlinks.
 
-    Any existing symlinks in `dest` are removed before new ones are created.
-    If `dest` contains any entries that are not symlinks, nothing is
-    modified and a `typer.Exit` is raised, since gather must never delete
-    real files from the consolidated directory.
+    The existing directory is removed and recreated. If `dest` contains any
+    entries that are not symlinks, nothing is modified and a `typer.Exit` is
+    raised, since gather must never delete real files from the consolidated
+    directory.
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


cstar/cli/workplan/gather.py
```diff
@@ -0,0 +1,178 @@
...
+        for name, paths in sorted(conflicts.items())
+    ]
+    console.print(
+        "Refusing to gather: the following filenames are produced by more than "
+        "one step's joined_output directory:\n" + "\n".join(lines),
+        soft_wrap=True,
+    )
+
+
+def relink(dest: Path, links: dict[str, Path]) -> None:
```
ankona, line 92: Consider blowing away the gathered parent directory and just re-creating. I think this removes the need to do `relink` and keeps the traversal to a single place

PR #658, opened 2026-09-01
Add method to detect scratch dir on Bouchet

ScottEilerman:
# Summary

Teaches C-Star to find scratch space on Yale's Bouchet cluster. Bouchet exports no `$SCRATCH`-style env var (so the `CSTAR_SCRATCH_DIRS` lookup finds nothing and `CSTAR_DATA_HOME` silently fell back to `~/cstar` on the home quota); instead, each user's home carries per-project `scratch_pi_<pi-netid>` symlinks with a per-user subdirectory named after the username. System contexts can now supply a machine-specific scratch locator, and Bouchet's implements that heuristic.

## New Features

- New `SystemContext.scratch_directory()` classmethod: a machine-specific scratch-filesystem locator for systems that expose no scratch env var (default `None` — env vars suffice).
  - `BouchetSystemContext` implements it: the first `scratch_pi_*` directory under home (sorted for determinism, non-directories skipped) plus the current username.
  - `hpc_data_directory()` falls back to the active system context's locator when none of the `CSTAR_SCRATCH_DIRS` env vars are set, so `CSTAR_DATA_HOME` defaults onto Bouchet scratch instead of `~/.local/share`. Env vars still take precedence, and any failure to resolve the context degrades to the previous behavior instead of raising.

## Notes for Reviewers

- The `cstar.system.manager` import inside `hpc_data_directory()` is deliberately function-local: a module-level import would cycle (`system.manager → base.log → base.env`). It uses `get_system_context()` (pure env-var reads), not the heavier `get_sysmgr()`.
- The except clause covers `CstarError`, `OSError`, and `RuntimeError` (the last for `Path.home()` failing to resolve), keeping the function's never-raises contract.
- Companion cstar-forge PR (`bouchet-support` branch) adds the same heuristic to Forge's own host detection; the copies are independent by design (Forge's host resolution is documented as disposable) but must agree on detection and the discovered root.
- Multi-project users (several `scratch_pi_*` links): the first sorted match wins; no per-project selection yet.

## Code Review Checklist
- [ ] Closes #xxxx
- [x] Tests and `pre-commit run --all-files` are passing (unit tier: 1634 passed, 13 skipped, 2 xfailed; integration tier not run, per repo testing policy)

### DIFF
--- a/cstar/system/manager.py
+++ b/cstar/system/manager.py
@@ -1,8 +1,10 @@
 import functools
+import getpass
 import os
 import platform as platform
 from collections.abc import Sequence
 from dataclasses import dataclass
+from pathlib import Path
 from typing import ClassVar, Final, Protocol, override
 
 from pydantic import Field
@@ -139,6 +141,24 @@ class SystemContext(Protocol):
     def is_match(cls) -> bool:
         """Return `True` if the context identifies the current system as a match."""
         return False
+
+    @classmethod
+    def scratch_directory(cls) -> Path | None:
+        """Locate a machine-specific scratch filesystem for systems that expose no
+        scratch env var.
+
+        Some HPC systems provide no `SCRATCH`-style environment variable (see
+        `CSTAR_SCRATCH_DIRS`) and instead require a machine-specific heuristic to
+        locate the scratch filesystem. Override this classmethod to implement such
+        a heuristic.
+
+        Returns
+        -------
+        Path | None
+            The located scratch directory, or `None` when no machine-specific
+            heuristic applies (i.e. the system's scratch env vars are sufficient).
+        """
+        return None
 
 
 CTX_REGISTRY: dict[str, type[SystemContext]] = {}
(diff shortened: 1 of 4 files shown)


cstar/system/manager.py
```diff
@@ -140,6 +142,24 @@ def is_match(cls) -> bool:
...
+        locate the scratch filesystem. Override this classmethod to implement such
+        a heuristic.
+
+        Returns
+        -------
+        Path | None
+            The located scratch directory, or `None` when no machine-specific
+            heuristic applies (i.e. the system's scratch env vars are sufficient).
+        """
+        return None
```
ankona, line 161: ~~Let's get rid of `CSTAR_SCRATCH_DIRS` completely and use this method in the `SystemContext` to perform the default search.~~

ankona (changes requested):
Looks good. I like this on `SystemContext` instead of a standalone variable. Let's go all-in on it and dump the `CSTAR_SCRATCH_DIRS` env var

cstar/system/manager.py
ankona: An alternative here is to just have the users set up `SCRATCH=~/scratch_<ey_xxx>/<uid>` where `ey_xxx` is their slurm account .

I've set up project directories to work the same way as `CSTAR_SCRATCH_DIRS` (e.g. var containing CSV of indirect vars to look for) and i'm second guessing 1) making them inconsistent 2) updating project dirs to work this way, too.

A user can currently configure `CSTAR_SCRATCH_DIRS=SCRATCH_PRIORITY_0,SCRATCH_PRIORITY_1`, then choose to not set `SCRATCH_PRIORITY_0` to let it fall through to use `SCRATCH_PRIORITY_1`. The `SystemContext.xxx` solution made sense when I thought the comment (directory name is unpredictable) was true, but it's not - scratch is totally predictable (e.g. `ls ~/scratch_${CSTAR_SLURM_ACCOUNT}/${USER}`). 

I think this user-facing capability to manage the resolution order is a much better capability to keep around

ankona: test instead, main + adding to `bouchet.env`:

```
...
SCRATCH=~/scratch_${CSTAR_SLURM_ACCOUNT}/${USER}
```

ScottEilerman:
Closing -- setting $SCRATCH on Bouchet is the recommended route for now.

PR #664, opened 2026-09-02
Inspect cpus-per-node, set nodes=1 when appropriate

ScottEilerman:
# Summary

SLURM jobs that fit on one node are now kept on one node, and single-process applications can no longer request more CPUs than a node has. C-Star queries each partition's CPUs per node via `sinfo`, derives the minimum node count from it when no explicit node count is given, and pins `--nodes` in the batch script. A new `Blueprint.single_node` marker lets an application (e.g. cstar-forge's input generation, which runs as one dask process) declare that it can never span nodes: the launcher then pins the job to one node and clamps its CPU request to the target partition's per-node capacity instead of requesting extra nodes that would sit idle.

## Breaking Changes
- N/A

## New Features
- `Blueprint.single_node` (default `False`): a blueprint can declare that its work runs in a single process. The workplan transformer records this alongside the step's CPU requirement, and the SLURM launcher pins such a job to one node and clamps `num_cpus` to the queue's CPUs per node (a user-supplied `cpus_per_node` is treated as the capacity of record). A single-node job that also requests more than one node is rejected at job creation.
  - `SlurmComputeSpec.single_node` and a matching `single_node` argument on `create_scheduler_job` / `SchedulerJob`, so the marker can also be set directly in a step's `compute_overrides["slurm"]`.
- Per-partition CPU capacity: `SlurmPartition.max_cpus_per_node` queries `sinfo --format=%c` for the partition, returning the smallest node type's value on heterogeneous partitions so derived node counts are always satisfiable.

## Bug Fixes
- On systems that do not require an explicit task distribution, jobs were submitted with a bare `--ntasks`, leaving SLURM free to scatter tasks across nodes and stranding single-process steps on a fraction of their CPUs. The minimum node count is now derived from the partition's (or system's) CPUs per node and emitted as `--nodes`, with `--ntasks-per-node` added when a per-node capacity was supplied. When no capacity can be determined the job is submitted without a node count and a warning is logged.

## Improvements
- `cpus_per_node` on non-task-distribution systems now doubles as the assumed per-node capacity when deriving the node count, so a user can target a heterogeneous partition's larger node types.

## Miscellaneous
- N/A

## Notes for Reviewers
- The capacity fallback chain for a single-node job is `cpus_per_node` → partition `sinfo` value → scheduler-wide `global_max_cpus_per_node`. When all three are unknown, the job is pinned to one node with its full request and a warning is logged; SLURM will reject it if the request exceeds a node.
- The Bouchet system context still hardcodes `max_cpus_per_node=128`. That value is only the last-resort fallback above and is wrong for Bouchet's `mpi`/`bigmem` (64) and `day` (48-core smallest nodes) partitions; the per-partition `sinfo` query takes precedence, so this was left for a follow-up.

## Code Review Checklist
- [ ] Closes #xxxx
- [x] Tests and `pre-commit run --all-files` are passing (unit suite: 1687 passed, 14 skipped, 2 xfailed; pre-commit incl. mypy clean on changed files)

### DIFF
--- a/cstar/orchestration/transforms.py
+++ b/cstar/orchestration/transforms.py
@@ -518,7 +518,11 @@ class WorkplanTransformer(LoggingMixin):
                     # children have materialized blueprints (overrides already
                     # baked in); record their cpu requirement directly
                     transformed_steps.extend(
-                        _inject_cpus(child, child.blueprint.cpus_needed)
+                        _inject_cpus(
+                            child,
+                            child.blueprint.cpus_needed,
+                            single_node=child.blueprint.single_node,
+                        )
                         for child in overridden_steps
                     )
             else:
@@ -741,14 +745,17 @@ def preflight_overrides(step: LiveStep) -> LiveStep:
         step.blueprint, dict(step.blueprint_overrides)
     )
 
-    return _inject_cpus(step, merged.cpus_needed)
+    return _inject_cpus(step, merged.cpus_needed, single_node=merged.single_node)
 
 
-def _inject_cpus(step: LiveStep, cpus: int) -> LiveStep:
+def _inject_cpus(step: LiveStep, cpus: int, single_node: bool = False) -> LiveStep:
     """Record a step's cpu requirement in its `compute_overrides`.
 
     Declared values win; the requirement is injected only when
-    `["slurm"]["num_cpus"]` is absent.
+    `["slurm"]["num_cpus"]` is absent. When the blueprint declares itself
+    `single_node`, that is recorded alongside as `["slurm"]["single_node"]`
+    so the launcher can clamp the cpu count to one node's capacity without
+    reading the blueprint.
 
     Parameters
     ----------
@@ -756,6 +763,10 @@ def _inject_cpus(step: LiveStep, cpus: int) -> LiveStep:
         The step to enrich.
     cpus : int
         The cpu requirement read from the step's blueprint.
+    single_node : bool, optional
+        Whether the blueprint confines its work to one node. Only recorded
+        when `True`; a `False` value leaves the overrides untouched so that
+        an explicitly declared `single_node` is never overwritten.
 
     Returns
     -------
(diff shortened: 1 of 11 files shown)


cstar/orchestration/transforms.py
```diff
@@ -775,8 +786,12 @@ def _inject_cpus(step: LiveStep, cpus: int) -> LiveStep:
         )
         raise CstarExpectationFailed(msg)
 
+    injected: dict[str, t.Any] = {"num_cpus": cpus}
+    if single_node:
+        injected["single_node"] = True
```
ankona, line 791:
```suggestion
        injected["cpus_per_node"] = step.blueprint.cpus_needed
        injected["num_nodes"] = 1
```
doesn't this cause the scheduling you desire without adding single_node to this method and the create_scheduler_job method?

cstar/system/scheduler.py
```diff
@@ -105,6 +105,36 @@ def query_max_walltime_via_sinfo(name: str) -> str | None:
...
+    A heterogeneous partition reports one value per node type; the smallest is
+    returned so node counts derived from it are always satisfiable.
+
+    Returns
+    -------
+    int or None
+        The number of CPUs per node on the partition's smallest node type, or
+        `None` if the value cannot be determined.
+    """
+    sp_cmd = f"sinfo --noheader --format='%c' --partition={name}"
```
ankona, line 123:
```suggestion
    sp_cmd = f"sinfo --noheader --format='%.5c' --partition={name}"
```
pass a format string that's sufficiently wide (here, we're allowing up to 9,999 cpus-per-node before the output needs to add the "+") to indicate the output is truncated. this gets rid of the rstrip and some uncertainty about whether the value is truncated

ankona: update.... this outputs the minimum for the specified partitions:

```
x-cmcbride in 🌐 login07 in ~
❯ sinfo --partition=ai --format='%c'
CPUS
96

x-cmcbride in 🌐 login07 in ~
❯ sinfo --partition=ai,shared --format='%c'
CPUS
96+

x-cmcbride in 🌐 login07 in ~
❯ sinfo --partition=shared --format='%c'
CPUS
128
```

getting a plus means it's giving out a value that is aggregated and may not be the correct value you seek. We shouldn't silently swallow those "+" characters... Maybe we should report the query and the result to ensure that the limit returned can be investigated for correctness

ankona: per offline convo, heterogeneous nodes in a partition can cause this... sounds safe to me to get the "lowest max". resolving

ankona (changes requested):
Sanity check:
1. you really need to change `create_scheduler_job` versus adding a `compute_overrides["slurm"]["num_nodes"] = 1`
2. does enabling `BouchetSystemContext.scheduler.requires_task_distribution=True` provide an alternative path  


update:

let's ensure we use `@lru_cache` on `query_max_cpus_per_node_via_sinfo` so we don't repeatedly call it in the guise of an attribute (`max_cpus_per_node`)

ankona (approved):

PR #668, opened 2026-09-03
Migrate blueprints in workplan-run preflight

ScottEilerman:
# Summary

`cstar workplan run` now migrates each step's blueprint to the current schema before running it, so a workplan that references an older-schema blueprint no longer fails during preflight. (Changes are currently uncommitted on this branch.)

## Breaking Changes
- N/A

## New Features
- N/A

## Bug Fixes
- `cstar workplan run` no longer crashes with a validation error when a step's blueprint file is at an older schema version (for example, one still containing a `model_params` block). Blueprints are now migrated to the current schema automatically during workplan preflight, matching the behavior `cstar blueprint run` already had.

## Improvements
- When a blueprint genuinely cannot be migrated — no upgrade path to the current schema, or migration disabled via `CSTAR_DISABLE_MIGRATION` — the run now stops with a clear message naming the schema versions involved and pointing to `cstar blueprint migrate`, instead of a raw validation traceback.

## Miscellaneous
- N/A

## Code Review Checklist
- [ ] Closes #xxxx

## Notes for Reviewers

- The fix adds `load_blueprint(path)` in `cstar/applications/core.py`: it reads the blueprint leniently, migrates its schema in memory (no file is rewritten), then validates against the application's current schema. The three schedule-time strict-parse boundaries now route through it — `Step.blueprint` (`orchestration.py`), and `OverrideTransform.__call__` plus the runtime directive `preprocess` (`transforms.py`) — so the same class of failure can't resurface from another entry point.
- Migration is only attempted when the blueprint's application is actually covered by registered adapters (`app.name in schema_bounds`). This matters because an application can inherit another's adapters (the test-only `sleep` app subclasses `RomsMarblApplication`); those blueprints correctly fall through to strict validation rather than erroring.
- `execute_migration` / `_localize_and_migrate` (the `cstar blueprint run` CLI path) were intentionally left as-is — they persist the migrated blueprint to disk and are CLI-coupled. Collapsing them onto the new in-memory helper is a possible follow-up.

### DIFF
--- a/cstar/applications/core.py
+++ b/cstar/applications/core.py
@@ -9,12 +9,19 @@ from itertools import chain
 from pathlib import Path
 
 from cstar.base.adapter import SchemaAdapter
+from cstar.base.env import ENV_CSTAR_DISABLE_MIGRATION
+from cstar.base.feature import is_flag_enabled
 from cstar.base.log import get_logger
 from cstar.entrypoint.config import JOBFILE_DATE_FORMAT
 from cstar.execution.file_system import local_copy
 from cstar.execution.handler import ExecutionStatus
-from cstar.orchestration.models import BlueprintCore
+from cstar.orchestration.models import Blueprint, BlueprintCore
 from cstar.orchestration.serialization import SerializableModel, deserialize
+from cstar.system.migration import (
+    BlueprintMigration,
+    CstarMigrationError,
+    CstarUnsupportedMigrationError,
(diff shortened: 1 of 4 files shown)


ankona:
we might steal that `load_blueprint` in the future but this PR is superceded by https://github.com/CWorthy-ocean/C-Star/pull/669

PR #675, opened 2026-09-04
auto-compose workplan from blueprint

ankona:
# Summary

This PR adds automatic creation of a single-step `Workplan` when a `Blueprint` is passed to `cstar workplan run`.

## Breaking Changes

- N/A

## New Features

- Blueprints passed to `cstar workplan run ...` will be executed in the context of a workplan.

## Bug Fixes

- N/A

## Improvements

- Fix parameter typo and include capability to re-attach in main help text.

## Miscellaneous

- N/A

## Code Review Checklist

- [x] Closes #CSD-1189

### DIFF
--- a/cstar/cli/common.py
+++ b/cstar/cli/common.py
@@ -22,10 +22,9 @@ from cstar.base.feature import is_flag_enabled
 from cstar.base.log import LogLevelChoices, get_logger, reset_log_level
 from cstar.base.utils import slugify
 from cstar.execution.file_system import DirectoryManager, is_remote_resource, local_copy
-from cstar.orchestration.models import BlueprintCore, Step, Workplan
+from cstar.orchestration.models import BlueprintCore
 from cstar.orchestration.serialization import (
     PersistenceMode,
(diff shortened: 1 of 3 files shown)


ScottEilerman (changes requested):
No tests... tried it manually, it seems to drop a workplan (without any file extension) next to the blueprint, then still tries to run the blueprint as a workplan which fails validation. Seems like this needs to cook a little longer next week.

ankona:
> No tests... tried it manually, it seems to drop a workplan (without any file extension) next to the blueprint, then still tries to run the blueprint as a workplan which fails validation. Seems like this needs to cook a little longer next week.

I fixed 2 bugs. 
- Perform immediate return of the auto-composed workplan path instead of fall-through
- Fixed missing extension

Also fixed potential order-of-operations issue when both `run-id` and blueprint `path` are passed (resulting in attempt to deserialize workplan from BP).

ScottEilerman (approved):
working for me now!

PR #678, opened 2026-09-08
Pair run-summary steps with their own sentinels

ScottEilerman:
# Summary
The run summary printed at the end of `cstar workplan run` attributed each step's status, task/job ID, and state-file path to the wrong step. The summary paired the workplan's steps with the run's recorded sentinel files by position, but those paths are stored as an unordered set, so the pairing was arbitrary. The orchestrator's own "Closed node" log lines and the `cstar workplan status` table were unaffected, which is why the two disagreed. Each step now looks up its own sentinel by name.

## Breaking Changes
- N/A

## New Features
- N/A

## Bug Fixes
- The per-step "Job Details" block in the `cstar workplan run` summary showed another step's status and task/job ID; each step now reports its own.

## Improvements
- N/A

## Miscellaneous
- N/A

## Notes for Reviewers
- Steps that were never launched have no recorded sentinel; they fall back to the canonical sentinel path for the step and continue to display as `Unsubmitted`.
- The new regression test fails deterministically on the previous code across hash seeds (set ordering), and passes with the fix.

## Code Review Checklist
- [ ] Closes #xxxx
- [x] Tests and `pre-commit run --all-files` are passing (unit tier: `cstar/tests/unit_tests/orchestration`, 756 passed; integration tier not run)

### DIFF
--- a/cstar/orchestration/dag_runner.py
+++ b/cstar/orchestration/dag_runner.py
@@ -501,18 +501,26 @@ class ExecutiveRunSummary(BaseModel):
         steps = [LiveStep.from_step(s) for s in workplan.steps]
         step_summaries: list[ExecutiveStepSummary] = []
 
+        # `run.sentinels` is an unordered set, so pair each step with its own
+        # sentinel by file name rather than by position. Steps that were never
+        # launched have no recorded sentinel and fall back to the canonical path.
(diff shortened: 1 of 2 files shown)


ankona (approved):
🤦 great catch. it looked right until it didn't.

PR #679, opened 2026-09-08
Add manual instructions for system context additions

ankona:
# Summary

Add generated documentation for manual system context additions.

## Breaking Changes

- N/A

## New Features

- N/A

## Bug Fixes

- N/A

## Improvements

- N/A

## Miscellaneous

- N/A

## Code Review Checklist

- [x] Closes #CSD-1012


docs/machines.rst
```diff
@@ -10,17 +10,21 @@ it has been tested on a limited number.
 |                                                                                      | build/run  |                         |
 |                                                                                      | support    |                         |
 +======================================================================================+============+=========================+
+| `Anvil (RCAC) <https://www.rcac.purdue.edu/anvil>`__                                 | ✅         | ✅                      |
++--------------------------------------------------------------------------------------+------------+-------------------------+
+| `Bouchet (YCRC) <https://docs.ycrc.yale.edu/clusters/bouchet/>`__                    | ✅         | Untested                |
```
ScottEilerman: Should we mark Bouchet and Elja "tested" at this point?

ScottEilerman: Additionally, Ben has done a fair amount of testing on non-HPC Linus at this point, I guess we could add that with the same caveats as MacOS... though actually those caveats aren't true anymore with the proxy local launcher. :flex:

docs/system-registration.rst
```diff
@@ -0,0 +1,123 @@
...
+----------------------------
+
+Add an entry for the system to ``~/.ssh/config`` if one does not already
+exist, for example::
+
+    Host <system-name>
+        HostName <login-node-address>
+        User <username>
+
+All remaining steps require an SSH session on the system.
```
ScottEilerman: I wonder if we should remove or generalize this SSH section as a "step 0"-- gesture vaguely at "you need to be able to access the machine" and that your HPC should provide instructions for you on how to do that? I feel like each HPC has been different enough in terms of auth, virtual terminals, etc.

docs/system-registration.rst
```diff
@@ -0,0 +1,123 @@
...
+
+#. List the modules loaded by default: entries marked ``(d,L)`` in the
+   output of ``module avail``.
+#. List the available netCDF packages with ``module avail "netCDF"``.
+#. Choose a netCDF Fortran package built with the GNU compiler.
+#. Choose a parallel netCDF package built with the same compiler version.
+#. Inspect dependencies pulled in by the netCDF packages with
+   ``module spider <package-name>``.
+#. If a GCC compiler is not already loaded by default or as a dependency,
+   find a compatible one with ``module avail``. Do the same for Open MPI
```
ScottEilerman: I wonder if MPI should be its own required line item -- it might get pulled in from one of the other modules, but it's definitely an important one, and could be mpich or another flavor instead of openmpi. I don't actually know what the decision space is in identifying/choosing the right one, other than "this one was loaded by default and it worked fine."

docs/system-registration.rst
```diff
@@ -0,0 +1,123 @@
...
+   with the netCDF ``lib`` directory.
+#. Set ``GIT_DISCOVERY_ACROSS_FILESYSTEM="1"`` so git works on distributed
+   filesystems such as Lustre.
+#. Comment out any variables copied from the template that do not apply.
+
+Example (``perlmutter.env``)::
+
+    MPIHOME=${CRAY_MPICH_PREFIX}/
+    NETCDFHOME=${CRAY_NETCDF_PREFIX}/
+    PNETCDFHOME=${CRAY_PARALLEL_NETCDF_PREFIX}/
```
ScottEilerman: A previous update _theoretically_ standardized our path env-vars to not use trailing `/`, but it seems it didn't get them all in perlmutter.env, and maybe a few others. May be out of scope for this PR to change and test them all, though looks like you have another branch that's doing some updates anyway, so we should probably do it there, and then update these docs as well with the convention.

docs/system-registration.rst
```diff
@@ -0,0 +1,123 @@
...
+In ``cstar/system/manager.py``, add
+``class <SystemName>SystemContext(SystemContext)``, placed in alphabetical
+order among the existing subclasses and decorated with
+``@register_sys_context``:
+
+#. Set the ``name``, ``compiler``, ``mpi_prefix``, and ``docs`` class
+   variables (see ``AnvilSystemContext`` for reference).
+#. Locate the system's web-based SLURM documentation and implement
+   ``create_scheduler``, returning a ``SlurmScheduler`` with one
+   ``SlurmPartition`` per documented partition.
```
ScottEilerman, line 149: This makes me wonder if we need to rip out the PBS-based machines. My guess is that we don't actually support them at this point. Probably a follow-on task...

ScottEilerman (commented):
overall, looking pretty good -- a few suggestions, some of which should probably be follow-on items.

docs/machines.rst
ankona: i did an update with latest `cstar-system` skill that located a newer set of modules. I'd like to re-test using my other PR branch (refresh-config) on bouchet before i mark it tested, but ... almost there.

docs/system-registration.rst
ankona: I'm going to regenerate this. I reworked the skill that generates these docs to generalize across a list of required deps instead of itemmized. Will fix

docs/system-registration.rst
ankona: i'll fix this trailing slash in my config update PR

docs/system-registration.rst
ankona: yes, PBS is totally borked ATM.

docs/machines.rst
ankona: Done

docs/system-registration.rst
ankona: Generalized

docs/system-registration.rst
ankona: - MPI is it's own line item in required modules list.
- Instructions are updated to "locate each module..."

docs/system-registration.rst
ankona: fixed in config refresh branch

docs/system-registration.rst
ankona: Added follow-up ticket for determining if PBS support should be improved or deprecated

ScottEilerman (approved):

PR #682, opened 2026-09-10
Overhaul nest-from/continue-from overlap, add start_date warning

ScottEilerman:
# Summary

Two related problems motivated this branch. First, ROMS takes its clock from the restart file while C-Star computes `ntimes` from `start_date`, so a disagreement silently ran the wrong period: a 45-day segment started from a 2023-02-01 restart instead of the intended 02-15, "succeeded" on 03-18, and the next segment then failed against boundary data starting 04-01. Second, `nest-from` bundled initial conditions and boundaries, so it could not be combined with `continue-from` for later child segments, and the two directives disagreed about which per-step directory (`output` vs `joined_output`) held a step's results.

This PR makes `nest-from` a boundary-only directive that shares `continue-from`'s `path` / `step` configuration, warns when a restart and an explicit `start_date` disagree, and fails fast on missing local inputs. It also retires the per-step `joined_output` directory: every step's final whole files now live in `output`, non-ParallelIO ROMS runs write their partitioned pieces to a new `temp_output` directory, and a new `cstar admin migrate-outputs` command converts run directories produced by earlier releases. Includes #683 and #690.

## Breaking Changes
- The per-step `joined_output` directory no longer exists. Whole (joined) output files live in `output`. Workplans and blueprints that referenced `.../joined_output/...` paths must point at `.../output/...` instead; the shipped tutorials were updated.
- Runs completed with an earlier release must be migrated with `cstar admin migrate-outputs <run or step dir>` before a new step continues from them or before gathering. `continue-from` and `nest-from` raise with that hint when they find the old layout.
- Non-ParallelIO ROMS runs now write partitioned files to `temp_output`, and every partition piece (restarts included) is deleted once its join succeeds. Previously partitioned restart files were kept in `output`.
- `continue-from: step:` and `nest-from: step:` read only the referenced step's `output` directory and never use partitioned files. A non-ParallelIO follow-on step re-partitions the whole restart at setup, which costs some time but removes a class of directory-layout bugs.
- `cstar workplan gather` writes to `<run dir>/gathered_output` instead of `<run dir>/joined_output`, and reads each step's `output` directory (so nest_ic and upscaler outputs are gathered too). Duplicate file names across steps no longer abort the command: each duplicate is linked as `<step>__<file>`.
- `nest-from` no longer requires `rst_path`. The key is deprecated and still honored for one release with a `FutureWarning`; use `continue-from` for the restart instead. A step that sets both `nest-from.rst_path` and `continue-from` is rejected at scheduling time and at runtime, since both would set the initial conditions.
- `nest-from.bry_path` is deprecated in favor of `nest-from.path` (still honored for one release with a `FutureWarning`).
- `continue-from` and `nest-from` now reject a configuration that sets both `path` and `step`. Previously `step` silently won.

## New Features
- `nest-from` takes the same `path` / `step` configuration as `continue-from`; `step: <parent>` finds the parent's boundary files in its `output` directory, so users no longer hand-build the boundary path.
- `continue-from` and `nest-from` can be combined on one step: the first child segment continues from the `nest_ic` step and takes boundaries from the parent; later child segments continue from the previous child segment with the same `nest-from`. The nesting tutorial workplan shows the new shape.
- `continue-from` and `nest-from` warn when a step explicitly sets `runtime_params.start_date` in its `blueprint_overrides` and that date differs from the restart file the directive selected. The restart date still wins, as before. Chained steps that leave `start_date` to the directive are not warned about.
- The ROMS-MARBL runner warns when a blueprint's initial-conditions file has a restart-style name (`*_rst.<YYYYmmddHHMMSS>.nc`) dated differently from `start_date`, catching hand-assembled blueprints.
- `cstar admin migrate-outputs <path> [--dry-run]` moves `joined_output` contents into `output`, moves leftover partition pieces into `temp_output`, removes old run-level gather symlink directories, and reports collisions without overwriting anything.
- `{{temp_output_dir: <step>}}` is available as a workplan template placeholder for ROMS steps.

## Bug Fixes
- Staging a local input dataset whose file does not exist now raises `FileNotFoundError` at setup, instead of creating a dangling symlink and letting the model fail later with an opaque PIO abort.

## Improvements
- The `step:` resolution shared by `continue-from` and `nest-from` is a single directory lookup with no ParallelIO or blueprint introspection.
- `RomsMarblTimeSplitter` predicts the next segment's restart as a whole file in `output`.
- The ROMS-specific file-system manager moved from the core execution package into the ROMS-MARBL application package, and the run-level gather directory is exposed as `StateDirectoryManager.user_dir()`.
- ROMS namelist output roots are derived from the step's directory layout relative to the run directory instead of hard-coded strings.

## Miscellaneous
- Tutorials (`docs/tutorials/*.yaml`, `tutorial_bp.ipynb`, `tutorial_wp.ipynb`) and the workplans guide describe the new layout, the `gathered_output` directory, and the migration command.

## Notes for Reviewers
- The `continue-from` warning deliberately compares against the explicit workplan override (read from the packaged `apply-overrides` directive config), not the base blueprint's `start_date`. Comparing against the base value would warn on every normal chained step, since the base date is unrelated to whatever restart gets discovered at run time. The runner's filename check and the directive warning never double-warn: for workplan steps the directive has already aligned `start_date` with the restart by the time the runner check runs.
- `nest-from` gets no restart warning during the deprecation period; `continue-from` is the only restart directive going forward, so the warning lives there rather than in a shared base class. The two directives now touch disjoint blueprint keys (`initial_conditions` + `start_date` vs `boundary`), so their relative order no longer matters; only the existing apply-overrides-first rule remains.
- Schedule-time validation goes through a new no-op `Directive.validate_directives` classmethod called from `package_runtime_overrides`, so the orchestration layer never imports application-specific directive keys. Unregistered directive keys are skipped there and caught by the runtime check.
- The `step:` form of both directives refuses to use a partition piece found in a step's `output` (only possible under the old layout) and raises with the migration hint instead of silently continuing from a single processor's file. A referenced step with no `output` directory yet raises a clear "has no output directory" error rather than the finder's generic message.
- `post_run` in ParallelIO mode only counts and logs the files in `output`; there is no move step any more. `RomsFileSystemManager.PARTITIONED_OUTPUT_GLOB` is the single definition of the partition-piece pattern, shared by `post_run` and the migration command.
- The namelist `output_root_name` / `extract_root_name` values are computed relative to `run_dir` (where ROMS runs) and resolve to `../output` (ParallelIO) or `../temp_output` (non-ParallelIO). A user `namelist_overrides.output_root_name` still wins, as before, and `post_run` keeps reading the C-Star directories in that case (pre-existing behavior).

### DIFF
--- a/cstar/applications/roms_marbl/app.py
+++ b/cstar/applications/roms_marbl/app.py
@@ -15,7 +15,10 @@ from cstar.applications.roms_marbl.migration import (
     RomsMarblSchemaAdapterV21V3,
 )
 from cstar.applications.roms_marbl.models import APP_NAME, RomsMarblBlueprint
-from cstar.applications.roms_marbl.transforms import RomsMarblTimeSplitter
+from cstar.applications.roms_marbl.transforms import (
+    RomsMarblTimeSplitter,
+    warn_on_restart_start_date_mismatch,
+)
 from cstar.base.exceptions import CstarError
 from cstar.base.utils import slugify
 from cstar.entrypoint.config import (
@@ -84,6 +87,12 @@ class RomsMarblRunner(BlueprintRunner[RomsMarblBlueprint]):
         if not self.simulation:
             msg = "Simulation creation failed. Unable to execute simulation runner"
             raise RuntimeError(msg)
+
+        ic = self.simulation.initial_conditions
+        if ic is not None and self.simulation.start_date is not None:
+            warn_on_restart_start_date_mismatch(
+                ic.source.location, self.simulation.start_date, log=self.log
+            )
 
         self.log.trace("Setting up simulation")
         self.simulation.setup()
--- a/cstar/applications/roms_marbl/transforms.py
+++ b/cstar/applications/roms_marbl/transforms.py
@@ -1,7 +1,8 @@
 import os
 import re
 import typing as t
-from collections.abc import Sequence
+import warnings
+from collections.abc import Mapping, Sequence
 from datetime import datetime
 from pathlib import Path
 
@@ -21,8 +22,10 @@ from cstar.base.feature import (
     ENV_FF_ORCH_TRX_TIMESPLIT_LONGNAME,
     is_feature_enabled,
 )
+from cstar.base.log import get_logger
 from cstar.base.utils import (
     DEFAULT_OUTPUT_ROOT_NAME,
+    coerce_datetime,
     deep_merge,
     min_padded_index,
     slugify,
@@ -31,6 +34,7 @@ from cstar.execution.file_system import RomsFileSystemManager
 from cstar.orchestration.orchestration import LiveStep
 from cstar.orchestration.serialization import serialize
 from cstar.orchestration.transforms import (
+    ApplyOverridesDirective,
     DirectiveConfig,
     OverrideDirective,
     SplitFrequency,
@@ -38,6 +42,12 @@ from cstar.orchestration.transforms import (
     get_time_slices,
 )
 from cstar.orchestration.utils import ENV_CSTAR_ORCH_TRX_FREQ
+
+if t.TYPE_CHECKING:
+    from cstar.base.log import TraceLogger
+    from cstar.orchestration.orchestration import LiveWorkplan
+
+log = get_logger(__name__)
 
 
 class RomsMarblTimeSplitter(Transform[LiveStep]):
@@ -390,6 +400,61 @@ class RestartFile(BaseModel):
         return self._ts.strftime(self.FMT_TS)
 
 
+def restart_timestamp(location: str | Path) -> datetime | None:
+    """Return the timestamp encoded in a restart-style file name, or `None`.
+
+    Only the file name is inspected (via `RestartFile`'s validators); the file
+    need not exist.
+
+    Parameters
+    ----------
+    location : str | Path
+        A path or file name to inspect.
+
+    Returns
+    -------
+    datetime | None
+        The timestamp encoded in the file name, or `None` if `location` does
+        not match the restart file naming convention.
+    """
+    try:
+        return RestartFile(path=Path(location)).timestamp
+    except ValueError:
+        return None
+
+
+def warn_on_restart_start_date_mismatch(
+    location: str | Path, start_date: datetime, *, log: "TraceLogger"
+) -> None:
+    """Warn when a restart-style initial-conditions file is dated differently
+    from `start_date`.
+
+    ROMS takes its clock from the restart file while `ntimes` is computed from
+    `start_date`, so a mismatch means the run will not cover the intended
+    period.
+
+    Parameters
+    ----------
+    location : str | Path
+        The path or file name of the initial-conditions file.
+    start_date : datetime
+        The blueprint's configured start date.
+    log : TraceLogger
+        The logger to emit the warning to.
+    """
+    ts = restart_timestamp(location)
+    if ts is not None and ts != start_date:
+        log.warning(
+            "Restart file %s is dated %s but start_date is %s. ROMS starts "
+            "from the restart's time; the previous segment may not have "
+            "reached its end date, or the restart directory / start_date is "
+            "wrong.",
+            Path(location).name,
+            ts,
+            start_date,
+        )
+
+
 class RestartFileTrxAdapter:
     """Convert a restart file into a dictionary useful for use in an OverrideTransform."""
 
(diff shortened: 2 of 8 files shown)


cstar/orchestration/transforms.py
```diff
@@ -846,6 +846,11 @@ def package_runtime_overrides(step: LiveStep) -> LiveStep:
     LiveStep
         The transformed step.
     """
+    for directive_key, directive_config in step.directives.items():
+        directive_cls = DirectiveConfig.directive_map.get(directive_key)
+        if directive_cls is not None and isinstance(directive_config, Mapping):
+            directive_cls.validate_directives(directive_config, step.directives)
```
ankona: Is it possible to use the built-in `pydantic` validation hooks instead of adding this classmethod (e.g. `@model_validator` or `@field_validator`)?

cstar/orchestration/transforms.py
```diff
@@ -908,6 +913,33 @@ def key(cls) -> str:
         """
         ...
 
+    @classmethod
+    def validate_directives(
```
ankona: This should either return a `bool` or have some specific `raise` types specified. As-is, we don't know what happens if something is invalid.

cstar/orchestration/transforms.py
```diff
@@ -908,6 +913,33 @@ def key(cls) -> str:
...
+            This directive's own configuration mapping.
+        directives : Mapping[str, t.Any]
+            The step's full directives mapping (this directive's key
+            included).
+
+        Returns
+        -------
+        None
+        """
+        return
```
ankona: per docstring and my prior comment, this might fit into a `pydantic` validator in `Workplan` - that model holds all directives in context...

cstar/orchestration/transforms.py
```diff
@@ -908,6 +913,33 @@ def key(cls) -> str:
         """
         ...
 
+    @classmethod
```
ankona: consider converting this from an un-implemented base-class method to a runtime-checkable protocol

cstar/applications/roms_marbl/app.py
```diff
@@ -85,6 +88,12 @@ def _on_start(self) -> None:
             msg = "Simulation creation failed. Unable to execute simulation runner"
             raise RuntimeError(msg)
 
+        ic = self.simulation.initial_conditions
```
ankona: i think this could be done in Workplan validators? same deal... we would keep validation logic in the place that owns the attributes and avoid duplicating it or having to search for where changes have to occur. 

it will also let us validate as early as possible (instance creation) AND better support validation during `--dry-run.` Putting this in the runner guarantees that we can't report issues without executing.

cstar/applications/roms_marbl/transforms.py
```diff
@@ -625,9 +692,91 @@ def adapt(cls, bry_files: Sequence[BoundaryFile]) -> dict[str, t.Any]:
...
+    if name not in workplan:
+        msg = f"Unable to locate step {name!r} in workplan"
+        raise KeyError(msg)
+
+    fsm = RomsFileSystemManager(workplan[name].fsm.root_dir)
+    candidates = (fsm.joined_output_dir, fsm.output_dir)
+
+    for search_path in candidates:
+        if search_path.is_dir() and (found := find(search_path)) is not None:
+            return found
```
ankona: consider an explicit use of `next` here to make it clear that we aren't really looping or trying to touch everything in candidates.

```python
if found := next(x for x in candidates if x.is_dir() and find(x)):
    return found
```

cstar/applications/roms_marbl/transforms.py
```diff
@@ -625,9 +692,91 @@ def adapt(cls, bry_files: Sequence[BoundaryFile]) -> dict[str, t.Any]:
...
+    step_name : str | None
+        The step's name, when known; `None` when unavailable.
+
+    Returns
+    -------
+    str
+    """
+    subject = f"step {step_name!r}" if step_name is not None else "a step"
+    return (
+        f"nest-from rst_path and continue-from both set initial conditions for "
```
ankona: unnecessary f-string use on first line
```suggestion
        "nest-from rst_path and continue-from both set initial conditions for "
```

cstar/applications/roms_marbl/transforms.py
```diff
@@ -639,6 +788,39 @@ class ContinuanceDirective(OverrideDirective):
...
+        -------
+        Sequence[LiveStep]
+            Zero-to-many steps resulting from applying the transform.
+        """
+        data = self._system_overrides.get("initial_conditions", {}).get("data", [])
+        location = data[0].get("location") if data else None
+
+        explicit_start: t.Any = None
+        config = step.directives.get(ApplyOverridesDirective.key(), {})
+        if isinstance(config, Mapping):
```
ankona: we don't need these `isinstance` checks if we return empty dictionaries. consider:

```python
config = step.directives.get("AOD", {})
overrides = config.get("ko", {})
rt_params = overrides.get("rp", {})
sd = rt_params.get("start_date", None)

if location and sd:
  ...
```

no nested conditionals necessary, type checks, or explicit null checks required.

cstar/applications/roms_marbl/transforms.py
```diff
@@ -666,37 +848,21 @@ def _generate_overrides(self) -> dict[str, t.Any]:
...
-                if (
-                    isinstance(blueprint, RomsMarblBlueprint)
-                    and blueprint.partitioning.use_pio
-                ):
-                    search_path = fsm.joined_output_dir
-
-            else:
-                msg = f"Unable to locate step {name!r} in workplan"
-                raise KeyError(msg)
+            step_restart = find_in_step_outputs(self.workplan, name, RestartFile.find)
```
ankona: wouldn't this be the same?

```python
roms_fsm = RomsFileSystemManager(step.fsm.root)
step_restart = next(roms_fsm.rglob(pattern, depth=2), None)
```

I like the idea of having `Xxx.attr` specify the specific search params, but we might be able to reduce the specification to a glob pattern and only house the recursive search part in one spot. After all "find a file" isn't going to change...

Maybe what we really want is a _filter_function_ like `Callable[[Path], bool]` to be specified so you could make arbitrarily complex / non-glob filters:

```python
def is_rst(p: Path) -> bool:
  return p.stem.ends_with("_rst")
```

This returns all restart files, but could be composed with another condition, or held in a stateful instance

```python
class RstFilter:
  def __init__(self, dt: datetime):
    self.dt = dt

  def is_rst(self, p: Path) -> bool:
    return f"_rst.{self.dt.strftime('%Y-%m-%d ...')}" in str(p)
...
# now, i can apply my simple filter starting from end date and
# decrement date in an outer loop... filter is dead simple and
# can be the injected behavior instead


next(filter(p for p in paths,  RstFilter(d).is_rst) for d in reversed(TimeSpan(start_date, end_date))
```

cstar/applications/roms_marbl/transforms.py
```diff
@@ -719,45 +885,184 @@ def suffix() -> str:
...
+    `bry_path` is a deprecated alias for `path` (conflicts with `path`/
+    `step` if both are supplied). `rst_path` is a deprecated, optional key
+    that restores the historical "nest-from also sets initial conditions"
+    behavior by applying a restart-file override directly; it is rejected
+    when a `continue-from` directive is also present on the step, since both
+    would set `initial_conditions`. Both deprecated keys emit a
+    `FutureWarning` and a log warning.
+    """
+
+    KEY_PATH: t.Final[str] = ContinuanceDirective.KEY_PATH
```
ankona: this is a strange/wonky dependency. i understand the desire is probably to have consistent keys....

Consider creating a standard constant in `cstar.orch.transforms`, like `DIRECTIVE_KEY_PATH: t.Final[str] = "path"` that is used by both `ContinuanceDirective` and `NestingDirective`

The result:

- remove declarations as members in each directive class
- no coupling between unrelated directive types

cstar/applications/roms_marbl/transforms.py
```diff
@@ -719,45 +885,184 @@ def suffix() -> str:
...
+
+        Raises
+        ------
+        ValueError
+            If the deprecated `rst_path` key is set alongside a
+            `continue-from` directive on the same step; both would set
+            `initial_conditions`.
+        """
+        if (
+            self.KEY_RST_PATH in self._config
```
ankona: this duplicates the entire body of `validate_directives` and could instead re-use the validator above

```python
def __call__(...):
  self.validate_directives(self._config, step.directives)
```

cstar/applications/roms_marbl/transforms.py
```diff
@@ -719,45 +885,184 @@ def suffix() -> str:
...
+        if self.KEY_BRY_PATH in self._config:
+            msg = (
+                f"{self.key()!r} config key {self.KEY_BRY_PATH!r} is deprecated "
+                f"and will be removed in a future release; use {self.KEY_PATH!r} "
+                "instead."
+            )
+            warnings.warn(msg, FutureWarning, stacklevel=2)
+            log.warning(msg)
+
+            if self.KEY_PATH in self._config or self.KEY_STEP in self._config:
```
ankona: definitely looks like we could validate this in a `@model_validator`  and not need to generate overrides. 

bonus points for being able to reduce conditional checks in the "main logic" since the validator would guarantee state meets requirements first

PR #683, opened 2026-09-11
Refactor joined_output -> output

ScottEilerman:
# Summary

Every workplan step now has a single final-output directory, `output`. The `joined_output` directory is gone. ROMS runs without ParallelIO write their partitioned files to a new `temp_output` directory, and the post-run join places whole files in `output` and removes the partition pieces. ParallelIO runs write whole files straight into `output`, so nothing is moved after the run. `continue-from` and `nest-from` `step:` look only in `output`, and `cstar workplan gather` reads `output` and writes a run-level `gathered_output`. A new `cstar admin migrate-outputs` command converts run directories produced by earlier releases.

Base branch: `warn-restart-start-date` (may be folded into that PR).

## Breaking Changes
- The per-step `joined_output` directory no longer exists. Whole (joined) output files live in `output`. Workplans and blueprints that referenced `.../joined_output/...` paths must point at `.../output/...` instead; the shipped tutorials were updated.
- Non-ParallelIO ROMS runs now write partitioned files to `temp_output`, and every partition piece (restarts included) is deleted once its join succeeds. Previously partitioned restart files were kept in `output`.
- `continue-from: step:` and `nest-from: step:` read only the referenced step's `output` directory. They no longer prefer partitioned restart or boundary files; a non-PIO follow-on step re-partitions the whole file at setup, which costs some time but removes a class of directory-layout bugs.
- `cstar workplan gather` writes to `<run dir>/gathered_output` instead of `<run dir>/joined_output`, and reads each step's `output` directory (so nest_ic and upscaler outputs are gathered too). Duplicate file names across steps no longer abort the command: each duplicate is linked as `<step>__<file>`.
- Runs completed with an earlier release must be migrated with `cstar admin migrate-outputs <run or step dir>` before a new step continues from them or before gathering.

## New Features
- `cstar admin migrate-outputs <path> [--dry-run]` moves `joined_output` contents into `output`, moves leftover partition pieces into `temp_output`, removes old run-level gather symlink directories, and reports collisions without overwriting anything.
- `{{temp_output_dir: <step>}}` is available as a workplan template placeholder for ROMS steps.

## Bug Fixes
- N/A

## Improvements
- The `step:` resolution shared by `continue-from` and `nest-from` is a single directory lookup with no ParallelIO or blueprint introspection.
- `RomsMarblTimeSplitter` predicts the next segment's restart as a whole file in `output`.

## Miscellaneous
- Tutorials (`docs/tutorials/*.yaml`, `tutorial_bp.ipynb`, `tutorial_wp.ipynb`) and the workplans guide describe the new layout, the renamed gather directory, and the migration command.

## Notes for Reviewers
- `post_run` in ParallelIO mode only counts and logs the files in `output`; there is no move step any more.
- The `step:` form of both directives refuses to use a partition piece found in a step's `output` (only possible under the old layout, which left partitioned restarts there) and raises with the migration hint instead of silently continuing from a single processor's file. A referenced step with no `output` directory yet raises a clear "has no output directory" error rather than the finder's generic message.
- `RomsFileSystemManager.PARTITIONED_OUTPUT_GLOB` is the single definition of the partition-piece pattern, shared by post_run and the migration command.
- The namelist `output_root_name` / `extract_root_name` wiring in `ROMSSimulation` derives one relative root from `use_pio` (`../output` or `../temp_output`). A user `namelist_overrides.output_root_name` still wins, as before, and post_run keeps reading the C-Star directories in that case (pre-existing behavior).
- The migration command never overwrites: a name that already exists in the destination is left in place and reported. Symlink-only `joined_output` directories are treated as old gather results and removed; the user is told to re-run gather.

## Code Review Checklist
- [ ] Closes #xxxx
- [x] Tests and `pre-commit run --all-files` are passing (unit tier: 1730 passed, 13 skipped, 2 xfailed; integration tier not run)


cstar/cli/admin/migrate_outputs.py
```diff
@@ -0,0 +1,276 @@
...
+        ),
+    ] = False,
+) -> None:
+    """Migrate an old-layout run or step directory to the current output layout."""
+    report = migrate_output_layout(path, dry_run=dry_run)
+
+    # `markup=False` everywhere below: paths and the "[dry-run]" prefix are
+    # plain text, not Rich markup (a literal `[dry-run]` would otherwise be
+    # parsed as an unknown style tag and silently dropped).
+    prefix = "[dry-run] " if dry_run else ""
```
ankona: modify the _entity_ returned by some operation instead of relying on magic strings.

...

or just don't use markup-like syntax that causes us to have this comment and logic to handle it.

cstar/cli/workplan/gather.py
```diff
@@ -8,89 +9,120 @@
...
+        list of source paths for names that still collide after mangling,
+        and the number of files whose link name was mangled.
     """
-    by_name: dict[str, list[Path]] = defaultdict(list)
+    by_basename: dict[str, list[tuple[LiveStep, Path]]] = defaultdict(list)
     skipped: list[str] = []
 
     for step in workplan.steps:
-        src_dir = RomsFileSystemManager(step.working_dir).joined_output_dir
+        src_dir = JobFileSystemManager(step.working_dir).output_dir
```
ankona: this reminds me that `RomsFileSystemManager` really should be moved out of `cstar.execution.file_system` and into the app-specific utilities so it's more clear that core code shouldn't rely on it

cstar/cli/workplan/gather.py
```diff
@@ -143,23 +175,21 @@ def gather(
...
+    links, conflicts, mangled_count = collect_links(workplan)
 
     if conflicts:
         report_conflicts(conflicts)
         raise typer.Exit(1)
 
-    dest = RomsFileSystemManager(
-        StateDirectoryManager.data_dir(run_id)
-    ).joined_output_dir
+    dest = StateDirectoryManager.data_dir(run_id) / GATHERED_OUTPUT_NAME
```
ankona: I think `gather` makes sense as a standard operation for any type of step/blueprint. Let's shift responsibility for maintaining this directory of state information to `StateDirectoryManager.gather_dir` and move our `GATHERED_OUTPUT_NAME` constant

cstar/cli/workplan/gather.py
```diff
@@ -173,6 +203,13 @@ def gather(
         soft_wrap=True,
     )
 
+    if mangled_count:
```
ankona: I don't know if it really makes sense to try to just dump outputs into a single directory (necessitating the mangling). Are we opposed to letting gather contain step-specific directories for a reason? It could be argued that looking in one place is easier if names don't have to be mangled but i think it's overall easier to just avoid mangling and let the user look for the expected output names in directories with fewer files.

cstar/execution/file_system.py
```diff
@@ -337,9 +341,11 @@ def _codebases_dir(self) -> Path:
         return self.input_dir / self._CODEBASES_NAME
 
     @property
-    def joined_output_dir(self) -> Path:
-        """The directory for de-partitioned outputs."""
-        return self.root_dir / self._JOINED_OUTPUT_NAME
+    def temp_output_dir(self) -> Path:
```
ankona: this is a roms-specific temp dir and should be moved to `RomsFileSystemManager`. The roms-only nature is further enforced above with the `PARTITIONED_OUTPUT_GLOB` constant.

cstar/execution/file_system.py
```diff
@@ -295,7 +295,11 @@ class RomsFileSystemManager(JobFileSystemManager):
     _RUNTIME_NAME: t.ClassVar[t.Literal["runtime_code"]] = "runtime_code"
     _INPUT_DATASETS_NAME: t.ClassVar[t.Literal["input_datasets"]] = "input_datasets"
     _CODEBASES_NAME: t.ClassVar[t.Literal["codebases"]] = "codebases"
-    _JOINED_OUTPUT_NAME: t.ClassVar[t.Literal["joined_output"]] = "joined_output"
+    _TEMP_OUTPUT_NAME: t.ClassVar[t.Literal["temp_output"]] = "temp_output"
+    PARTITIONED_OUTPUT_GLOB: t.ClassVar[str] = "*.??????????????.*.nc"
```
ankona: Does this really belong in a shared FS entity?

I think this might make the most sense on something analogous to `RestartFile` or a slightly higher level abstraction since we also use the same basic pattern in `NestIcRunner`, `UpscalerRunner` and `PlotterBlueprint`.

`RomsFileSystemManager` looks like a decent fit, but it is meant to just specify the directory convention. Perhaps a slightly higher level abstraction...  Per our offline convo on my plan to enhance output-awareness of the apps, I think this is a great example of how the `XxxOutputModel` is meant to be used.

cstar/roms/simulation.py
```diff
@@ -836,12 +835,16 @@ def roms_runtime_settings(self) -> "RomsNamelistBase":
         # These next values must stay short: ucla-roms builds
         # the extraction filename in a fixed character(len=99) buffer
         # so an absolute path could silently truncate.
-        nml.simulation_name_settings.output_root_name = "../output/output"
+        # ParallelIO writes whole files directly into `output`; a non-PIO run
+        # writes partitioned pieces into `temp_output`, joined by post_run
+        # into `output`.
+        root = "../output" if self.use_pio else "../temp_output"
+        nml.simulation_name_settings.output_root_name = f"{root}/output"
```
ankona: the output root should only come from `RomsFileSystemManager.output_dir`.

at the very least, this should not be declaring constants `/output` and `/temp_output` when we already declare constants for these names, elsewhere.

We already probably will move those constants to `RomsFileSystemManager`, so we just need to carry it through to this.

